PR #10, opened 2016-05-20
save_csv does not report proper latency distribution

samanbarghi:
when running memtier_benchmark along with --client-stats the printed latency distributions are not actual latencies. Moreover, SET ops percentage are false ( greater than 100% and calculated using total_get_ops).


oranagra:
sorry for the delay, thanks for fixing.

PR #17, opened 2017-01-31
64bits keyspace

swilly22:


obj_gen.cpp
```diff
@@ -384,7 +398,7 @@ data_object* object_generator::get_object(int iter)
         new_size = m_data_size.size_fixed;
     } else if (m_data_size_type == data_size_range) {
         if (m_data_size_pattern && *m_data_size_pattern=='S') {
-            float a = (m_key_index-m_key_min)/(float)(m_key_max-m_key_min);
+            float a = (m_key_index-m_key_min)/(m_key_max-m_key_min);
```
ushachar: Why remove the cast? the new_size calc won't be correct....

ushachar:
@swilly22 - Note your git config sets 'root' as the author....

obj_gen.cpp
```diff
@@ -54,26 +54,40 @@ void random_generator::set_seed(int seed)
...
+    ret = random_r(&m_data_blob, &rn);//max is RAND_MAX
+    assert(ret == 0);
+    llrn |= rn; // set lower 32bits
 #elif (defined HAVE_DRAND48)
     rn = nrand48(m_data_blob); // max is 1<<31
+    llrn = rn;
+    llrn = llrn << 32; // move to upper 32bits
+
+    rn = nrand48(m_data_blob); // max is 1<<31
+    llrn |= rn; // max is 1<<31
```
ushachar: Comment is false...

obj_gen.cpp
```diff
@@ -54,26 +54,40 @@ void random_generator::set_seed(int seed)
...
+    return llrn;
 }
 
-unsigned int random_generator::get_random_max() const
+unsigned long long random_generator::get_random_max() const
 {
 #ifdef HAVE_RANDOM_R
-    return RAND_MAX;
+    unsigned long long rand_max = RAND_MAX;
+    return (rand_max << 32) | RAND_MAX;
```
ushachar, line 88: Not really important,  but you can do ((uint64_t)RAND_MAX << 32) | RAND_MAX; instead.

obj_gen.cpp
```diff
@@ -398,7 +398,7 @@ data_object* object_generator::get_object(int iter)
         new_size = m_data_size.size_fixed;
     } else if (m_data_size_type == data_size_range) {
         if (m_data_size_pattern && *m_data_size_pattern=='S') {
-            float a = (m_key_index-m_key_min)/(m_key_max-m_key_min);
+            float a = (m_key_index-m_key_min)/(float)(m_key_max-m_key_min);
             new_size = (m_data_size.size_range.size_max-m_data_size.size_range.size_min)*a + m_data_size.size_range.size_min;
```
ushachar: This now triggers a compilation warnings on size mismatch - need to add some casts (both here and in random_range two lines below) - and let's use static_cast<> instead of C style...

ushachar:
@swilly22 - Note comment - after you fix please squash and repush, then I'll merge.

obj_gen.cpp
ushachar: (static cast to double -- not to float)

PR #18, opened 2017-02-22
add support for JSON output

elad-ash:
* Add JSON output support: add thin json handling class that could be
overridden for extensive uses
* Add --json-out-file: Name of JSON output file, if not set, will not
print to json
* Refactor some code duplications in the prints sections
* Change license to 2017

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,9 @@
+* Version 1.2.9 
+    * Add JSON output support: add thin json handling class that could be overridden for extensive uses
+    * Add --json-out-file: Name of JSON output file, if not set, will not print to json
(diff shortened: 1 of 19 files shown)


itamarhaber:
@elad-ash Nice! (although you should use ReJSON for that :P)

elad-ash:
@itamarhaber I've used a thin implementation to avoid dependencies. 
By the way, one of the triggers to this activity is that we'd like to push results of performance tests to your JSON module and analyze it with Redis

itamarhaber:
@elad-ash cool - ReJSON is aching for users...

PR #19, opened 2017-02-23
Fix random issue

elad-ash:
Fixes the random problem originated by calling the random creation BEFORE setting the variable from the configuration.


elad-ash:
_**Note** - Version tested vs. Redislabs internal ticket to verify the fix in our QA._

PR #20, opened 2017-02-28
different values for each key when random data is set

yoav-steinberg:
Useful in case we want to make sure the dataset content is constantly changing during the benchmark.
Doesn't impact performance because only a single byte is modified in the buffer per write. On the other hand data isn't really random (but still more random than before).

### DIFF
--- a/obj_gen.cpp
+++ b/obj_gen.cpp
@@ -146,7 +146,9 @@ object_generator::object_generator() :
     m_key_stddev(0),
     m_key_median(0),
     m_value_buffer(NULL),
-    m_random_fd(-1)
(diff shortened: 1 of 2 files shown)


yoav-steinberg:
Currently there's absolutely no way to make the dataset change during 
the benchmark. In my case this limited my ability to verify data 
persistence after a "debug load"/"debug loadaof" call while testing data 
persistence during high load. I needed both high performance and data to 
mutate. The question here should be are we looking for random data or 
data mutation. The original purpose for the "-R" option was to create 
non compressible data. Adding this simple mutation preserves that 
purpose and keeps high performance.

PR #21, opened 2017-03-21
Fix missing closing of JSON nesting for results printing

elad-ash:
Added if (jsonhandler != NULL){ jsonhandler->close_nesting();} for making sure that Best, Worst and Average results will be brothers in the JSON

### DIFF
--- a/client.cpp
+++ b/client.cpp
@@ -1593,6 +1593,10 @@ void run_stats::print(FILE *out, bool histogram, const char * header/*=NULL*/,  
         }
         if (jsonhandler != NULL){ jsonhandler->close_nesting();}
     }
+    // This close_nesting closes either:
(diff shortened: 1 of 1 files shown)


client.cpp
```diff
@@ -1593,5 +1593,6 @@ void run_stats::print(FILE *out, bool histogram, const char * header/*=NULL*/,
         }
         if (jsonhandler != NULL){ jsonhandler->close_nesting();}
     }
+    if (jsonhandler != NULL){ jsonhandler->close_nesting();}
```
swilly22: Add comment:
"This close_nesting closes either:
jsonhandler->open_nesting(header); or
jsonhandler->open_nesting("UNKNOWN STATS");
From the top (beginning of function).

PR #22, opened 2017-03-29
Fix mem corruption when using random data

yoav-steinberg:


elad-ash:
Before this change, we got duplicated free events in the output, fixed after the change.

PR #37, opened 2017-09-27
add support for oss cluster

YaacovHazan:
new 'oss_client' client add to support oss cluster.
the new client use CLUSTER_SLOTS command to retrive the
key-->node mapping, create separate connection fot each node,
and send the commands to the appropriate connection according
to hash calculation.

### DIFF
--- a/oss_client.cpp
+++ b/oss_client.cpp
@@ -1,0 +1,907 @@
+/*
+ * Copyright (C) 2011-2017 Redis Labs Ltd.
+ *
+ * This file is part of memtier_benchmark.
+ *
+ * memtier_benchmark is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, version 2.
+ *
+ * memtier_benchmark is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with memtier_benchmark.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#ifdef HAVE_SYS_TYPES_H
+#include <sys/types.h>
+#endif
+#ifdef HAVE_FCNTL_H
+#include <fcntl.h>
+#endif
+#include <unistd.h>
+#include <stdlib.h>
+#include <string.h>
+#include <errno.h>
+#ifdef HAVE_SYS_SOCKET_H
+#include <sys/socket.h>
+#endif
+#ifdef HAVE_NETINET_TCP_H
+#include <netinet/tcp.h>
+#endif
+#ifdef HAVE_LIMITS_H
+#include <limits.h>
+#endif
+
+#ifdef HAVE_ASSERT_H
+#include <assert.h>
+#endif
+
+#include "oss_client.h"
+#include "obj_gen.h"
+#include "memtier_benchmark.h"
+
+///////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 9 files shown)


oss_client.cpp
```diff
@@ -0,0 +1,907 @@
...
+        protocol = NULL;
+    }
+
+    if (pipeline != NULL) {
+        delete pipeline;
+        pipeline = NULL;
+    }
+}
+
+void shard_connection::setup_connection(int conn_id, oss_client* parent_client, abstract_protocol* abs_protocol) {
```
yossigo, line 156: There is quite a lot of duplicate code between oss_client.c and client.c, I think it should be avoided.

oss_client.h
```diff
@@ -0,0 +1,79 @@
+//
```
yossigo, line 1: Standard header is missing

oss_client.cpp
```diff
@@ -0,0 +1,907 @@
...
+        return m_cluster_slots_state == slots_done;
+    else
+        return false;
+}
+
+bool oss_client::send_conn_setup_commands(struct timeval timestamp)
+{
+    bool sent = false;
+
+    if (m_config->authenticate && m_authentication != auth_done) {
```
yossigo, line 451: Unless I'm missing something here, the authentication and DB selection are always done on the main connection and never on the other connections.

In fact, the authentication and selected DB state belong to the shard connection rather than the client.

oss_client.cpp
```diff
@@ -0,0 +1,907 @@
...
+        addr[mbulk_addr_el->value_len] = '\0';
+
+        // port
+        mbulk_element* mbulk_port_el = shard->mbulk_array[2]->mbulk_array[1];
+        char* port = (char*) malloc(mbulk_port_el->value_len + 1);
+        memcpy(port, mbulk_port_el->value, mbulk_port_el->value_len);
+        port[mbulk_port_el->value_len] = '\0';
+
+        // if it is the main connection, no need to add connection, just update range
+        if (strcmp(addr, m_config->server) == 0) {
```
yossigo, line 514: There could be a few shards on the same server, you need to compare ports as well.

oss_client.cpp
```diff
@@ -0,0 +1,907 @@
...
+        m_tot_wait_ops++;
+
+        unsigned int num_slaves = m_obj_gen->random_range(m_config->num_slaves.min, m_config->num_slaves.max);
+        unsigned int timeout = m_obj_gen->normal_distribution(m_config->wait_timeout.min,
+                                                              m_config->wait_timeout.max, 0,
+                                                              ((m_config->wait_timeout.max - m_config->wait_timeout.min)/2.0) + m_config->wait_timeout.min);
+
+        benchmark_debug_log("WAIT num_slaves=%u timeout=%u\n", num_slaves, timeout);
+
+        // currently wait commands goes to main connection
```
yossigo, line 648: Why only to the main connection?  It means it's only effective for 1/N shards.

oss_client.cpp
```diff
@@ -0,0 +1,907 @@
...
+    if (ret < 0) {
+        benchmark_error_log("connect: failed to setup socket\n");
+        return -1;
+    }
+
+    // set up read/write events
+    conn->setup_events(m_event_base);
+
+    // call connect
+    if (::connect(conn->sockfd,
```
yossigo, line 413: Again, low level networking code duplication should be avoided.

oss_client.cpp
```diff
@@ -0,0 +1,907 @@
...
+        conn->pending_reqs++;
+        m_remain_reqs--;
+
+        if (conn->pipeline->size() >= m_config->pipeline) {
+            m_write_pipe = true;
+            m_full_pipe_id = 0;
+        }
+    }
+        // are we set or get? this depends on the ratio
+    else if (m_set_ratio_count < m_config->ratio.a) {
```
yossigo, line 662: Once again, a huge part of this is duplicated with client.c with no apparent reason.  It would make sense to pack it in a class hierarchy to share all identical behavior between the Redis Cluster and Standard Redis clients.

PR #42, opened 2017-11-22
add support for cluster mode

YaacovHazan:
new 'cluster_client' add to support cluster mode.
the new client use CLUSTER_SLOTS command to retrive the
key-->node mapping, create separate connection fot each node,
and send the commands to the appropriate connection according
to hash calculation.

### DIFF
--- a/cluster_client.cpp
+++ b/cluster_client.cpp
@@ -1,0 +1,357 @@
+/*
+ * Copyright (C) 2011-2017 Redis Labs Ltd.
+ *
+ * This file is part of memtier_benchmark.
+ *
+ * memtier_benchmark is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, version 2.
+ *
+ * memtier_benchmark is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with memtier_benchmark.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#ifdef HAVE_SYS_TYPES_H
+#include <sys/types.h>
+#endif
+#ifdef HAVE_FCNTL_H
+#include <fcntl.h>
+#endif
+#include <unistd.h>
+#include <stdlib.h>
+#include <string.h>
+#include <errno.h>
+#ifdef HAVE_SYS_SOCKET_H
+#include <sys/socket.h>
+#endif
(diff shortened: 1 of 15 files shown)


cluster_client.cpp
```diff
@@ -0,0 +1,358 @@
...
+        // check if the key match for this connection
+        unsigned int hslot = calc_hslot_crc16_cluster(m_key_buffer, m_key_len);
+        if (m_slot_to_shard[hslot] == conn_id) {
+            m_reqs_generated++;
+            return true;
+        }
+
+        // store key for other connection, if queue is not full
+        key_index_pool* key_idx_pool = m_key_index_pools[m_slot_to_shard[hslot]];
+        if (key_idx_pool->size() < KEY_INDEX_QUEUE_MAX_SIZE) {
```
yossigo, line 294: Did you confirm that in all possible scenarios it is ok for the obj_gen to return a key that is simply discarded (because the queue is full)?

cluster_client.cpp
```diff
@@ -0,0 +1,358 @@
...
+
+static uint32_t calc_hslot_crc16_cluster(const char *str, size_t length)
+{
+    uint32_t rv = (uint32_t) crc16(str, length) & MAX_CLUSTER_HSLOT;
+    return rv;
+}
+
+///////////////////////////////////////////////////////////////////////////////////////////////////////
+
+cluster_client::cluster_client(client_group* group) : client(group)/*, m_finished_set(false), m_reqs_generated(0)*/
```
yossigo: Make sure unused code is removed, not just commented

YaacovHazan: ok, I will remove it

cluster_client.cpp
YaacovHazan: The answer divide for two parts:
1. For Random/Gaussian:
We think that it is ok for those options, since the discarding of keys suppose to be relative to the selected pattern, so in the overall the distribution will stay as it suppose to be (Random/Gaussian).

2. For Sequential:
Here we have problem that we know about it...
The issue start when the different connections are not symmetrical (networking/shards conuts ...).

In that case one connection can be slow from other and his queue starting to grow over and over and the cluster-client start to discard keys.
In sequential distribution we have issue if user ask for keys (for ex) between  1 - 1000000.

If user set memtier to --test-time and not --requests option, there isn't really issue.
Since the connections are independent, it could be gaps in the DB even if cluster-client will not discard any keys.

The problem is only when user set memtier to --requests option, and if he ask 1000000 request in range of 1 - 1000000 he expect that at the end of the test, the DB will be with those exactly keys (1-10000000), and since we discards keys this will not be the result.

We think about two solutions for this case:
1. add option for user to set the queue limit size.
2. add option to do some balancing between the connections, so if some connection reach to some threshold, the other connection start to send less requests or stop send request, until the slower connection go down below the the threshold.

What do you think?
Is this issue is critical? or we can solved in the future.
The solutions are good, do you have any other idea?

yossigo: I don't think it's critical, it should probably be documented though in a section about cluster support caveats.

PR #54, opened 2018-06-17
support MOVED in cluster mode

YaacovHazan:
when using cluster mode, "-ERR MOVED..." response can return from
cluster.

adding support to handle MOVED reply, and ask for new topolgy
by resending CLUSTER SLOTS command.


itamarhaber:
Nice!!!

Makefile.am
```diff
@@ -24,6 +24,7 @@ bin_PROGRAMS = memtier_benchmark
 memtier_benchmark_CPPFLAGS = $(LIBEVENT_CFLAGS)
 memtier_benchmark_SOURCES = \
 	memtier_benchmark.cpp memtier_benchmark.h \
+	run_stats.cpp run_stats.h \
```
yossigo: @YaacovHazan I think it would be a good idea to separate the stats code reorganization from the MOVED commit.

client.cpp
```diff
@@ -350,8 +311,13 @@ int client::prepare(void)
     return 0;
 }
 
-void client::handle_response(struct timeval timestamp, request *request, protocol_response *response)
+void client::handle_response(unsigned int conn_id, struct timeval timestamp,
```
yossigo, line 346: @YaacovHazan I may be missing something, but what is the purpose of conn_id in handle_request/handle_response? I don't see it being used.

cluster_client.cpp
```diff
@@ -234,14 +249,24 @@ void cluster_client::handle_cluster_slots(protocol_response *r) {
             if (strcmp(addr, m_connections[j]->get_address()) == 0 &&
```
yossigo, line 249: @YaacovHazan How often is this used (I did not follow all code paths)? I wonder if there's a point to optimize connection references with a map.  In general, we need to be wary of performance regressions...

cluster_client.cpp
```diff
@@ -282,15 +320,29 @@ bool cluster_client::get_key_for_conn(unsigned int conn_id, int iter, unsigned l
...
+        // check if the key match for this connection
         if (m_slot_to_shard[hslot] == conn_id) {
             m_reqs_generated++;
             return true;
         }
 
+        // handle key for other connection
+        unsigned int other_conn_id = m_slot_to_shard[hslot];
+
+        // in case we generated key for connection that is disconnected, 'slot to shard' map may need to be updated
```
yossigo, line 334: @YaacovHazan It's not clear why would this happen and what should happen next.

cluster_client.cpp
```diff
@@ -355,3 +408,41 @@ void cluster_client::create_request(struct timeval timestamp, unsigned int conn_
     }
 }
 
+void cluster_client::handle_response(unsigned int conn_id, struct timeval timestamp,
+                                     request *request, protocol_response *response) {
+    // handle "-MOVED"
+    if (response->is_error() &&
+        strncasecmp(response->get_status(), MOVED_MSG_PREFIX, MOVED_MSG_PREFIX_LEN) == 0) {
```
yossigo: @YaacovHazan I guess `strncmp()` could do here as well.

cluster_client.cpp
```diff
@@ -355,3 +408,41 @@ void cluster_client::create_request(struct timeval timestamp, unsigned int conn_
...
+
+        // connection already issued 'cluster slots' command, wait for slots mapping to be updated
+        if (m_connections[conn_id]->get_cluster_slots_state() != slots_done)
+            return;
+
+        // queue may stored uncorrected mapping indexes, empty them
+        key_index_pool empty_queue;
+        std::swap(*m_key_index_pools[conn_id], empty_queue);
+
+        // set connection to send 'CLUSTER SLOTS' command
```
yossigo, line 439: @YaacovHazan Is there a deliberate design decision to use `CLUSTER SLOTS` and not just update the specific slot mapping (as you already have enough information to do that)?

Makefile.am
YaacovHazan: ok, I will move it back

client.cpp
YaacovHazan: I'm using it in cluter_client

cluster_client.cpp
YaacovHazan: We not using it so often.
We call it once when memtier starts to get cluster topology, and then only when we got MOVED errors

cluster_client.cpp
YaacovHazan: This scenario related to how our cluster and DMC handle the topology in cluster mode.
when we migrate shard from node to node it's take some time to all DMC's to get this information.

consider this scenario:
topology
node1 with shard1 0-8192
node2 with shard2 8193-16384

memtier client run and open two connections (conn1 & conn2)
in some point we migrate shard1 to node2.
conn1 will get moved from DMC1 and will send CLUSTER_SLOT to DMC1 that will send the new topolgy:
node2 0-16384

but since DMC2 didn't receive yet the new topolgy, when node2 will send key that map to 0-8192,
he will get MOVED and will send CLUSTER_SLOT to DMC2 that will send the old topolgy (DMC2 didn't updated yet):
node1 with shard1 0-8192
node2 with shard2 8193-16384

in this point conn1 may not connected any more, since we have no shards in node1 and we removing the endpoint from node1.
so node2 will send only keys that map to 8193-16384 with no MOVED reply, which is correct, but he will not know that he can and need to send also keys to 0-8192.

for that case I added this handle of situation that we generate key to connection that is disconnected, but I'm not sure we want to add this internal handling explanation in the comment

PR #60, opened 2018-07-31
support ASK in cluster mode

YaacovHazan:
when using cluster mode, "-ASK ..." response can return from
cluster.

adding support to handle ASK reply.

### DIFF
--- a/client.cpp
+++ b/client.cpp
@@ -349,7 +349,10 @@ void client::handle_response(unsigned int conn_id, struct timeval timestamp,
                              request *request, protocol_response *response)
 {
     if (response->is_error()) {
-        benchmark_error_log("error response: %s\n", response->get_status());
+        benchmark_error_log("server %s:%s handle error response: %s\n",
+                            m_connections[conn_id]->get_address(),
+                            m_connections[conn_id]->get_port(),
+                            response->get_status());
     }
 
     switch (request->m_type) {
@@ -676,6 +679,7 @@ void run_stats::one_second_stats::reset(unsigned int second)
     m_ops_get = m_ops_set = m_ops_wait = 0;
     m_get_hits = m_get_misses = 0;
     m_moved_get = m_moved_set = 0;
+    m_ask_get = m_ask_set = 0;
     m_total_get_latency = 0;
     m_total_set_latency = 0;
     m_total_wait_latency = 0;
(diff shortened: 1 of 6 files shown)


client.cpp
```diff
@@ -349,7 +349,10 @@ void client::handle_response(unsigned int conn_id, struct timeval timestamp,
                              request *request, protocol_response *response)
 {
     if (response->is_error()) {
-        benchmark_error_log("error response: %s\n", response->get_status());
+        benchmark_error_log("server %s:%s handle error response: %s\n",
+                            m_connections[conn_id]->get_address(),
+                            m_connections[conn_id]->get_port(),
```
yossigo: Isn't this log incompatible with unix sockets?  Maybe better to hide server ID behind an abstraction in shard_connection.

client.cpp
```diff
@@ -938,8 +982,8 @@ void run_stats::save_csv_one_sec_cluster(FILE *f,
                                          unsigned long int& total_set_ops,
                                          unsigned long int& total_wait_ops) {
     fprintf(f, "Per-Second Benchmark Data\n");
-    fprintf(f, "Second,SET Requests,SET Average Latency,SET Total Bytes,SET Moved,"
-               "GET Requests,GET Average Latency,GET Total Bytes,GET Misses,GET Hits,GET Moved,"
+    fprintf(f, "Second,SET Requests,SET Average Latency,SET Total Bytes,SET Moved, SET Ask,"
```
yossigo: Not sure about breaking output format just for the sake of ASK stats.  I'd either make this optional or at least push it to the end of the CSV to minimize risk of breaking something.

cluster_client.cpp
```diff
@@ -408,38 +411,69 @@ void cluster_client::create_request(struct timeval timestamp, unsigned int conn_
...
-                                        request->m_size + response->get_total_len(),
-                                        ts_diff(request->m_sent_time, timestamp));
-        } else if (request->m_type == rt_set) {
-            m_stats.update_moved_set_op(&timestamp,
-                                        request->m_size + response->get_total_len(),
-                                        ts_diff(request->m_sent_time, timestamp));
-        } else {
-            assert(0);
+    if (response->is_error()) {
+        benchmark_debug_log("server %s:%s handle response: %s\n",
```
yossigo: See above re/ Unix sockets.

PR #65, opened 2018-11-29
restructure run_stats

YaacovHazan:
in order to simplify the ability to add statistics to
a new command, the run_stats class were restructured


run_stats_types.cpp
```diff
@@ -0,0 +1,144 @@
+//
```
yossigo, line 1: @YaacovHazan Please stick to the standard header for new files in this project.

yaacovhazan-Redislabs:
replaced with 'Stats restructure and reorganization #66'

Issue #28, opened 2017-04-15
[Feature request] Allow benchmarking arbitrary commands


itamarhaber:
Similar to how `redis-benchmark` works - would be extremely useful for module authors...

accandap:
Can you please tell me the format to pass the arguments. I tried everything and it returns "ERR wrong number of arguments for 'publish' command"

yaacovhazan-Redislabs:
@accandap did you add quotation marks around the command? (--command="PUBLISH FOO TEST")
If it does not help, please provide the full memtier command that you run

accandap:
@yaacovhazan-Redislabs  This was silly from my side, did not try that .. Above solution worked. Thanks!!

PR #67, opened 2018-12-19
add support for aribitrary command

YaacovHazan:
add new option to run arbitrary command instead of
the SET:GET commands.

### DIFF
--- a/config_types.cpp
+++ b/config_types.cpp
@@ -263,3 +263,215 @@ const char* server_addr::get_last_error(void) const
 {
     return gai_strerror(m_last_error);
 }
+
+static int hex_digit_to_int(char c) {
+    switch (c) {
+        case '0':
+            return 0;
+
+        case '1':
+            return 1;
+
+        case '2':
+            return 2;
+
+        case '3':
+            return 3;
+
+        case '4':
+            return 4;
+
+        case '5':
+            return 5;
+
+        case '6':
+            return 6;
+
+        case '7':
+            return 7;
+
+        case '8':
+            return 8;
+
+        case '9':
+            return 9;
+
+        case 'a':
+        case 'A':
+            return 10;
+
+        case 'b':
+        case 'B':
+            return 11;
+
+        case 'c':
+        case 'C':
+            return 12;
+
+        case 'd':
+        case 'D':
+            return 13;
+
+        case 'e':
+        case 'E':
+            return 14;
+
+        case 'f':
+        case 'F':
+            return 15;
+
+        default:
+            return 0;
+    }
+}
+
+#define MAX_ARGS 4096
+bool arbitrary_command::split_command_to_args(const char* command) {
+    const char *p = command;
+    size_t command_len = strlen(command);
+
+    char buffer[command_len];
+    unsigned int buffer_len = 0;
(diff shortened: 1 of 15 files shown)


itamarhaber:
Fixes #28

config_types.cpp
```diff
@@ -263,3 +263,215 @@ const char* server_addr::get_last_error(void) const
 {
     return gai_strerror(m_last_error);
 }
+
+static int hex_digit_to_int(char c) {
```
yossigo, line 267: 1) Does it really make sense to implement ourselves?
2) This implementation does not distinguish '0' from an invalid input, which is a bad idea.
3) Why not simple arithmetics on the value?

config_types.cpp
```diff
@@ -263,3 +263,215 @@ const char* server_addr::get_last_error(void) const
...
+            return 0;
+    }
+}
+
+#define MAX_ARGS 4096
+bool arbitrary_command::split_command_to_args(const char* command) {
+    const char *p = command;
+    size_t command_len = strlen(command);
+
+    char buffer[command_len];
```
yossigo, line 283: buffer should be bigger for null terminator no?

config_types.cpp
```diff
@@ -263,3 +263,215 @@ const char* server_addr::get_last_error(void) const
...
+        case 'f':
+        case 'F':
+            return 15;
+
+        default:
+            return 0;
+    }
+}
+
+#define MAX_ARGS 4096
```
yossigo: Do we really need this limit? And if we do, is this a reasonable (i.e. not too big) one?

protocol.cpp
```diff
@@ -379,144 +365,177 @@ int redis_protocol::write_command_wait(unsigned int num_slaves, unsigned int tim
...
-                    m_last_response.set_status(line);
+                    // from bulks counter perspective every count < 0 is equal to 0, because it's not followed by bulks.
+                    if (count < 0) {
+                        count = 0;
+                    }
 
-                } else if (line[0] == ':') {
-                    char *bulk_value = strdup(line + 1);
-                    assert(bulk_value != NULL);
+                    if (m_keep_value) {
```
yossigo, line 405: Does it actually work with keep_value == true? (i.e. verify client)
If not we should at least disable it with arbitrary commands where we know nothing of the reply structure.

config_types.cpp
YaacovHazan: Done

config_types.cpp
YaacovHazan: This is buffer for bytes, I'm not using it as a null-terminated string and when I initialize the std::string I pass it with length.

config_types.cpp
YaacovHazan: Actually, I took it from DMC code... but you are right there isn't any need of this limit... so I removed it

protocol.cpp
YaacovHazan: I'm not sure I understood your comment... why do you think it can't work?

yossigo: If I understand the change correctly, it flattens a multi-level multibulk which I thought could have an effect on the verify client.  But I see this combination of commands is anyway not possible.

PR #70, opened 2019-03-09
Dockerfile optimizations

pataquets:
Optimize final image build size by grouping commands in the same build step and deleting unneeded package installation files. Add non-interactive env vars.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,7 @@ COPY --from=builder /usr/local/bin/memtier_benchmark /usr/local/bin/memtier_benc
 RUN \
   apt-get update && \
   DEBIAN_FRONTEND=noninteractive \
-    apt-get install -y \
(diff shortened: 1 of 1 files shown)


itamarhaber:
@pataquets one small thing - could you also test whether adding `no-install-recommends` to apt reduces the build time and at the same time does not break it?

pataquets:
@itamarhaber : Done. Works for me, it didn't broke.
Also, I've done a few improvements in separate commits. Mainly changing ```-yy``` for ```-y```, since some nasty errors might be hidden otherwise. Make it a bit more "fail fast, fail loud" in case something is wrong when building.
I've set up an automated build at https://hub.docker.com/r/pataquets/memtier_benchmark ease testing.

itamarhaber:
Thanks @pataquets - merged.

PR #72, opened 2019-03-24
add support for multiple arbitrary commands

YaacovHazan:
this commit expand the ability of running arbitrary command
to enable runing multiple arbitrary commands by
adding the --command option for each command.

### DIFF
--- a/memtier_benchmark.cpp
+++ b/memtier_benchmark.cpp
@@ -287,10 +287,10 @@ static bool verify_cluster_option(struct benchmark_config *cfg) {
 
 static bool verify_arbitrary_command_option(struct benchmark_config *cfg) {
     if (cfg->key_pattern) {
-        fprintf(stderr, "error: when using arbitrary command, key pattern doesn't configured via key-pattern option.\n");
+        fprintf(stderr, "error: when using arbitrary command, key pattern is configured with --command-key-pattern option.\n");
         return false;
     } else if (cfg->ratio.is_defined()) {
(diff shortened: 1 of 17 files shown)


memtier_benchmark.cpp
```diff
@@ -278,45 +277,20 @@ static bool verify_cluster_option(struct benchmark_config *cfg) {
...
-         cfg->key_pattern[key_pattern_get] != 'S' &&
-         cfg->key_pattern[key_pattern_get] != 'G' &&
-         cfg->key_pattern[key_pattern_get] != 'P')) {
-
-        fprintf(stderr, "error: key-pattern must be in the format of [S/R/G/P]%s.\n",
-                cfg->command ? "" : ":[S/R/G/P]");
-
+static bool verify_arbitrary_command_option(struct benchmark_config *cfg) {
+    if (cfg->key_pattern) {
+        fprintf(stderr, "error: when using arbitrary command, key pattern doesn't configured via key-pattern option.\n");
```
yossigo: typo: isn't.
also: maybe hint to the correct option to use?

memtier_benchmark.cpp
```diff
@@ -278,45 +277,20 @@ static bool verify_cluster_option(struct benchmark_config *cfg) {
...
+        return false;
+    } else if (cfg->ratio.is_defined()) {
+        fprintf(stderr, "error: when using arbitrary command, ratio doesn't configured via ratio option.\n");
```
yossigo: same as above.

PR #74, opened 2019-04-08
add bash completion script

YaacovHazan:
add bash completion for memtier-benchmark


itamarhaber:
@YaacovHazan :100:

itamarhaber:
@YaacovHazan I take back my :100: - what about updating the README with relevant instructions on how this should be used (i.e. `cp ...`)?

PR #85, opened 2019-05-05
Version bump 1.2.17

YaacovHazan:


CHANGES
```diff
@@ -1,3 +1,7 @@
+* Version 1.2.17
+    * fix parallel key-pattern command line usage
+    * fix and clean up build on macOS
```
itamarhaber, line 3: Should be "OSX" or perhaps "Darwin"

yossigo: Not a Mac enthusiast but didn't Apple go from OS X to macOS a few years ago?

itamarhaber: Not a Mac enthusiast as well, I don't know.

PR #88, opened 2019-06-20
Add example/testing Docker Compose manifests

pataquets:
For both Redis and Memcached stacks.
Using Docker Hub's public images by default.
With commented out sample lines for building local code and for customizing Redis/Memcached command line parameters.


pataquets:
@itamarhaber ping?

itamarhaber:
Hello @pataquets 

Thanks for the PR - the yamls look ok. Can you explain why they should be a part of the project?

pataquets:
They serve both as Docker usage quick examples and as a basis to allow local develop/build/run workflow based on Docker, too (no dev libs/tools needed to install).

itamarhaber:
Thanks, I had figured out that much, but I'm still not convinced that they should be added. @RoeyPrat @YaacovHazan what do you say?

RoeyPrat:
I think examples should be added in markdown files, in code snippets.

As for testing, we should probably add continuous integration using circle-ci or github actions

pataquets:
Examples in markdown would be another way of doing it, although I prefer the manifest because it's runnable straight away and this the code are the docs, but just a matter of personal preferences, I suppose.
However, for the local dev workflow, it takes away all the complexities not by just setting up and installing all the build deps, but also taking care of installing and running whatever Redis/Memcached service is also necessary with ready and obvious command line tuning. You don't have to download and install them and you even can use any version, independent of you host OS packaged ones. Not everyone knows/wants to install a daemon just for development, so a manifest leaves your system tidy after you're finished without leaving traces.
As for the docs you mention, I can add an example run command in the readme, if it helps.

pataquets:
@itamarhaber @RoeyPrat any feedback on this?

pataquets:
@itamarhaber @RoeyPrat checking the Dockerfile, I've realized that the build is done using source code _downloaded_ from Github. Most of the Docker workflow advantages come from building the artifact from local Docker build context, especialy the source code.
I've changed this, so now it makes more sense to have a Docker Compose manifest for local development, since it will build from local source, too. Thus, anybody can start hacking right away without having to install anything else besides Docker and Docker Compose. Now it makes more sense, IMO.

pataquets:
@RoeyPrat : Done. Now that #113 is in, makes more sense to build images by using local context instead of pulling them.

RoeyPrat:
@itamarhaber my +1 doesn't allow merging :)

RoeyPrat:
thanks @pataquets!

PR #90, opened 2019-08-18
TLS Support

yossigo:


memtier_benchmark.h
```diff
@@ -45,6 +49,14 @@ struct benchmark_config {
     const char *protocol;
     const char *out_file;
     const char *client_stats;
+    bool tls;
```
yaacovhazan-Redislabs: why not to insert the variables under the "#ifdef USE_TLS"?

shard_connection.cpp
```diff
@@ -188,97 +179,94 @@ shard_connection::~shard_connection() {
...
+        bufferevent_free(m_bev);
+    }
 
-    if (!m_event) {
-        m_event = event_new(m_event_base, m_sockfd, EV_WRITE,
-                            cluster_client_event_handler, (void *)this);
-        assert(m_event != NULL);
+#ifdef USE_TLS
+    if (m_config->openssl_ctx) {
+        SSL *ctx = SSL_new(m_config->openssl_ctx);
```
yaacovhazan-Redislabs, line 189: don't you want to check if SSL_new return NULL?

memtier_benchmark.cpp
```diff
@@ -751,6 +790,21 @@ static int config_parse_args(int argc, char *argv[], struct benchmark_config *cf
                     }
                     break;
                 }
+                case o_tls:
```
yaacovhazan-Redislabs, line 794: I'm understanding that having that code without define USE_TLS do nothing, but since we already wrap every piece of code that related to TLS with USE_TLS maybe it will be more clearly to wrap this section too (this come with my below comment about wrapping the variables declaration too)

memtier_benchmark.h
yossigo: I'll move everything TLS into the ifdef.

shard_connection.cpp
yossigo: Adding an `assert()` as we currently do the same for libevent failures, etc.

Issue #68, opened 2019-02-19
Provide more latency data in results


zhouyuan:
Hi memtier_benchmark developers, 

Currently memtier_benchmark can report average latency in the results, which is usually enough for simple benchmarks. it also provides the option to print the detail latency distribution. The data is aggregated through different latency level. However for storage/DB performance, people will also care about the tail latency(99%, 99.9% or 99.99%,  [example](https://github.com/axboe/fio/blob/307f2246d65d9dbea7e5f56cd17734cb68c2817f/HOWTO#L3528) from fio).  Is it possible to provide these tail latency data also? 

```
ALL STATS
========================================================================
Type        Ops/sec     Hits/sec   Misses/sec      Latency       KB/sec
------------------------------------------------------------------------
Sets       21536.09          ---          ---      0.23100     22561.95
Gets           0.00         0.00         0.00      0.00000         0.00
Waits          0.00          ---          ---      0.00000          ---
Totals     21536.09         0.00         0.00      0.23100     22561.95
```
thanks, -yuan

yaacovhazan-Redislabs:
Hi @zhouyuan, we think that this is a good idea, and can really help the users to understand the benchmark results.
We will investigate how to best implement your suggestion for practical use.

zhouyuan:
Thanks a lot! @yaacovhazan-Redislabs 


FWIW I did some naive hack here to extract the 99% latency data, not that accurate but can help me to do more analysis with the results

```
diff --git a/run_stats.cpp b/run_stats.cpp
index 3b89a14..de82c6f 100644
--- a/run_stats.cpp
+++ b/run_stats.cpp
@@ -38,6 +38,7 @@

 #define MIN(a,b) (((a)<(b))?(a):(b))
 #define MAX(a,b) (((a)>(b))?(a):(b))
+#define TAIL_LAT 99.00

 void output_table::add_column(table_column& col) {
     assert(columns.empty() || columns[0].elements.size() == col.elements.size());
@@ -867,20 +868,34 @@ void run_stats::print(FILE *out, bool histogram,
         } else {
             // SETs
             // ----
+           double prev = 0;
+           double curr = 0;
             if (jsonhandler != NULL){ jsonhandler->open_nesting("SET",NESTED_ARRAY);}
             for( latency_map_itr_const it = m_set_latency_map.begin() ; it != m_set_latency_map.end() ; it++) {
                 total_count += it->second;
[...]

PR #100, opened 2019-11-10
enable full percentile spectrum analysis on latency

filipecosta90:
# Full percentile spectrum analysis
Fixes: #68 
This PR enable full percentile spectrum analysis on query latency so that we can fully understand the performance and stability characteristics of the system we are measuring. 

Right now memtier_benchmark is displaying single values targets like the average. 
Considering that latency does not obey a normal distribution we should enable ourselves to do full percentile spectrum analysis. Just looking at a number might be misleading. 

A useful visualization for pointing out the folly of chasing a mean measurement is illustrated below - they have nearly identical simple descriptive statistics, yet have very different distributions and appear very different when graphed.

image

Source: https://en.wikipedia.org/wiki/Anscombe%27s_quartet

_____
_______
# Enabling full percentile latencies on memtier_benchmark:
The previous features can be applied with the changes added on this PR using the following added options:
```
      --hdr-file-prefix=FILE     Prefix of HDR Latency Histogram output files, if not set, will not save latency histogram files
      --print-percentiles          Specify which percentiles info to print on the results table (by default prints percentiles: 50,99,99.9)

```


memtier_benchmark.cpp
```diff
@@ -272,6 +272,14 @@ static void config_init_defaults(struct benchmark_config *cfg)
     }
     if (!cfg->requests && !cfg->test_time)
         cfg->requests = 10000;
+    if (!cfg->hdr_prefix)
+        cfg->key_prefix = "";
+    if (!cfg->save_hdr)
```
yossigo: @filipecosta90 I think this is redundant...

memtier_benchmark.cpp
```diff
@@ -790,6 +815,14 @@ static int config_parse_args(int argc, char *argv[], struct benchmark_config *cf
                     }
                     break;
                 }
+                case o_rate_limit_rps:
+                endptr = NULL;
```
yossigo: @filipecosta90 nitpicking: indentation is wrong here.

run_stats.cpp
```diff
@@ -985,7 +1214,22 @@ void run_stats::print(FILE *out, benchmark_config *config,
...
+        char average_header[50];
+        int ndigts = 0;
+        float num = abs(quantile);
+        num = num - int(num);
+        while (abs(num)>0.001 && ndigts < 3 ) {
+            num = num * 10;
+            ndigts++;
+            num = num - int(num);
+        }
+        sprintf(average_header,"q%.*f Latency", ndigts, quantile);
```
yossigo: @filipecosta90 Better stick to `snprintf()`.

run_stats.cpp
```diff
@@ -422,18 +434,207 @@ void run_stats::save_csv_arbitrary_commands(FILE *f, arbitrary_command_list& com
...
+            snprintf(fmtbuf, sizeof(fmtbuf) - 1, "%s_FULL_RUN_%d.hgrm", config->hdr_prefix, run_number);
+            fprintf(stderr, "Writing Full Run command HDR latency histogram results in HistogramLogProcessor format to %s...\n", fmtbuf);
+            hdr_outfile = fopen(fmtbuf, "w");
+            if (!hdr_outfile){
+                perror(config->hdr_prefix);
+                return false;
+            }
+            hdr_log_write_header(hdr_outfile, "Full Run command HDR latency histogram results", &start_timespec);
+            hdr_log_write(hdr_outfile, &start_timespec, &end_timespec, m_totals_latency_histogram);
+            fflush(hdr_outfile);
```
yossigo: @filipecosta90 `fflush()` is not needed before `fclose()`.

run_stats.cpp
```diff
@@ -422,18 +434,207 @@ void run_stats::save_csv_arbitrary_commands(FILE *f, arbitrary_command_list& com
...
+        // Prepare output file
+        FILE *hdr_outfile;
+        char fmtbuf[1024];
+        snprintf(fmtbuf, sizeof(fmtbuf) - 1, "%s_GET_command_run_%d.txt", config->hdr_prefix, run_number);
+        fprintf(stderr, "Writing GET command HDR latency histogram results to %s...\n", fmtbuf);
+        hdr_outfile = fopen(fmtbuf, "w");
+        if (!hdr_outfile){
+            perror(config->hdr_prefix);
+            return false;
+        }
```
yossigo: @filipecosta90 Consider consolidating duplicate code, possibly making a single generic save method that just accepts the name/pointer to hdr/etc. parameters.

run_stats.cpp
```diff
@@ -479,17 +680,26 @@ void run_stats::debug_dump(void)
...
+        benchmark_debug_log("  GET <= %u msec: %u\n", value,percentiles->percentile);
     }
-    for(  latency_map_itr it = m_set_latency_map.begin() ; it != m_set_latency_map.end() ; it++) {
-        if (it->second)
-            benchmark_debug_log("  SET <= %u msec: %u\n", it->first, it->second);
+    hdr_iter_percentile_init(&iter, m_set_latency_histogram, LATENCY_HDR_GRANULARITY);
+    percentiles = &iter.specifics.percentiles;
+    while (hdr_iter_next(&iter)){
+        double value = iter.highest_equivalent_value / (double) LATENCY_HDR_RESULTS_MULTIPLIER;
+        benchmark_debug_log("  SET <= %u msec: %u\n", value,percentiles->percentile);
```
yossigo: @filipecosta90 I am a bit concerned that adding such verbose debug logs by default will make the debug output far less usable in most cases. What do you think? Perhaps we need an extra level of verbosity.

shard_connection.cpp
```diff
@@ -542,6 +570,12 @@ void shard_connection::send_set_command(struct timeval* sent_time, const char *k
     push_req(new request(rt_set, cmd_size, sent_time, 1));
 }
 
+void shard_connection::send_empty_rate_limit_event() {
```
yossigo: @filipecosta90 Can you add a description of this mechanism and why it is needed?

run_stats.cpp
filipecosta90: I agree that it's too much of a verbose log. and if people want the hdr output they can simply add the filename parameter and have outputted there. what do you think?

run_stats.cpp
filipecosta90: splitted into two generic methods ( one for each file type ). what do you think?

yossigo: I think it looks much better now!

run_stats.cpp
yossigo: I agree.

memtier_benchmark.cpp
```diff
@@ -1156,6 +1191,16 @@ int main(int argc, char *argv[])
     }
 
     config_init_defaults(&cfg);
+    // We need this here since the the defaults might change the calculations
```
yaacovhazan-Redislabs: 2x the the

memtier_benchmark.cpp
```diff
@@ -1156,6 +1191,16 @@ int main(int argc, char *argv[])
...
+    // The rate of commands per second needs to be higher that the total number of connections
```
yaacovhazan-Redislabs: that --> than?

memtier_benchmark.cpp
```diff
@@ -1156,6 +1191,16 @@ int main(int argc, char *argv[])
...
+    if ( ( cfg.rate_limit_rps > 0 ) && ( cfg.rate_limit_rps < cfg.clients*cfg.threads) ) {
```
yaacovhazan-Redislabs: is the rate-limit implementation force this limitation?
and maybe it should also mentioned in the command line usage

rate_limiter.cpp
```diff
@@ -0,0 +1,48 @@
...
+    ready_to_emit -= from_buffered;
+    return wait;
+}
+
+double RateLimiter::get_rate() const {
+    return 1000000.0 / emission_interval;
+}
+
+bool RateLimiter::set_rate(long long period_secs,double rate) {
+    bool result = false;
```
yaacovhazan-Redislabs: probably I'm missing something.. but where/when result will be equal to true?

run_stats.h
```diff
@@ -28,11 +28,14 @@
...
 #include "run_stats_types.h"
 #include "JSON_handler.h"
+#include "deps/hdr_histogram/hdr_histogram.h"
+#include "deps/hdr_histogram/hdr_histogram_log.h"
 
-
-typedef std::map<float, int> latency_map;
-typedef std::map<float, int>::iterator latency_map_itr;
-typedef std::map<float, int>::const_iterator latency_map_itr_const;
+#define LATENCY_HDR_MIN_VALUE 1
```
yaacovhazan-Redislabs: if the only usage of those #define's are in run_stats.cpp pls consider to move them (to run_stats.cpp)..

shard_connection.h
```diff
@@ -172,6 +174,8 @@ class shard_connection {
     struct bufferevent *m_bev;
     struct event_base* m_event_base;
 
+    bool limited;
```
yaacovhazan-Redislabs: the only usage of this one is in fill_pipeline... why we can't use local variable instead of class's member?

shard_connection.cpp
```diff
@@ -542,6 +572,23 @@ void shard_connection::send_set_command(struct timeval* sent_time, const char *k
...
+ * empty rate limit command:
+ *
+ * we send the empty rate limit command so that when a client has exceed is quota
+ * instead of writing new requests, it will just submit empty event to the event loop.
+ * This is needed since the normal rate limit based on suspending the thread that is
+ * exceeding the quota can affect the other clients on that same thread and the
+ * latency of the request as well.
+ * In this way we are not effecting other clients overall latency measurement.
+ */
+void shard_connection::send_empty_rate_limit_event() {
```
yaacovhazan-Redislabs: if I understood correctly, every time that we want to suspend the requests .. we send error command(s) to the server?
isn't the latency of the command can affect the rate limit?

filipecosta90: @yaacovhazan-Redislabs , we write data to the file descriptor without going to the server at all. We simply write `"^\r\n"` just so that it can be discarded correctly on `void shard_connection::process_response(void)`.

memtier_benchmark.cpp
filipecosta90: given the example command:

```
memtier_benchmark --hide-histogram -n 10 --rate-limit-rps 100
```

Not forcing can lead to situations in which we have a final correct average RPS that will target the described rps, but we will have instantaneous rps of 0 and 200 rps, 0 and 200 again. So we will have bursts followed by seconds without requests in a cyclic pattern. 

```
[RUN #1 8%,  10 secs]  4 threads:        1600 ops,     199 (avg:     159) ops/sec, 12.69KB/sec (avg: 9.77KB/sec),  1.82 (avg:  2.03) msec latencyy
[RUN #1 11%,  17 secs]  4 threads:        2200 ops,       0 (avg:     129) ops/sec, 0.00KB/sec (avg: 7.92KB/sec),   nan (avg:  2.00) msec latencyy
[RUN #1 14%,  22 secs]  4 threads:        2800 ops,     199 (avg:     127) ops/sec, 11.50KB/sec (avg: 7.84KB/sec),  1.88 (avg:  1.98) msec latency
[RUN #1 100%, 194 secs]  0 threads:       20000 ops,       0 (avg:     103) ops/sec, 0.00KB/sec (avg: 6.28KB/sec),   nan (avg:  2.02) msec latency
```
Will mention it on command line usage :)

shard_connection.h
filipecosta90: 💯 agree. corrected in the latest commit.

shard_connection.cpp
```diff
@@ -542,6 +572,23 @@ void shard_connection::send_set_command(struct timeval* sent_time, const char *k
...
+    if ( bufferevent_write(m_bev,"^\r\n",3) == 0 ){
```
yaacovhazan-Redislabs: Please check it again... when we use bufferevent_write we are adding data to the output buffer of m_bev ... means.. writing data on the socket to the server

filipecosta90: you're totally right @yaacovhazan-Redislabs . 
confirmed that yes, we're sending the "rate-limiting-dummy" command to redis.  will correct it 👍

yossigo: @filipecosta90 The libevent `bufferevent` has some built in rate limiting mechanism. It's obviously based on throughput and knows nothing about requests/responses but perhaps it's something you can leverage for this. I don't remember how it works or what APIs are exposed so it may or may not be possible but worth looking.

filipecosta90: hi there @yossigo and @yaacovhazan-Redislabs, sorry for the delayed reply. Considering that the histogram and the rate-limiting features are separate ones and one is stalling other I would go for splitting this PR into three:
- 1st one ( this one ) -  with overall percentile analysis feature
- 2nd one - extended percentile over time ( as discussed with @yossigo on January )
- 3rd one - rate limiting feature using libevent's Bufferevents and Rate-limiting features 
In that manner we would not delay the possibility of having extended metrics on memtier. What do you guys think?

filipecosta90:
Hi there @yossigo and @yaacovhazan-Redislabs I've pruned the PR and I believe it's ready for a new review.
I've updated the hdrhistogram to the latest official one ( v0.11.0 ), added a test case for the added features, and extended the PR to store per second latency/throughput metrics at the output json ( The stored percentiles comes from the `--print-percentiles` property ). Here is a snippet:
```
,"ALL STATS":{
		"Runtime":{
			"Start time": 1597016193374
			,"Finish time": 1597016207579
			,"Total duration": 14205
			,"Time unit": "MILLISECONDS"
			}
		,"Sets":{
			"Count": 182000
			,"Ops/sec": 12812.51
			,"Hits/sec": 0.00
			,"Misses/sec": 0.00
			,"Latency": 1.419
			,"Average Latency": 1.419
			,"Min Latency": 0.032
			,"Max Latency": 14.415
			,"KB/sec": 986.79
			,"Time-Serie":{
				"0":{
					"Count": 14225
					,"Average Latency": 1.29
					,"Min Latency": 0.54
					,"Max Latency": 6.43
					,"p50.00": 1.17
					,"p99.00": 3.73
					,"p99.90": 4.96
					}
				,"1":{
					"Count": 14739
					,"Average Latency": 1.23
					,"Min Latency": 0.10
					,"Max Latency": 7.78
					,"p50.00": 1.20
					,"p99.00": 2.62
					,"p99.90": 5.86
					}
(...)
```
WDYT?

yossigo:
@filipecosta90 Thanks a lot for this effort! It's a great improvement and definitely worth the effort and time invested in it.

PR #107, opened 2019-12-19
Add SNI support

adamflott:
Our setup uses SNI for routing, and this is a quick hack to include it as a command line option.

Feedback welcomed.

### DIFF
--- a/memtier_benchmark.cpp
+++ b/memtier_benchmark.cpp
@@ -92,6 +92,7 @@ static void config_print(FILE *file, struct benchmark_config *cfg)
         "key = %s\n"
         "cacert = %s\n"
         "tls_skip_verify = %s\n"
+        "sni = %s\n"
 #endif
         "out_file = %s\n"
         "client_stats = %s\n"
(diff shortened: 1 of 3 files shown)


memtier_benchmark.cpp
```diff
@@ -839,6 +847,7 @@ void usage() {
             "      --key=FILE                 Use specified private key for TLS\n"
             "      --cacert=FILE              Use specified CA certs bundle for TLS\n"
             "      --tls-skip-verify          Skip verification of server certificate\n"
+            "      --sni=STRING               Add an SNI header\n"
```
ushachar, line 850: @itamarhaber  do you see any reason to toggle this on/off? Why not always add it when doing TLS?

yossigo: @ushachar I think that'll work for the mainstream cases, but there's probably a long tail of cases where you'd want a custom name, specify an IP as a hostname + SNI (e.g. no DNS resolution), etc.

We *could* consider to use the hostname by default (if it's a name and not an IP) if we still have the option to control that manually.

adamflott:
I chatted with one of the TLS RFC contributors and he said you should always include the SNI, but for my use case I need to override the connection hostname with a different value. 

Thanks for reviewing! This is utility is going to see a lot of use at Akamai :)

PR #109, opened 2020-01-09
Add support for Redis 6.x ACL AUTH.

yossigo:


protocol.cpp
```diff
@@ -203,13 +203,53 @@ int redis_protocol::authenticate(const char *credentials)
...
-        "$4\r\n"
-        "AUTH\r\n"
-        "$%u\r\n"
-        "%s\r\n",
-        (unsigned int)strlen(credentials), credentials);
+    /* Credentials may be one of:
+     * <PASSWORD>           For Redis <6.0 simple AUTH commands.
+     * <USER>:<PASSWORD>    For Redis 6.0+ AUTH with both username and password.
+     *
+     * A :<PASSWORD> will be handled as a special case of quoting a password that
```
yaacovhazan-Redislabs, line 210: do we need to think of a case where the USER may contain colon?
what about ask for --user and --password in the command line?

yossigo: Generally we already support the `<USER>:<PASSWORD>` notation on `--authenticate` when supporting binary memcache, so it made more sense to apply that to Redis as well rather than introduce new arguments.

I think not supporting a username with a colon is reasonable. Old-style passwords with a colon are supported (by prepending a colon) so backwards compatibility is maintained.

PR #114, opened 2020-02-27
Initial CI tests integration ( single oss instance, and oss-cluster )

filipecosta90:
This PR makes usage of RLTest and MBdirector to flow test memtier_benchmark, while keeping the minimum possible code duplication effort ( environment setup and memtier_benchmark run are taken care of by those two test dependencies ). 
It tests both OSS single instance and OSS cluster simple SET/GET benchmarks, and asserts that we've reached the DB instances with the exact get/set ratio and overall issued commands.


.github/workflows/ci.yml
```diff
@@ -14,6 +14,25 @@ jobs:
       run: sudo apt-get install autoconf automake pkg-config libevent-dev libpcre3-dev libssl-dev
     - name: Build
       run: autoreconf -ivf && ./configure && make
+    - name: Install test dependencies
+      run: |
+        git clone git://github.com/antirez/redis.git --branch 5.0.7
+        cd redis && make malloc=libc && make && cd ..
```
yossigo: @filipecosta90 Any particular reason you use `malloc=libc`?

.github/workflows/ci.yml
```diff
@@ -14,6 +14,25 @@ jobs:
...
+        ./redis/src/redis-server --version
+        sudo apt-get install python3-pip 
+        sudo apt-get install python3-setuptools
```
yossigo: GitHub Actions has a more direct support for Python envs.
You should use something like:
```
    - name: Setup Python
      uses: actions/setup-python@v1
      with:
        python-version: '3.x'
        architecture: 'x64'
```

Then you can just run `pip install`.
And if you want to take it another step, you should also consider caching some steps to make this run a bit faster (probably not a big deal though).

tests/tests_oss_simple_flow.py
```diff
@@ -0,0 +1,159 @@
+import glob
+import os
+
+from mbdirector.benchmark import Benchmark
+from mbdirector.runner import RunConfig
+
+MEMTIER_BINARY = os.environ.get("MEMTIER_BINARY","memtier_benchmark")
+
+# run each test on different env
```
yossigo: Consider reorganizing this code for better reuse.

.github/workflows/ci.yml
filipecosta90: For this purpose none. We compile with malloc=libc on other CI's since we might want to use valgrind to check for leaks on the modules. On this case no real usage. changing towards default jemalloc

tests/tests_oss_simple_flow.py
filipecosta90: split code into two files. One with the code parts expected to be reused across tests. Can you see if this was what you expected?

.github/workflows/ci.yml
filipecosta90: updated to use direct python support. also caching dependencies as you've stated :)

.github/workflows/ci.yml
```diff
@@ -6,14 +6,94 @@ jobs:
...
+      run: pip install -r tests/test_requirements.txt
+
+    - name: Install Redis Server test dependencies
+      run: |
+        git clone git://github.com/antirez/redis.git --branch unstable
+        cd redis
+        make BUILD_TLS=yes
+        ./utils/gen-test-certs.sh
+        ./src/redis-server --version
+        cd ..
```
yossigo, line 51: I think it would be a good idea to cache also the redis-server binary and the test cert files... but it can be done as a future optimization.

tests/tls/ca.crt
```diff
@@ -0,0 +1,31 @@
+-----BEGIN CERTIFICATE-----
```
yossigo: Why do you include the certs in the repo? For tests it doesn't really matter but it's not a good practice, can trigger false positive security alerts for the repo - and besides we generate them anyway.

filipecosta90: included as a lasy local run ( so that I didnt had to specify their paths :) ). Will remove them :+1:

.github/workflows/ci.yml
filipecosta90: will do so :+1:

yossigo:
Thanks @filipecosta90, that's great progress!

PR #117, opened 2020-03-23
Enable arbitrary commands to be run on cluster mode

filipecosta90:
This PR enables running arbitrary commands on cluster mode and adds the required testing to it. 
Example command:
```
memtier_benchmark -s 10.3.0.54 -p 12002 -c 50 -t 14 \
            --command="hset __key__ f1 v1 f2 v2 f3 v3 f4 v4 f5 v5 f6 v6 f7 v7" \
            --hide-histogram --test-time 180 --run-count 3 --cluster-mode
```

Given it detected the issue in #204 it fixes it as well.


client.cpp
```diff
@@ -76,15 +76,12 @@ bool client::setup_client(benchmark_config *config, abstract_protocol *protocol,
     if ((config->arbitrary_commands->is_defined() && config->arbitrary_commands->at(0).key_pattern == 'P') ||
         (config->key_pattern[key_pattern_set]=='P')) {
         unsigned long long client_index = config->next_client_idx % total_num_of_clients;
-
```
yossigo: @filipecosta90 Please avoid changes that are purely white space (suggest to clean them on final rebase of the PR).

cluster_client.cpp
```diff
@@ -358,9 +358,55 @@ bool cluster_client::get_key_for_conn(unsigned int conn_id, int iter, unsigned l
...
+
+            assert(value != NULL);
+            assert(value_len > 0);
+
+            cmd_size += m_connections[conn_id]->send_arbitrary_command(arg, value, value_len);
+        }
+    }
+
+    m_connections[conn_id]->send_arbitrary_command_end(m_executed_command_index, &timestamp, cmd_size);
+//    m_reqs_generated++;
```
yossigo: Please avoid commented out code as it's ambiguous.

cluster_client.cpp
```diff
@@ -358,9 +358,55 @@ bool cluster_client::get_key_for_conn(unsigned int conn_id, int iter, unsigned l
...
+            cmd_size += m_connections[conn_id]->send_arbitrary_command(arg);
+        } else if (arg->type == key_type) {
+            unsigned long long key_index;
+
+            // get key
+            if (!get_key_for_conn(conn_id,  get_arbitrary_obj_iter_type(cmd, m_executed_command_index), &key_index)) {
+                return;
+            }
+
+            assert(key_index != NULL);
```
yossigo:
```suggestion
            assert(key_index != 0);
```

`key_index` is not a pointer, so `NULL` is wrong but isn't 0 a valid index (making this assertion false?)

yossigo (changes requested):
@filipecosta90 This looks good to me, with a few minor changes (and we'll need to rebase all the commits). In your example above you only specify a single `__key__`, I guess there should be no problem to use an arbitrary number of `__key__` and `__data__` arguments right?

filipecosta90:
> @filipecosta90 This looks good to me, with a few minor changes (and we'll need to rebase all the commits). In your example above you only specify a single `__key__`, I guess there should be no problem to use an arbitrary number of `__key__` and `__data__` arguments right?

Yes, assuming that the keys are all in the same slot ( given that we request for a connection based on the key ). I'll make that change and mention that the first key position will always be the one used by memtier to internal map the command to the proper shard connection. WDYT?

filipecosta90:
> @filipecosta90 This looks good to me, with a few minor changes (and we'll need to rebase all the commits). In your example above you only specify a single `__key__`, I guess there should be no problem to use an arbitrary number of `__key__` and `__data__` arguments right?

@yossigo I've added the multi-placeholder test on the last commit. I believe all issues have been solved and we're ready to review :)

yossigo (dismissed):
@filipecosta90 Had a quick look, seems OK to me.
@YaacovHazan Any other inputs?

YaacovHazan:
@filipecosta90 I see that in the example test you are still using one __key__. Do we support more than one __key__? if yes what user should expect when the __key__'s are mapped to different slots?

filipecosta90:
> @filipecosta90 Looks good. I see that in the example test you are still using one `__key__`. Do we support more than one `__key__`? if yes what user should expect when the `__key__`'s are mapped to different slots?



@YaacovHazan with the following example:
```
memtier_benchmark --command "MSET __key__ v __key__ v" -p 30001 --cluster-mode
```

the expected CROSSSLOT error is replied:


```
server 127.0.0.1:30001 handle error response: -CROSSSLOT Keys in request don't hash to the same slot
```

I believe this is the expected behavior correct?

filipecosta90:
@YaacovHazan I've added an extra test for multi-key placeholders in https://github.com/RedisLabs/memtier_benchmark/pull/117/commits/6ef066aaf11c01fe3a4191f8859ccd2e31df0a7d

on the tests we confirm that the number of requests matches the expected.

YaacovHazan:
@filipecosta90 why did you add "Don't fill key_index_pool for other connections on cluster client" commit, this is by design

filipecosta90:
> @filipecosta90 why did you add "Don't fill key_index_pool for other connections on cluster client" commit, this is by design

@YaacovHazan if we kept the code as is it hanged indefinitely given we were generating random keys that did not belong to that slot and then stopped prior generating any other key for that connection. As soon as I removed the check for total generated keys that included the keys that were buffered we were then issuing more commands that were required. The solution, ( proper one IMHO ) is not to buffer any key that does not belong to that conn. In this manner, we've solved both issues.

filipecosta90:
> @filipecosta90 assuming Cluster Mode with SET/GET (1 key command), the idea is that you as a client have some key generator based on the user requirements.
> 
> Once you generated a key you should deliver it, if currently, you are in a context of a connection that the key does not belong to, it doesn't mean you can't or don't want to send it. You should use it but with the right connection.
> 
> If we were not connection driven, you could think of it as you have some basic component that generates keys and sends them with the right connection based on the current Cluster topology (and not that one connection put keys for another connection).
> 
> Otherwise, you are impacting the randomization of the keys based on the connection (you will use more keys that route to a good connection than a bad one). The implementation in Memtier-Benchmark is that you choose your `key generator` and follow that creates the requests.
> 
> Now for Cluster with Arbitrary-Command (Multiple keys), the current implementation will not work well and will cause the client to hang. As I said at the early stage of this PR, we should think and consider what is the right way to handle that. But IMOH just dropping keys is not the right solution and if yes, maybe only for this case of arbitrary command in Cluster mode.


@YaacovHazan given arbitrary commands won't allow multiple key placeholders as in the current implementation I guess the only missing concern that is still to be addressed is the part of dropping the keys from different slots when not in arbitrary command. If we fix it IMHO we can move forward with this PR without compromising any of the current usage of cluster mode. Agree? Trying to see if we can make this as simple as possible to merge and have this feature. We need this so much! =)

YaacovHazan:
@filipecosta90 ,Not sure I got you about the last concern. I think that once we limit the arbitrary command to one key placeholder, the PR is ok.

Two more notes:
1. I do think that we need to come up with a complete solution and add the ability to have arbitrary commands with more than one key placeholder.
2. I know that @ushachar is working on a different approach to generating keys per connection, and it could be the basic step to achieve that.

protocol.cpp
```diff
@@ -706,17 +706,22 @@ int redis_protocol::write_arbitrary_command(const char *rand_val, int rand_val_l
...
 
         // check arg type
         if (current_arg->data.find(KEY_PLACEHOLDER) != std::string::npos) {
+            key_placeholder_count++;
             if (current_arg->data.length() != strlen(KEY_PLACEHOLDER)) {
                 benchmark_error_log("error: key placeholder can't combined with other data\n");
                 return false;
             }
-
+            if (key_placeholder_count > 1) {
```
YaacovHazan: This should apply only to cluster mode. The non-cluster mode does support more than one key.

YaacovHazan (changes requested):

protocol.cpp
YaacovHazan: Also, _"IMOH just dropping keys is not the right solution and if yes, maybe only for this case of arbitrary command in Cluster mode."_ was relevant for the initial PR where we allowed more than one key.
Since we agree to limited the arbitrary command in cluster mode for one key (for now) there is no need to "play" with the pool

filipecosta90: > This should apply only to cluster mode. The non-cluster mode does support more than one key.

@YaacovHazan both for standalone and cluster versions we don't allow for more than one key placeholder.

memtier_benchmark.cpp
```diff
@@ -1293,6 +1290,11 @@ int main(int argc, char *argv[])
             exit(1);
         }
 
+        // Cluster mode supports only a single key commands
+        if (cfg.cluster_mode && cfg.arbitrary_commands->at(i).keys_count != 1) {
```
filipecosta90: @YaacovHazan we should change from `!=1` to `>1`.
Example of a failing command that should work:
```
$ ./memtier_benchmark --cluster --command="SET key __data__" -p 30001 
error: Cluster mode supports only a single key commands
```
and another keyless one
```
./memtier_benchmark --cluster --command="PUBLISH channel __data__" -p 30001 
error: Cluster mode supports only a single key commands
```

I'm including that example on CI and addressing it.

filipecosta90: @YaacovHazan I've added the test and confirmed behaviour is as expected with the code change.

YaacovHazan (approved):

PR #129, opened 2020-06-11
add print frequency

wanghenshui:
log message is to much when encountered an connect error and I add this option to control print frequency

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -23,7 +23,7 @@ AM_INIT_AUTOMAKE
 AM_SILENT_RULES([yes])
 LT_INIT
 
-CXXFLAGS="-O2 -g -Wall"
+CXXFLAGS="-O2 -g -Wall -static-libgcc -static-libstdc++"
 
 # Checks for programs.
 AC_PROG_CXX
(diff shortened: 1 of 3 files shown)


wanghenshui:
ping @yossigo

yossigo:
@wanghenshui thank you! Just be sure I understand correctly, so you basically want the ability to bump up the print rate to overcome lots of messages produced by `benchmark_*_log()` prints on connection errors, etc.?

Did you consider an option to suppress those messages altogether, possibly narrowing it down to an additional errors counter?

wanghenshui:
> 
> 
> @wanghenshui thank you! Just be sure I understand correctly, so you basically want the ability to bump up the print rate to overcome lots of messages produced by `benchmark_*_log()` prints on connection errors, etc.?
> 
> Did you consider an option to suppress those messages altogether, possibly narrowing it down to an additional errors counter?

we use memtier_benchmark as a stress tool, stress our redis-like server every day, print a log per second is also too much for analyze.

`narrowing it down to an additional errors counter` is also a good idea, maybe i implement it in next PR

Issue #128, opened 2020-05-21
'All Stats' output contains zeros for all the columns except 'Latency'


speddam:
Hi,

The latest memtier_benchmark shows the 'All Stats' output containing all zeros for all the columns except 'Latency', like this:

    $ ./memtier_benchmark --clients=50 --threads=4 --requests=10000 --data-size-range=32-1000 --pipeline=100 --ratio=5:5
    ...
    ALL STATS
    =========================================================================
    Type         Ops/sec     Hits/sec   Misses/sec      Latency       KB/sec 
    -------------------------------------------------------------------------
    Sets            0.00          ---          ---     19.71200         0.00 
    Gets            0.00         0.00         0.00     19.73500         0.00 
    Waits           0.00          ---          ---      0.00000          --- 
    Totals          0.00         0.00         0.00     19.72300         0.00
    ...

I have done the git bisect and it pointed to this commit [0]:

    $ git bisect bad
    bec9cc46079476c8414ab7104516c5a9c966b90f is the first bad commit
    commit bec9cc46079476c8414ab7104516c5a9c966b90f
    Author: YaacovHazan <31382944+YaacovHazan@users.noreply.github.com>
    Date:   Wed Dec 11 17:02:15 2019 +0200

        fix issue #94 cluster mode stuck (#106)
    
        add missing bufferevent_disable for connections that
        complete to send new requests

    :100644 100644 b9fe31c3568e172369897c0eea0111138dcc5e8b 6fbcd856130c8709f8786c97e6b38ad101ca342c M	shard_connection.cpp

I'm not familiar with the code base, but by commenting out the bufferevent_disable() call, I see that the output now shows all the columns:

    ALL STATS
    =========================================================================
    Type         Ops/sec     Hits/sec   Misses/sec      Latency       KB/sec 
    -------------------------------------------------------------------------
    Sets       509744.79          ---          ---     19.52900    281230.68 
    Gets       509744.79       509.74    509235.05     19.53200     20145.57 
    Waits           0.00          ---          ---      0.00000          --- 
    Totals    1019489.58       509.74    509235.05     19.53000    301376.25

---

   Regards,
   Subhash

[0] https://github.com/RedisLabs/memtier_benchmark/commit/bec9cc46079476c8414ab7104516c5a9c966b90f?diff=unified#commitcomment-38744660

rohitjoshi:
I have the same issue and by commenting out the line, it works fine.

```
[...]

PR #130, opened 2020-06-24
fix issue #128 zeros in ouptut

YaacovHazan:
the issue relates to the way each connection disable its events
move that code to one place and handle it separately for read/write

### DIFF
--- a/shard_connection.cpp
+++ b/shard_connection.cpp
@@ -445,11 +445,23 @@ void shard_connection::process_response(void)
         }
     }
 
+    fill_pipeline();
+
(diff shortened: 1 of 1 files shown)


lubaihua33:
After this commit, the memcached test always falls into endless loop. The command is :
Server: memcached -u root -p 21789
Client: memtier_benchmark -s 10.10.10.5 -p 21789 -P memcache_text -x 3 -n 1000000 -t 1 -c 1 -d 4000 --ratio 1:1 --key-pattern S:S

PR #144, opened 2020-10-18
Prevents the program from entering endless loop

hbina:
See https://github.com/RedisLabs/memtier_benchmark/issues/135

The fix is simple enough, we just abort if we detect this particular
setting.

Signed-off-by: Hanif Bin Ariffin <hanif.ariffin.4326@gmail.com>

### DIFF
--- a/shard_connection.cpp
+++ b/shard_connection.cpp
@@ -438,7 +438,7 @@ void shard_connection::process_response(void)
     }
 
     if (m_config->reconnect_interval > 0 && responses_handled) {
-        if ((m_conns_manager->get_reqs_processed() % m_config->reconnect_interval) == 0) {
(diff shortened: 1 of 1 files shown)


yossigo:
Hi @hbina, thanks for this PR. Did you consider fixing the root cause of the problem?

hbina:
From what I understand, this seems like am expected behavior. If I am requesting 5 request and also say I will reset at every 5 request, I will just do this endlessly. Maybe I am wrong.

yossigo:
@hbina reconnect-interval specifies a maximum number of commands to execute on a connection before dropping it and re-establishing a new one. I believe it's just a bug in the implementation, but you are more than welcome to take a look for yourself. Thanks!

hbina:
Okay, I will take a look.

hbina:
Is this correct? I hope I am not messing with other stuff. This does prevent the endless loop

@yossigo

yossigo (approved):

yossigo:
@hbina Thank you, looks good!

PR #147, opened 2020-11-10
Added extra note to avoid unnecessary error

AshishVarshneyy:
To avoid unnecessary error while parsing data file through --import-data.


yossigo (approved):
@AshishVarshneyy thank you!

PR #154, opened 2021-01-14
Enable key placeholder to be combined with other data within the same command argument

### LINKED ISSUE #153: Error in complex arbitrary command
I want to test the performance of RedisGraph with memtier benchmark, and my input is:
``` sh
memtier_benchmark --server 127.0.0.1 --port 6379 -n 1000 -c 1 -t 1 --hide-histogram --command="graph.query GRAPH1 \"MATCH (s:vertex)-[*6]->(t) WHERE id_in_graph(s)=__key__ RETURN count(t)\"" --key-maximum=100000 --distinct-client-seed --command-key-pattern G
```
and the benchmark raised an error:
``` sh
error: key placeholder can't combined with other data
```
So I checked the source code:
``` cpp
            if (current_arg->data.length() != strlen(KEY_PLACEHOLDER)) {
                benchmark_error_log("error: key placeholder can't combined with other data\n");
                return false;
            }
```
Maybe my quoted command is treated as a whole parameter? And how to enable my command?
Thanks

filipecosta90:
Fixes #153 
This PR adds the option to combine key placeholders with other characters within the same argument. This is important for RedisGraph and RediSearch for us to be able to mix for example node ids with other conditions, etc... 
I've used this feature on RG 2.0 benchmark and forgot to open the PR to kick off this feature discussion. 
Example of usage:
```
memtier_benchmark -c 10 -t 1 \
    --command="graph.query GRAPH1 \"MATCH (s:vertex)-[*6]->(t) WHERE id_in_graph(s)=__key__ RETURN count(t)\"" \
    --key-maximum=100000 --distinct-client-seed --command-key-pattern G --key-prefix=""
```
This PR includes two extra tests


protocol.cpp
```diff
@@ -631,11 +631,6 @@ bool redis_protocol::format_arbitrary_command(arbitrary_command &cmd) {
 
         // check arg type
         if (current_arg->data.find(KEY_PLACEHOLDER) != std::string::npos) {
```
YaacovHazan, line 714: do we want to allow two or more `__key__`, for ex: "--command "SET bla__key__bla__key__bla value"?

client.cpp
```diff
@@ -256,8 +257,14 @@ void client::create_arbitrary_request(const arbitrary_command* cmd, struct timev
 
             assert(key != NULL);
             assert(key_len > 0);
-
-            cmd_size += m_connections[conn_id]->send_arbitrary_command(arg, key, key_len);
+            //when we have static data mixed with the key placeholder
+            if (arg->data.length() != strlen(KEY_PLACEHOLDER)) {
+                std::string str (arg->data);
```
YaacovHazan: I don't think this is good idea to do this str.replcae in "run time", on every request that we send.
I think we can preapre the prefix and suffix of the __key__ before and then just send them before and after. what do you think?

YaacovHazan (changes requested):

protocol.cpp
filipecosta90: @YaacovHazan  I believe we need it we should address it on another PR.

client.cpp
filipecosta90: @YaacovHazan I've followed your review and we're now using recalculated prefix and suffix strings ( only when required ).

YaacovHazan: Now that key_type is composed by prefix + key + suffix what do you think about something like this:

```
if (prefix>0)
   send_arbitrary_command(prefix)

send_arbitrary_command(key)

if (suffix>0)
   send_arbitrary_command(suffix)
```

YaacovHazan (changes requested):

protocol.cpp
YaacovHazan: Agree, but let's verify it, to avoid confusion now that we allow more than just __key__

client.cpp
filipecosta90: @YaacovHazan  the approach of:
```
if (prefix>0)
   send_arbitrary_command(prefix)
send_arbitrary_command(key)
if (suffix>0)
   send_arbitrary_command(suffix)
would make us send 3 bulk strings ( when we should send 1 )
```
I believe it's not feasible to be done that way
notice the `$%d\r\n` sent each time we call send_arbitrary_command

client.cpp
```diff
@@ -286,7 +288,47 @@ bool client::create_arbitrary_request(unsigned int command_index, struct timeval
...
+            if (arg->has_key_affixes) {
+                // Pre-calculate total length to avoid reallocations
+                const char* key = m_obj_gen->get_key();
+                unsigned int key_len = m_obj_gen->get_key_len();
+                size_t prefix_len = arg->data_prefix.length();
+                size_t suffix_len = arg->data_suffix.length();
+                size_t total_len = prefix_len + key_len + suffix_len;
+
+                // Optimization: use stack buffer for small keys to avoid heap allocation
+                constexpr size_t STACK_BUFFER_SIZE = 512;
```
paulorsousa: suggestion: move `STACK_BUFFER_SIZE` to a macro (?)

paulorsousa (approved):

client.cpp
fcostaoliveira: addressed in https://github.com/RedisLabs/memtier_benchmark/pull/154/commits/b5f5df6e71e05c01746665af9966beea8b52d301

PR #166, opened 2021-11-25
Initial support for keys following zipfian

lfsu-ali:
This patch allows memtier generate keys that follows zipfian
distribution. An additional parameter --key-zipf-exp is introduced,
meaning P(Key = n) ~ n^{-exp}, which is bounded to (0, 5) to be sane.
The range of keys are limited to positive in this version.

Signed-off-by: Su Lifan <su-lifan@linux.alibaba.com>

### DIFF
--- a/client.h
+++ b/client.h
@@ -124,6 +124,8 @@ public:
             return OBJECT_GENERATOR_KEY_RANDOM;
         } else if (cfg->key_pattern[index] == 'G') {
             return OBJECT_GENERATOR_KEY_GAUSSIAN;
+        } else if (cfg->key_pattern[index] == 'Z') {
+            return OBJECT_GENERATOR_KEY_ZIPFIAN;
         } else {
             if (index == key_pattern_set)
                 return OBJECT_GENERATOR_KEY_SET_ITER;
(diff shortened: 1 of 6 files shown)


tests/tests_oss_zipfian_distribution.py
```diff
@@ -0,0 +1,96 @@
...
+    # Verify that higher exponents lead to more skewed distributions
+    sorted_exponents = sorted(zipf_exponents)
+    for exp1, exp2 in pairwise(sorted_exponents):
+        dist1 = distributions[exp1]
+        dist2 = distributions[exp2]
+
+        concentration1 = calculate_concentration_ratio(dist1)
+        concentration2 = calculate_concentration_ratio(dist2)
+
+        # Higher exponent should have higher concentration in top keys
```
filipecosta90, line 95: > Higher exponent should have higher concentration in top keys

lets add this to the description of `--key-zipf-exp` help section.

filipecosta90 (changes requested):
small request for change @paulorsousa . Given your tests clearly explain the --key-zipf-exp implications I believe we should also make sure the user gets it when doing the `--help`, meaning please add your extra detail on --key-zipf-exp description. 

PS: really cool testing strategy!!! Kudos!!!

tests/tests_oss_zipfian_distribution.py
paulorsousa: Excellent idea! Thank you!! 🙂

PR #168, opened 2022-02-02
Enable OpenSSL to load a configuration file during initialization

slice4e:
Enable OpenSSL to load a configuration file using the standard openssl.cnf file. 
In this way, we can change the default OpenSSL ENGINE, as well as configure other capabilities. 
This aligns memtier benchmark to a similarly added capability in Redis here : https://github.com/redis/redis/pull/8143

By providing this capability to memtier benchmark, we can ensure that memtier benchmark and redis are using the same OpenSSL configuration (e.g. ENGINE) while performing a test.

### DIFF
--- a/memtier_benchmark.cpp
+++ b/memtier_benchmark.cpp
@@ -34,6 +34,8 @@
 #include <sys/resource.h>
 
 #ifdef USE_TLS
+#include <openssl/crypto.h>
+#include <openssl/conf.h>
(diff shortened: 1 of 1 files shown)


filipecosta90 (approved):

yossigo (approved):

yossigo:
Thanks @slice4e!

Issue #171, opened 2022-03-06
Generating value buffer randomly looks like wrong


caipengbo:
I took a look at the random value generation code and felt something was wrong, `buf2_idx` might index out of bounds:
https://github.com/RedisLabs/memtier_benchmark/blob/01f3b1a1c59ef84040947ecb3e983ff231398b9f/obj_gen.cpp#L242-L266

YaacovHazan:
@caipengbo thanks. you are right. do you want to open a PR with a solution?

caipengbo:
Yes, I'm interested in it. I'll fix it as soon as possible.

caipengbo:
Do you know why `memtier-benchmark` used this method to generate random buffers?  What is its real intention?
Why use two temporary buffers(`buf1` and `buf2`)? @YaacovHazan

YaacovHazan:
@caipengbo this buffer is used for the `--random-data` option, to generate random data (value) for the commands that `memtier-benchmark` sent. I assume that it use two buffers (`buf1 ^ buf2 ^ Iter`) to maximize the randomization

caipengbo:
I guess it's also intended to maximize randomization and reduce the number of times read the `/dev/urandom` file.

PR #172, opened 2022-03-11
Correctly randomize value buffer

caipengbo:
The current logic for randomizing value_buffer looks like something's wrong, and I've tried to fix it.

I guess the original intent was to read the `/dev/urandom` file every `20*64*64` bytes, and I modified it to follow this logic.

Related issue #171.

### DIFF
--- a/obj_gen.cpp
+++ b/obj_gen.cpp
@@ -239,26 +239,29 @@ void object_generator::alloc_value_buffer(void)
 
             char buf1[64] = { 0 };
             char buf2[64] = { 0 };
-            unsigned int buf1_idx = sizeof(buf1);
-            unsigned int buf2_idx = sizeof(buf2);
+            unsigned int buf1_idx = 0;
+            unsigned int buf2_idx = 0;
             char *d = m_value_buffer;
-            int ret;
             int iter = 0;
-
+            int ret = read(m_random_fd, buf1, sizeof(buf1));
+            assert(ret > -1);
+            ret = read(m_random_fd, buf2, sizeof(buf2));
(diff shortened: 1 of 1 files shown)


obj_gen.cpp
```diff
@@ -239,26 +239,29 @@ void object_generator::alloc_value_buffer(void)
...
             char buf2[64] = { 0 };
-            unsigned int buf1_idx = sizeof(buf1);
-            unsigned int buf2_idx = sizeof(buf2);
+            unsigned int buf1_idx = 0;
+            unsigned int buf2_idx = 0;
             char *d = m_value_buffer;
-            int ret;
             int iter = 0;
-
+            int ret = read(m_random_fd, buf1, sizeof(buf1));
```
YaacovHazan: Not sure I like the duplication of that code... IIUC the only issue with the original code is that in the first iteration we do not initialize the buffers because of the `if (iter == 20)`.
I assume that we can just initialize `iter` to `20` and modify the `if` statement to check for `>= 20` (like we do with (`buf2_idx`).
Alternatively, we can just modify the `if` to check for ` % 20 == 0` and move the increment of `iter` to be after the `if` as you suggested.

caipengbo: Yes, I made some changes according to your suggestions

YaacovHazan:
@caipengbo LGTM, did you test it?

caipengbo:
@YaacovHazan  Yeah, I've tested it :)

YaacovHazan (approved):

caipengbo:
CI seems to have some problems.

PR #176, opened 2022-04-20
Moved integration testing logic out of github actions and included testing section on Readme

filipecosta90:
This PR clarifies how to test memtier_benchmark locally and uses the exact same script for CI testing. 

Details on how to to test memtier:

The project includes a basic set of integration tests.

**Integration tests**

Integration tests are based on [RLTest](https://github.com/RedisLabsModules/RLTest), and specific setup parameters can be provided
to configure tests and topologies (OSS standalone and OSS cluster). By default the tests will be ran for all common commands, and with OSS standalone setup.

To run all integration tests in a Python virtualenv, follow these steps:

    $ mkdir -p .env
    $ virtualenv .env
    $ source .env/bin/activate
    $ pip install -r tests/test_requirements.txt
    $ ./tests/run_tests.sh

To understand what test options are available simply run:

    $ ./tests/run_tests.sh --help

### DIFF
--- a/tests/gen-test-certs.sh
+++ b/tests/gen-test-certs.sh
@@ -1,0 +1,58 @@
+#!/bin/bash
+
+# Generate some test certificates which are used by the regression test suite:
+#
+#   tests/tls/ca.{crt,key}          Self signed CA certificate.
+#   tests/tls/redis.{crt,key}       A certificate with no key usage/policy restrictions.
+#   tests/tls/client.{crt,key}      A certificate restricted for SSL client usage.
+#   tests/tls/server.{crt,key}      A certificate restricted for SSL server usage.
+#   tests/tls/redis.dh              DH Params file.
+
+generate_cert() {
+    local name=$1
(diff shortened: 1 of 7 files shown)


tests/gen-test-certs.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash
```
yossigo, line 1: This comes from the Redis source tree no, perhaps we can re-use that?

filipecosta90: I wanted to avoid depending on extra scripts. Assuming we don't need to clone redis ( we only do it in CI ) there is no need for any memtier developer to clone redis just to get the script. agree?

filipecosta90: After discussing with @yossigo this subject, we went a step further ron the simplification and we're now using redis-server from installed apt meaning no need to clone or build anything else apart from memtier on CI.

yossigo (approved):

PR #182, opened 2022-06-21
add support for resp3

YaacovHazan:
- add new command line argument, '--resp', to send HELLO command
  at the begining of a connection to set the RESP mode: 2 or 3.
- the redis protocol component updated to support the new RESP 3 types.
  exceptional are:
  - Push type: Maybe will be added on next phase, but need to understand better
    how and what does it mean to benchmark Redis with push commands.
  - Streamed strings & Streamed aggregated: Currently not used by Redis.

### DIFF
--- a/memtier_benchmark.cpp
+++ b/memtier_benchmark.cpp
@@ -352,6 +352,7 @@ static int config_parse_args(int argc, char *argv[], struct benchmark_config *cf
     enum extended_options {
         o_test_time = 128,
         o_ratio,
+        o_resp,
         o_pipeline,
         o_data_size_range,
         o_data_size_list,
@@ -400,6 +401,7 @@ static int config_parse_args(int argc, char *argv[], struct benchmark_config *cf
         { "port",                       1, 0, 'p' },
         { "unix-socket",                1, 0, 'S' },
         { "protocol",                   1, 0, 'P' },
+        { "resp",                       1, 0, o_resp },
 #ifdef USE_TLS
         { "tls",                        0, 0, o_tls },
         { "cert",                       1, 0, o_tls_cert },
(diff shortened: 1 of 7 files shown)


memtier_benchmark.cpp
```diff
@@ -400,6 +401,7 @@ static int config_parse_args(int argc, char *argv[], struct benchmark_config *cf
         { "port",                       1, 0, 'p' },
         { "unix-socket",                1, 0, 'S' },
         { "protocol",                   1, 0, 'P' },
+        { "resp",                       1, 0, o_resp },
```
yossigo: Did you consider extending `--protocol` instead of adding another flag?

protocol.cpp
```diff
@@ -253,6 +261,21 @@ int redis_protocol::authenticate(const char *credentials)
     return size;
 }
 
+int redis_protocol::configure_protocol(enum PROTOCOL_CONFIGURATION conf) {
+    int size = 0;
+    if (conf == PROTOCOL_CONF_RESP2 || conf == PROTOCOL_CONF_RESP3) {
+        m_resp3 = conf == PROTOCOL_CONF_RESP3;
+        size = evbuffer_add_printf(m_write_buf,
```
yossigo: Do we want to maintain a backwards compatible mode where we don't send `HELLO` and we assume we're using RESP2? Without it, we'd lose compatibility with Redis <6.0.

protocol.cpp
```diff
@@ -403,6 +426,48 @@ int redis_protocol::write_command_wait(unsigned int num_slaves, unsigned int tim
     return size;
 }
 
+bool redis_protocol::aggregate_type(char c) {
```
yossigo, line 429: I wonder if these additional branches are going to have some measurable impact on performance, being executed in tight loops. If that's the case, maybe we'd want to subclass `redis_protocol` into V2 and V3 instead.

memtier_benchmark.cpp
YaacovHazan: Yes, but for me, it seems more internal configuration under the Redis protocol. and for backport compatibility, we probably need to add another 2 options (`redis-resp2`, `redis-resp3`) as the existing one refers to not set `resp` version at all.

protocol.cpp
YaacovHazan: Not sure I got it. In the current implementation if the user doesn't add `--resp` he will get the old behavior (will not get into this piece of code).

protocol.cpp
YaacovHazan: Well not sore it is noticeable and that simple benchmark (of memtier-benchmark) will show us something.
regarding your suggestion to subclass it, do you mean that V2 will not have the impact on the performance as V3?
@filipecosta90 WDYT about this performance concern?

yossigo (approved):

PR #188, opened 2022-07-29
Fixed memory usage increasing over time due to per second histogram usage

### LINKED ISSUE #157: Memory usage increasing over time
Hi,

I have observed a behavior that the RSS of memtier_benchmark increases over time during tests on Redis. I found that for each client (connection), the RSS increases by 26.2 KB per second.

The test setup is as follows:

OS: Debian-based Linux

1. Start redis (following https://redis.io/download#from-source-code):
```
src/redis-server
```

2. Run memtier_benchmark:
```
memtier_benchmark -c 10 -t 10 -d 500 --test-time=600 --hide-histogram
```

3. Collect RSS data of memtier_benchmark (on a different tmux tab):
```
echo > /tmp/tmp; for i in `seq 1 300`; do echo -e "$i,`ps --no-headers -o rss -p $(pidof memtier_benchmark)`" | tee -a /tmp/tmp; sleep 1s; done
```

This is the plot of the RSS data for four combinations of threads/clients over the first 5 minutes:

Screen Shot 2021-05-07 at 4 21 48 PM
Screen Shot 2021-05-07 at 4 22 40 PM

I have checked the following:
* Valgrind shows that there is no memory leak.
* The output of `ss -m` shows that the socket buffer doesn't increase.

Has anyone seen this behavior?

I'll dig into source code more to see whether I can find anything. But appreciated it if anyone has an answer on top of head. Thanks!

filipecosta90:
Fixes #157 . 
Instead of keeping the raw histograms per second, we keep the computed percentiles that we need to store at the end of the benchmark. 

Using @weisiw script to check the memory we can confirm that the issue is gone 
image

### DIFF
--- a/run_stats.cpp
+++ b/run_stats.cpp
@@ -117,6 +117,7 @@ run_stats::run_stats(benchmark_config *config) :
 {
     memset(&m_start_time, 0, sizeof(m_start_time));
     memset(&m_end_time, 0, sizeof(m_end_time));
+    quantiles_list = config->print_percentiles.quantile_list;
 
     if (config->arbitrary_commands->is_defined()) {
         setup_arbitrary_commands(config->arbitrary_commands->size());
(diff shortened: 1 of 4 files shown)


filipecosta90:
As discussed with @yaacovhazan-Redislabs sharing the throughput over time for both master and this PR to confirm there is no change/no affected logic due to the changes contained in it. 

image

YaacovHazan (approved):

PR #206, opened 2023-02-07
Simplify and improve key generation when working in cluster mode.

ushachar:
Instead of randomly generating keys and storing them per shard until we get a key which matches our desired shard, generate one which will always match.
This improved performance and reduces memory overhead when working with a lot of primaries.


filipecosta90:
@ushachar / @YaacovHazan can we merge #203 before this one to ensure all code added here is covered on the flow tests. Assuming it is but worth the double check. agree?

ushachar:
> @ushachar / @YaacovHazan can we merge #203 before this one to ensure all code added here is covered on the flow tests. Assuming it is but worth the double check. agree?

Sure -- +1'ed #203 
This PR doesn't add any tests - but should slightly improve coverage since it removed a big chunk of code.
note that I am working on a test around the handling of MOVED.

ushachar:
@filipecosta90 rebased to include coverage

filipecosta90 (approved):

cluster_client.cpp
```diff
@@ -309,53 +290,13 @@ bool cluster_client::hold_pipeline(unsigned int conn_id) {
...
-        if (key_idx_pool->size() < KEY_INDEX_QUEUE_MAX_SIZE) {
-            key_idx_pool->push(*key_index);
-            m_reqs_generated++;
-        }
-
-        // don't exceed requests
-        if (m_config->requests > 0 && m_reqs_generated >= m_config->requests)
-            return false;
-    }
+    *key_index = m_obj_gen->get_key_index(iter);
```
filipecosta90, line 294: @ushachar in cluster mode we're growing the keyspace range to above the max/min range. This is because we're using the crc16 + key index, meaning we can now have #primaries * keyspace range. The PR #208 includes tests to track this.

filipecosta90 (changes requested):

PR #207, opened 2023-02-08
Always print out stderr/out output on test failures.

ushachar:
Also do some minor cleanups around unused args.


filipecosta90 (approved):
LGTM. Thank you @ushachar !

PR #208, opened 2023-02-13
Include keyspace range checks on flow tests

filipecosta90:
Up until now we didn't really checked the keyspace range after write tests. This PR introduces it. 
I noticed that due to the PR #206 having a but and precisely increasing the keyspace range.

### DIFF
--- a/tests/include.py
+++ b/tests/include.py
@@ -59,13 +59,15 @@ def debugPrintMemtierOnError(config, env):
                     env.debugPrint(line.rstrip(), True)
 
 
-def get_expected_request_count(config):
+def get_expected_request_count(config, key_minimum=0, key_maximum=1000000):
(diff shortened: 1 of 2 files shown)


tests/include.py
```diff
@@ -36,39 +39,55 @@ def add_required_env_arguments(benchmark_specs, config, env, master_nodes_list):
         config['redis_process_port'] = master_nodes_list[0]['port']
 
 
-def debugPrintMemtierOnError(config, env):
```
ushachar: @filipecosta90 I think you reverted the prev. commit by mistake.

filipecosta90: @ushachar you're absolutely right. fixed now :)

YaacovHazan (approved):

Issue #211, opened 2023-04-14
Large key-median value causes error even though it is between key-minimum and key-maximum


clkbug:
When a value exceeding 32 bits is set to `key-median`, the value will not be read correctly.

For example, 

```bash
$ ./memtier_benchmark --key-pattern=G:G --key-minimum=10000000000 --key-maximum=30000000000 --key-median=20000000000
error: key-median must be between key-minimum and key-maximum.
```

(10000000000 is 34-bit value)

`key-minimum` and `key-maximum` are read correctly, but `key-median` is not.

PR #212, opened 2023-04-14
Fix issue #211. Allow key_stddev and key_median options to be set values

clkbug:
I have fixed issue 211.
This changed key_stddev and key_median to read as unsigned long long.
In real-world use cases, the key ranges will be wide enough that the fractional portions will not matter.

### DIFF
--- a/memtier_benchmark.cpp
+++ b/memtier_benchmark.cpp
@@ -694,7 +694,7 @@ static int config_parse_args(int argc, char *argv[], struct benchmark_config *cf
                     break;
                 case o_key_stddev:
                     endptr = NULL;
-                    cfg->key_stddev = (unsigned int) strtof(optarg, &endptr);
+                    cfg->key_stddev = strtod(optarg, &endptr);
                     if (cfg->key_stddev<= 0 || !endptr || *endptr != '\0') {
                         fprintf(stderr, "error: key-stddev must be greater than zero.\n");
                         return -1;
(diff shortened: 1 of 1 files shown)


memtier_benchmark.cpp
```diff
@@ -694,15 +694,15 @@ static int config_parse_args(int argc, char *argv[], struct benchmark_config *cf
                     break;
                 case o_key_stddev:
                     endptr = NULL;
-                    cfg->key_stddev = (unsigned int) strtof(optarg, &endptr);
+                    cfg->key_stddev = strtoull(optarg, &endptr, 10);
```
oranagra: I don't like the loss of fractional part. Specifically in stddev.
But anyway I think the problem is just the wrong use of 32bit float instead of double (although even 32bit float can hold values much larger than 32bit integer).
I.e. `strtod` should solve it and keep both thr fractional part and a huge range.

oranagra: i missed the fact that we had casting (`(unsigned int)`), so we didn't have fractional part anyway.
but that casting is actually the problem (not the `strtof`), so it was maybe enough to just drop the casting.
but if we change these lines, let's got with `strtod` (changing to `strtoull` would be a breaking change, since old commands that used to work, will now error)

clkbug: I understand.
I have changed it to use strtod.

oranagra (approved):

YaacovHazan:
@clkbug Thank you

PR #215, opened 2023-04-20
Support forcing of IPv4/IPv6 resolution.

YaacovHazan:
Add new --ipv4 (-4) and --ipv6 (-6) command line options to force address resolution.

The default value is unspecified, and we will try resolving both.

### DIFF
--- a/memtier_benchmark.cpp
+++ b/memtier_benchmark.cpp
@@ -99,6 +99,7 @@ static void config_print(FILE *file, struct benchmark_config *cfg)
         "server = %s\n"
         "port = %u\n"
         "unix socket = %s\n"
+        "resolution %s\n"
         "protocol = %s\n"
 #ifdef USE_TLS
         "tls = %s\n"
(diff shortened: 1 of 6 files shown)


memtier_benchmark.cpp
```diff
@@ -203,6 +205,7 @@ static void config_print_to_json(json_handler * jsonhandler, struct benchmark_co
     jsonhandler->write_obj("server"            ,"\"%s\"",      	cfg->server);
     jsonhandler->write_obj("port"              ,"%u",          	cfg->port);
     jsonhandler->write_obj("unix socket"       ,"\"%s\"",      	cfg->unix_socket);
+    jsonhandler->write_obj("resolution"        ,"\"%s\"",      	cfg->resolution == AF_UNSPEC ? "Unspecified" : cfg->resolution == AF_INET ? "AF_INET" : "AF_INET6");
```
yossigo: The name `resolution` is a bit unclear in this context, I suggest to refer to it as "address family".

yossigo (approved):

yossigo (approved):

PR #218, opened 2023-05-03
Supporting multi platform dockers

chayim:
In theory, combining this directory structure with the existing webhook should publish a docker. Ideally we might extend the dockerhub build side.

### DIFF
--- a/dockers.yml
+++ b/dockers.yml
@@ -1,0 +1,50 @@
+---
+
+name: Build dockers
+
+on:
+  push:
+    branches:
+      - master
+    tags:
+      - '*'
+
+  workflow_dispatch:
+
+jobs:
+
+  docker:
+    name: Build dockers
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: docker/setup-qemu-action@v2
+      - uses: docker/setup-buildx-action@v2
+        with:
+          install: true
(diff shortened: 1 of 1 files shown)


dockers.yml
```diff
@@ -0,0 +1,50 @@
...
+      - uses: docker/setup-qemu-action@v2
+      - uses: docker/setup-buildx-action@v2
+        with:
+          install: true
+          platforms: linux/amd64,linux/arm64/v8,linux/arm/v7
+
+      - uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
```
yossigo, line 32: @chayim Did you consider using an environment for secrets to reduce the chances of leakage? Like we do [here](https://github.com/RedisLabs/memtier_benchmark/blob/a1821dfd705c444b0f7ee1db8bd4f02e345e709e/.github/workflows/release.yml#L47)

chayim: I did.. but I'd argue this doesn't do what we think. Environments allow for different credentials on a per environment basis, with the same name. The kicker is that since this job would always use this environment - the attack surface is the same.

If we're going to use multiple environments, I solidly agree. WDYT?

yossigo (approved):

PR #221, opened 2023-05-15
Restoring docker build

chayim:
It appears that github actions did not kick off, due to the inclusion of a tags specifier, in the push definition. This PR removes that.

Similarly, it appears that this is somehow not merged into master, though the git log shows it is. Hence, the PR.


yossigo (approved):

filipecosta90:
@yossigo and @chayim merging.

PR #234, opened 2023-09-12
Enable specifying the tls protocol version to use via --tls-protocols argument

filipecosta90:
By default it will always try to use v1.2 and v1.3. 
It allows specifying multiple versions via comma delimiter. 

# Examples:

## Using either v1.1 or v1.2
```
./memtier_benchmark   --cert ~/redislabs/redis/tests/tls/redis.crt     --key ~/redislabs/redis/tests/tls/redis.key --cacert ~/redislabs/redis/tests/tls/ca.crt  -c 1 -t 1 --tls --hide-histogram --tls-protocols tlsv1.1,tlsv1.2
```

Confirmation of data capture that indeed we're using v1.2:
image

## Using only v1.3
```
./memtier_benchmark   --cert ~/redislabs/redis/tests/tls/redis.crt     --key ~/redislabs/redis/tests/tls/redis.key --cacert ~/redislabs/redis/tests/tls/ca.crt  -c 1 -t 1 --tls --hide-histogram --tls-protocols tlsv1.3
```

Confirmation of data capture that indeed we're using v1.3:
image

## Trying to use an non allowed version by the server (v1.1) -- check error

```
$ ./memtier_benchmark   --cert ~/redislabs/redis/tests/tls/redis.crt     --key ~/redislabs/redis/tests/tls/redis.key --cacert ~/redislabs/redis/tests/tls/ca.crt  -c 1 -t 1 --tls --hide-histogram --tls-protocols tlsv1.1 -n 2
Writing results to stdout
[RUN #1] Preparing benchmark client...
[RUN #1] Launching threads now...
TLS connection error: no protocols available
TLS connection error: (null)
[RUN #1 0%,   0 secs]  0 threads:           0 ops,       0 (avg:       0) ops/sec, 0.00KB/sec (avg: 0.00KB/sec),  0.00 (avg:  0.00) msec latency

1         Threads
1         Connections per thread
2         Requests per client


ALL STATS
============================================================================================================================
Type         Ops/sec     Hits/sec   Misses/sec    Avg. Latency     p50 Latency     p99 Latency   p99.9 Latency       KB/sec 
----------------------------------------------------------------------------------------------------------------------------
Sets            0.00          ---          ---             ---             ---             ---             ---         0.00 
Gets            0.00         0.00         0.00             ---             ---             ---             ---         0.00 
Waits           0.00          ---          ---             ---             ---             ---             ---          --- 
Totals          0.00         0.00         0.00            -nan         0.00000         0.00000         0.00000         0.00 

```

### DIFF
--- a/memtier_benchmark.cpp
+++ b/memtier_benchmark.cpp
@@ -39,6 +39,19 @@
 #include <openssl/ssl.h>
 #include <openssl/err.h>
 #include <openssl/rand.h>
+
+#define REDIS_TLS_PROTO_TLSv1       (1<<0)
+#define REDIS_TLS_PROTO_TLSv1_1     (1<<1)
+#define REDIS_TLS_PROTO_TLSv1_2     (1<<2)
+#define REDIS_TLS_PROTO_TLSv1_3     (1<<3)
+
+/* Use safe defaults */
+#ifdef TLS1_3_VERSION
+#define REDIS_TLS_PROTO_DEFAULT     (REDIS_TLS_PROTO_TLSv1_2|REDIS_TLS_PROTO_TLSv1_3)
+#else
+#define REDIS_TLS_PROTO_DEFAULT     (REDIS_TLS_PROTO_TLSv1_2)
+#endif
+
 #endif
 
 #include <stdexcept>
@@ -296,6 +309,8 @@ static void config_init_defaults(struct benchmark_config *cfg)
         cfg->hdr_prefix = "";
     if (!cfg->print_percentiles.is_defined())
         cfg->print_percentiles = config_quantiles("50,99,99.9");
+    if (!cfg->tls_protocols)
+        cfg->tls_protocols = REDIS_TLS_PROTO_DEFAULT;
 }
 
 static int generate_random_seed()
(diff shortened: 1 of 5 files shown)


memtier_benchmark.cpp
```diff
@@ -903,6 +948,7 @@ void usage() {
             "      --key=FILE                 Use specified private key for TLS\n"
             "      --cacert=FILE              Use specified CA certs bundle for TLS\n"
             "      --tls-skip-verify          Skip verification of server certificate\n"
+            "      --tls-protocols            Specify the tls protocol version to use, comma delemited. Use a combination of 'TLSv1', 'TLSv1.1', 'TLSv1.2' and 'TLSv1.3'"
```
YaacovHazan, line 951: Doesn't `--tls-versions` will be more correct?

YaacovHazan (changes requested):

memtier_benchmark.cpp
filipecosta90: @YaacovHazan i've tried to keep the same config name from redis. https://github.com/redis/redis/blob/unstable/redis.conf#L268

YaacovHazan (approved):

PR #237, opened 2023-10-15
Add rate limiting to control the number of requests per second

YaacovHazan:
A new 'rate-limiting' option was added to control the number of request per second.

The rate limiting is based on the 'Token Bucket' algorithm, and according to the configured rate, on each interval, a "new" amount of requests allowed to be sent to the server.

The rate-limiting is at the connection level. Therefore, in cluster mode, the limitation is for each shard, so if, for example, the cluster has three shards and the user configured one request per second. On every second, memtier-benchmark will send three requests, one for each shard.


filipecosta90 (approved):
LGTM @YaacovHazan . I've tested this manually over the last days and I've also included tests for testing the rate-limiting expected request count.

Issue #238, opened 2023-10-23
Error reported when running with "--data-import"


DarrenJiang13:
# Problem
Hi guys, when I run the newest master branch with "--data-import", an error happened:
```
memtier_benchmark: obj_gen.cpp:438: const char* object_generator::get_value(long long unsigned int, unsigned int*): Assertion `0' failed.
```

And I found this is brought by commit https://github.com/RedisLabs/memtier_benchmark/commit/1c6735f06d5fe63a55b671860baf7b858de5551b , before this commit "--data-import" worked OK.

# Discussion
IMHO the problem should be in `client::create_set_request()`.  I change it like this and things go right. But I am not sure this is the right solution
![image](https://github.com/RedisLabs/memtier_benchmark/assets/29703715/ad10cc4d-ccdc-483d-b5ff-b17f88c42294)

YaacovHazan:
@DarrenJiang13 , thanks for reporting this bug.
You are right, the `--data-import` broke by commit https://github.com/RedisLabs/memtier_benchmark/commit/1c6735f06d5fe63a55b671860baf7b858de5551b, but unfortunately the fixed is not that simple.

I will work on a fix and update.

PR #240, opened 2023-10-31
see #238, data import broken

YaacovHazan:
The PR includes some refactoring and cleanup around the following:
object_generator
- generate_key(), a new function that generates a key and stores it
   inside internal buffer (no need for a buffer at client level)

import_object_generator
- Align it to object_generator APIs of getting key/value/expire
- read_next_item - new function for reading the next item before
   creating SET command
- read_next_key - new function for reading (pointing) to the next
   key before creating the GET command

data_object
- Removed.
   We were already getting key/value/expire separately. Now that
   data-import and verify-data also moved to do it like that,
   we could remove it completely.

verify_client
- Unify the craate_request with its base class, and derive the
   create_x_request functions.

### DIFF
--- a/client.cpp
+++ b/client.cpp
@@ -256,7 +256,14 @@ get_key_response client::get_key_for_conn(unsigned int command_index, unsigned i
         iter = obj_iter_type(m_config, command_index);
 
     *key_index = m_obj_gen->get_key_index(iter);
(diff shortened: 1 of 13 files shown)


filipecosta90 (dismissed):
LGTM

filipecosta90 (dismissed):

filipecosta90 (approved):

PR #264, opened 2024-08-28
Fix a bug when cluster mode and limited rate are used at the same time

### LINKED ISSUE #94: memtier_benchmark doest not exit as expected
In my test enviroment, I got the following outputs:
$ ./memtier_benchmark -s 192.168.111.126 -p 2001 -n 1 -c 1 -t 2 -d 20000 --cluster-mode
[RUN #1] Preparing benchmark client...
[RUN #1] Launching threads now...
[RUN #1 100%,   0 secs]  2 threads:           2 ops,    6557 (avg:    6557) ops/sec, 125.38MB/sec (avg: 125.38MB/sec),  0.39 (avg:  0.39) msec latency

It will hang here like forever. By reading the code, it seems it should exit and output like this:
[RUN #1] Preparing benchmark client...
[RUN #1] Launching threads now...
[RUN #1 100%,   0 secs]  0 threads:           2 ops,       0 (avg:    6756) ops/sec, 0.00KB/sec (avg: 129.20MB/sec),  0.00 (avg:  0.37) msec latency

2         Threads
1         Connections per thread
1         Requests per client

ALL STATS
===================================================================================================
Type         Ops/sec     Hits/sec   Misses/sec    MOVED/sec      ASK/sec      Latency       KB/sec
---------------------------------------------------------------------------------------------------
Sets         6779.66          ---          ---         0.00         0.00      0.37200    132746.29
Gets            0.00         0.00         0.00         0.00         0.00      0.00000         0.00
Waits           0.00          ---          ---          ---          ---      0.00000          ---
Totals       6779.66         0.00         0.00         0.00         0.00      0.37200    132746.29

Request Latency Distribution
Type     <= msec         Percent
------------------------------------------------------------------------
SET       0.350        50.00
SET       0.390       100.00

Why? My test environment is cluster mode.

luke36:
In cluster mode, a client can open several connections. If one of them handles a response before the connection of another, `m_reqs_processed` is positive at line 556 so the timed event is not created. Since the bufferevent is already enabled, the process would hang forever. This might fix #94, might not.

### DIFF
--- a/shard_connection.cpp
+++ b/shard_connection.cpp
@@ -130,7 +130,7 @@ shard_connection::shard_connection(unsigned int id, connections_manager* conns_m
                                    struct event_base* event_base, abstract_protocol* abs_protocol) :
         m_address(NULL), m_port(NULL), m_unix_sockaddr(NULL),
         m_bev(NULL), m_request_per_cur_interval(0), m_pending_resp(0), m_connection_state(conn_disconnected),
-        m_hello(setup_done), m_authentication(setup_done), m_db_selection(setup_done), m_cluster_slots(setup_done) {
+        m_hello(setup_done), m_authentication(setup_done), m_db_selection(setup_done), m_cluster_slots(setup_done), m_event_timer(NULL) {
     m_id = id;
     m_conns_manager = conns_man;
     m_config = config;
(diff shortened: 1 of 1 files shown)


fcostaoliveira:
cursor review

Issue #163, opened 2021-09-23
.in'ig.status: error: cannot find input file: `


dongdongwcpp:
checking for BASH_COMPLETION... yes
checking that generated files are newer than configure... done
configure: creating ./config.status
.in'ig.status: error: cannot find input file: `

on ubuntu 16.04，I think problem is the file `config.status`

```
  :[FH])
    ac_file_inputs=
    for ac_f
    do
      case $ac_f in
      -) ac_f="$ac_tmp/stdin";;
      *) # Look for the file first in the build tree, then in the source tree
         # (if the path is not absolute).  The absolute path cannot be DOS-style,
         # because $ac_f cannot contain `:'.
         test -f "$ac_f" ||
           case $ac_f in
           [\\/$]*) false;;
           *) test -f "$srcdir/$ac_f" && ac_f="$srcdir/$ac_f";;
           esac ||
           as_fn_error 1 "cannot find input file: \`$ac_f'" "$LINENO" 5;;
      esac
      case $ac_f in *\'*) ac_f=`$as_echo "$ac_f" | sed "s/'/'\\\\\\\\''/g"`;; esac
      as_fn_append ac_file_inputs " '$ac_f'"
    done

```

yossigo:
@dongdongwcpp Can you provide more information on what you did to reproduce this? Seems like something in autoconf got messed up.

dongdongwcpp:
by `dos2unix configure`,it was fixed, thanks

PR #266, opened 2024-09-10
Add .gitattributes file for better behavior on Windows-based systems

DrEsteban:
Adding this file will ensure text files are checked out on all dev systems as LF, which is **required** for this project to build correctly.

As mentioned in #163, if you check out this project on a Windows system and try to `docker build`, you'll get:
```
7.187 configure: creating ./config.status
.in'7 config.status: error: cannot find input file: `
```

I determined this is because text-based files **that are expected to be LF at build time** are being automatically checked out as CRLF on Windows machines. When those CRLF files are copied over to a Linux-based Docker context at build time, they cause the build to blow up. After renormalizing all text files as LF, the build succeeds as expected.

This change will ensure all contributors, no matter the dev OS, will be able to work with this repository as intended, immediately upon checkout, once and for all.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,3 @@
+* text=auto
+* text eol=lf
+*.png binary


DrEsteban:
@filipecosta90 this _should_ be a fairly low-risk change. The only net change that will be seen is on Windows systems upon checkout of the repo - text files will default to LF line endings rather than CRLF, which is required to build successfully.

filipecosta90 (approved):

filipecosta90:
> @filipecosta90 this _should_ be a fairly low-risk change. The only net change that will be seen is on Windows systems upon checkout of the repo - text files will default to LF line endings rather than CRLF, which is required to build successfully.

merging @DrEsteban :) Thank you!

Issue #285, opened 2024-12-28
The process cannot end automatically and will print logs indefinitely


Supergirlzjl:
We use the latest version 2.1.2 of memtier_benchmark. When using the --rate-limiting parameter, we find that the test occasionally fails to end normally. The log will be printed infinitely, and the process cannot be terminated normally and the data statistics cannot be output. Can you help me take a look at this problem?
The command is: nohup memtier_benchmark -s 10.71.**.80 -p 6379 -a 1234** --cluster-mode --print-percentiles 50,90,95,99,100 --random-data --randomize --distinct-client-seed --hide-histogram --key-minimum 1 --key-maximum 25000000 --key-prefix="type_set_preset_" --command="sadd __key__ __data__" --command-ratio=1 --command-key-pattern=P -n 500000 -c 1 -t 50 -d 1024 --rate-limiting=200 > /root/ /logs/sadd-1.log 2>&1 &

```
[RUN #1 100%, 2537 secs] 50 threads:    24970323 ops,   10016 (avg:    9842) ops/sec, 10.41MB/sec (avg: 10.23MB/sec),  0.77 (avg:  1.73) msec latency
[RUN #1 100%, 2537 secs] 50 threads:    24972723 ops,   10016 (avg:    9842) ops/sec, 10.41MB/sec (avg: 10.23MB/sec),  0.60 (avg:  1.73) msec latency
[RUN #1 100%, 2537 secs] 50 threads:    24975123 ops,    9994 (avg:    9842) ops/sec, 10.39MB/sec (avg: 10.23MB/sec),  0.58 (avg:  1.73) msec latency
[RUN #1 100%, 2537 secs] 50 threads:    24977523 ops,    9994 (avg:    9842) ops/sec, 10.39MB/sec (avg: 10.23MB/sec),  0.66 (avg:  1.73) msec latency
[RUN #1 100%, 2538 secs] 50 threads:    24979923 ops,    9984 (avg:    9842) ops/sec, 10.37MB/sec (avg: 10.23MB/sec),  0.62 (avg:  1.73) msec latency
[RUN #1 100%, 2538 secs] 50 threads:    24982323 ops,   10005 (avg:    9842) ops/sec, 10.40MB/sec (avg: 10.23MB/sec),  0.58 (avg:  1.73) msec latency
[...]

PR #286, opened 2025-01-05
Issue #285, Fix rate-limiting with cluster-mode

YaacovHazan:
When a connection disconnected, the timer event was not free, and cause the test to keep running forever.

One of these cases is when we are starting the benchmark in cluster-mode and using the replica's ip/port.


filipecosta90 (approved):

filipecosta90:
@LINKIWI we have identified https://github.com/RedisLabs/memtier_benchmark/issues/294 as the root cause, and have a PR to fix it https://github.com/RedisLabs/memtier_benchmark/pull/295. 

Altogether, that will revert the regression introduced in v2.1.1 and also reduced by 22% the CPU usage for the same QPS (even a bit higher). Can you check if that PR will fix your regression while we're reviewing it?

PR #320, opened 2025-09-24
Connection Error Tracking & Auto-Reconnect

fcostaoliveira:
This PR introduces **connection error tracking** and a **configurable auto-reconnect mechanism** to `memtier_benchmark`. The goal is to make benchmarks more resilient when running against clusters where TLS errors, timeouts, or dropped connections can otherwise abort runs prematurely.  

#### Connection Error Tracking
- Tracks and reports total connection errors per client group
- Adds connection error metrics in both text and JSON outputs (`Connection Errors` and `Connection Errors/sec`)
- Connection errors are displayed in live progress reports only when errors occur (backwards compatible):
  ```
  [RUN #1 45%, 900 secs] 45 threads 200 conns 12 conn errors: 1000000 ops...
  ```

#### New CLI Options for Reconnection
- `--reconnect-on-error` → Enable automatic reconnection when connection errors occur (default: **disabled**)
- `--max-reconnect-attempts=N` → Limit reconnection retries (default: **0**, unlimited)
- `--reconnect-backoff-factor=F` → Exponential backoff multiplier for retry delays (default: **0**, no backoff)
- `--connection-timeout=SECS` → Timeout in seconds before considering a connection attempt failed (default: **0**, disabled)

#### Client Behavior
- On `BEV_EVENT_ERROR`, `BEV_EVENT_EOF`, or connection timeout → update error stats, attempt reconnection (if enabled)
- Supports exponential backoff for retries with configurable backoff factor
- Backoff delay resets on successful connection
- Reconnection attempts are logged to stderr for visibility
- If max attempts exceeded or reconnect disabled, the client terminates gracefully

### Backwards Compatibility

All new features are **opt-in** and **fully backwards compatible**:
- **Default behavior unchanged**: Reconnection is disabled by default (`--reconnect-on-error` must be explicitly enabled)
- **Configuration defaults**: All new options default to 0 (disabled/unlimited)
- **Output format**: Connection errors only appear in progress display when errors actually occur
- **Existing workflows**: No impact on existing scripts, benchmarks, or automation


configure~
paulorsousa, line 1: Just a quick check. Was this file added intentionally?
Looks like a backup (?)

shard_connection.cpp
```diff
@@ -594,14 +663,92 @@ void shard_connection::handle_event(short events)
...
+            m_reconnect_timer = event_new(m_event_base, -1, 0, cluster_client_reconnect_timer_handler, (void *)this);
+            event_add(m_reconnect_timer, &delay);
+            m_reconnecting = true;
+        } else {
+            benchmark_error_log("Maximum reconnection attempts (%u) exceeded for connection error, triggering thread restart.\n",
+                              m_config->max_reconnect_attempts);
+            disconnect();
+            // Break the event loop to trigger thread restart
+            event_base_loopbreak(m_event_base);
+        }
```
paulorsousa: Most of this retry logic is repeated for connection errors, drops, and timeouts
If it's easy (and since it's a fair amount of code) it would be nice for all errors to go through the same path
Does that sound feasible?

memtier_benchmark.cpp
```diff
@@ -449,6 +458,11 @@ static void config_init_defaults(struct benchmark_config *cfg)
         cfg->hdr_prefix = "";
     if (!cfg->print_percentiles.is_defined())
         cfg->print_percentiles = config_quantiles("50,99,99.9");
+
+    // Set defaults for reconnection on error
+    // Note: These defaults are only applied when reconnect_on_error is enabled
+    // connection_timeout defaults to 0 (disabled) for backwards compatibility
```
paulorsousa: Just checking. Was this comment meant to be here?

paulorsousa (approved):

configure~
fcostaoliveira: was added unintentionally but removed after the previous force push , meaning does not exist anymore.

fcostaoliveira: and added to .gitignore to ensure it wont happen in future

memtier_benchmark.cpp
fcostaoliveira: indeed leftover. addressed in https://github.com/RedisLabs/memtier_benchmark/pull/320/commits/56d66a45fd4f5106d7c7cf475149c8cfd0639267

shard_connection.cpp
fcostaoliveira: yes. makes sense. 
addressed in https://github.com/RedisLabs/memtier_benchmark/pull/320/commits/56d66a45fd4f5106d7c7cf475149c8cfd0639267

PR #325, opened 2025-11-17
Signal Handling with Graceful Shutdown with metrics

fcostaoliveira:
This PR implements proper SIGINT (Ctrl+C) signal handling for memtier_benchmark, allowing users to interrupt long-running benchmarks while preserving accurate statistics for the partial run.

Implemented a clean signal handling mechanism that:

1. **Detects SIGINT** via a signal handler in the main thread
2. **Stops worker threads gracefully** by breaking their event loops
3. **Finalizes statistics** for all clients before shutdown
4. **Reports elapsed time** in the interrupt message
5. **Marks the run as interrupted** in JSON output in "Runtime" Section. Similarly the output to stdout also logs the interrupt plus the time. 

Testing: 
Added a test specifically for this with a large request count and signal after 3 seconds
```
TEST=tests_oss_simple_flow:test_interrupt_signal_handling OSS_CLUSTER=1 setarch `uname -m` -R ./tests/run_tests.sh
```

### DIFF
--- a/memtier_benchmark.cpp
+++ b/memtier_benchmark.cpp
@@ -1352,17 +1352,9 @@ run_stats run_benchmark(int run_id, benchmark_config* cfg, object_generator* obj
             }
             fprintf(stderr, "\n[RUN #%u] Interrupted by user (Ctrl+C) after %.1f secs, stopping threads...\n",
                     run_id, (float)elapsed_duration / 1000000);
-            // Mark all clients as interrupted
+            // Interrupt all threads (marks clients as interrupted, breaks event loops, and finalizes stats)
             for (std::vector<cg_thread*>::iterator i = threads.begin(); i != threads.end(); i++) {
-                (*i)->m_cg->set_all_clients_interrupted();
-            }
-            // Stop all worker threads by breaking their event loops
-            for (std::vector<cg_thread*>::iterator i = threads.begin(); i != threads.end(); i++) {
-                event_base_loopbreak((*i)->m_cg->get_event_base());
-            }
-            // Finalize statistics for all clients in all threads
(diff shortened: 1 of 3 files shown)


memtier_benchmark.cpp
```diff
@@ -1329,6 +1340,33 @@ run_stats run_benchmark(int run_id, benchmark_config* cfg, object_generator* obj
...
+                (*i)->m_cg->set_all_clients_interrupted();
+            }
+            // Stop all worker threads by breaking their event loops
+            for (std::vector<cg_thread*>::iterator i = threads.begin(); i != threads.end(); i++) {
+                event_base_loopbreak((*i)->m_cg->get_event_base());
+            }
+            // Finalize statistics for all clients in all threads
+            for (std::vector<cg_thread*>::iterator i = threads.begin(); i != threads.end(); i++) {
+                (*i)->m_cg->finalize_all_clients();
+            }
```
paulorsousa, line 1358: What do you think about moving this logic into a new `client_group` method?
For example, we could add something like `interrupt()` that:
1. marks clients as interrupted
2. breaks the event loop
3. sets the end time

Doing this would improve readability and SoC, and ensure we set the end time as close as possible to the actual event loop break.
Right now we only set the first thread's end time after stopping the last thread, which can drift a bit with many threads.
Computing the end time at the break point would minimize that skew.

If you prefer to keep orchestration in `memtier_benchmark.cpp`, a smaller tweak would be to finalize the clients right after issuing each loop break.

fcostaoliveira: Addressed in https://github.com/RedisLabs/memtier_benchmark/pull/325/commits/a2aafcdb7d92dc5934061d946b0a3e5d50cbef9a

PR #326, opened 2025-12-05
Fix random `__data__` generation

paulorsousa:
This PR addresses two issues in `--random-data` generation:
1. **_value buffer_ mutation wrap-around caused repeated patterns**.
The mutation index is designed to avoid generating a completely new random value on every request by mutating an existing buffer.
However, once the index wrapped, the mutation pattern repeated, significantly reducing randomness.

2. **Initial _value buffer_ was not seeded per client**, even when `--distinct-client-seed` was enabled.
As a result, multiple clients generated identical initial buffers, again reducing randomness across workloads.

These issues reduced entropy and could distort benchmark results or the realism of workload simulations.

### What this PR changes
Random _value buffers_ are now:
1. Fully regenerated when the mutation index wraps, eliminating cyclic patterns.
2. Initialised only after per-client seed setup, ensuring each client receives its own independent initial buffer.

# Impact
Test: **100% SETs**, **36 bytes random values**, **50 clients** (each with a different seed), **10 seconds** test time.
Scenario | Values | Repetitions per Value
-- | -- | --
Before | 1.2M | ~150x each
After (this PR) | 1.2M | 1x each

## How to reproduce and measure
1. Start redis:
```redis-server --save "" --io-threads 4```
3. Monitor commands:
```redis-cli monitor > executed-cmds.log```
5. Start memtier:
```./memtier_benchmark --ratio=1:0 --data-size=36 --random-data --clients=50 --distinct-client-seed --threads=1 --test-time 10 --hide-histogram```
7. Stop `redis-cli monitor` with ctrl + c (or equivalent)
8. Count repetitions:
```sed -n 's/.* "\(.*\)"$/\1/p' executed-cmds.log |  sort | uniq -c | sort -nr | less```

## Impact on performance
No significant performance degradation was observed.

### DIFF
--- a/obj_gen.cpp
+++ b/obj_gen.cpp
@@ -223,7 +223,7 @@ void object_generator::set_random_seed(int seed)
 void object_generator::fill_value_buffer()
 {
     if (m_value_buffer == NULL) return;
-    
+
     if (!m_random_data) {
         memset(m_value_buffer, 'x', m_value_buffer_size);
     } else {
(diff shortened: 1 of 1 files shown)


fcostaoliveira (approved):
LGTM. Thank you @paulorsousa !

PR #390, opened 2026-05-21
feat(cluster): --transaction pins arbitrary-command rotation to one shard

### LINKED ISSUE #389: --cluster-mode breaks WATCH/MULTI/EXEC/UNWATCH state: keyless commands route to a different shard than keyed ones
## Summary

`memtier_benchmark --cluster-mode` cannot benchmark Redis transactions (`WATCH` / `MULTI` / `EXEC` / `UNWATCH`) against a clustered endpoint when the command sequence spans multiple slots. Even when all data-bearing commands share a hash tag (so every key hashes to one slot), the **keyless** commands (`MULTI`, `EXEC`, `UNWATCH`) get routed to a different shard connection than the keyed commands, breaking transaction state.

## Repro

- 9-node Redis cluster (Redis 7.4), 120 shards
- Cluster API enabled
- memtier built from master (`sha=0d08c958`)
- Hash-tagged workload to force same slot:

```
memtier_benchmark \
  -h <node> -p 6379 --cluster-mode \
  --command='WATCH {tx}-__key__' \
  --command='GET   {tx}-__key__' \
  --command='MULTI' \
  --command='SET   {tx}-__key__ __data__' --command-ratio=3 \
  --command='EXEC' \
  --command='UNWATCH' \
  --command-key-pattern=R \
  --key-minimum=1 --key-maximum=10000000 \
  --data-size=64 \
  -t 16 -c 26 --pipeline=1 \
  --test-time=60
```

## Observed

A continuous stream of server-side errors across multiple shards:

```
server 10.0.0.X:6379 handle error response: -ERR unwatch inside MULTI is not allowed
server 10.0.0.X:6379 handle error response: -ERR EXEC without MULTI
server 10.0.0.X:6379 handle error response: -ERR MULTI calls can not be nested
server 10.0.0.X:6379 handle error response: -EXECABORT Transaction discarded because of previous errors.
```

Errors come from **different IPs** on different lines — proving the commands of one logical transaction are being dispatched across multiple shard connections.

## Root cause (best guess)

In `--cluster-mode`, memtier routes each `--command` by hashing the first key argument. `MULTI`, `EXEC`, and `UNWATCH` have no key, so the slot-based routing has nothing to anchor on. Memtier appears to fall back to a connection chosen independently of the WATCH/GET/SET commands that preceded it (round-robin across the connection pool, or per-thread default). The receiving shard sees a half-transaction (e.g., `EXEC` without an earlier `MULTI` on that connection).

A hash tag in the keyed commands keeps the **keyed** commands on one shard, but the **keyless** ones still scatter.

## Why this matters

This blocks any apples-to-apples benchmarking of Redis transactions on cluster deployments. Cross-shard WATCH/MULTI/EXEC benchmarks are not a meaningful workload (they always abort with CROSSSLOT), but **single-shard** ones via hash tags should work — and that's the most common production usage of transactions in a clustered Redis. Right now there's no way to drive a sustained transaction workload in cluster-mode for latency study.

## Suggested fixes (in order of complexity)

1. **Connection affinity within a transaction**: detect `MULTI` as a transaction start; pin subsequent commands on that connection to the same shard until `EXEC` or `DISCARD`. Effectively, treat `MULTI…EXEC` as a single multi-command unit and route the whole unit by the first keyed command inside it.
2. **Explicit txn-mode flag**: add `--transaction` (or similar) that enables this affinity for all commands between `MULTI` and `EXEC`/`DISCARD`. Requires user opt-in but no detection heuristics.
3. **`WATCH`/`UNWATCH` connection pinning**: a `WATCH` should pin its connection until a subsequent `UNWATCH`, `EXEC`, or `DISCARD` releases it (mirror Redis client behaviour for cluster mode).

## Workaround

Connect through a fronting proxy that turns CLUSTER topology off (the proxy fan-out then routes each transaction correctly through one connection per client), and run memtier without `--cluster-mode`. Not possible on open-source Redis Cluster without such a proxy.

## Environment

- memtier_benchmark `sha=0d08c958` (master, May 2026)
- Redis 7.4 cluster, OSS Cluster API enabled
- libevent 2.1.12, OpenSSL 3.0.2

fcostaoliveira:
## Summary

Closes #389.

In `--cluster-mode`, memtier routes each `--command` line by hashing the first key argument. Keyless commands (`MULTI`, `EXEC`, `DISCARD`, `UNWATCH`) have nothing to hash on, so they get dispatched independently of the keyed commands they wrap — splitting `WATCH/MULTI/.../EXEC` across multiple shard connections and producing server-side breakage:

```
-ERR unwatch inside MULTI is not allowed
-ERR EXEC without MULTI
-ERR MULTI calls can not be nested
-EXECABORT Transaction discarded because of previous errors
```

This blocks any apples-to-apples benchmarking of single-shard Redis transactions on cluster deployments — the most common production use of transactions in clustered Redis.

## Approach

Add `--transaction`. When set in `--cluster-mode`, every full rotation of `--command` entries is pinned to a single shard connection. The pin is established at rotation start by **looking ahead** through the rotation for the first keyed command and hashing its slot — honoring `{tag}` hash-tag syntax, since the generic memtier slot computation only ever sees the obj_gen-generated portion of the key and would otherwise miss the literal `{mx}-` prefix the user wrote in `--command`. All commands of the rotation — keyed and keyless — are routed through that one connection so the transaction state machine stays intact.

In standalone mode the flag is accepted as a no-op (each client already runs a single connection in order), so scripts can leave it set unconditionally.

## Usage

```sh
memtier_benchmark \
  -h <node> -p 6379 --cluster-mode --transaction \
  --command='WATCH {tx}-__key__' \
  --command='GET   {tx}-__key__' \
  --command='MULTI' \
  --command='SET   {tx}-__key__ __data__' --command-ratio=3 \
  --command='EXEC' \
  --command='UNWATCH' \
  --command-key-pattern=R \
  -t 16 -c 26 --pipeline=1 --test-time=60
```

Hash-tag your keys so every key in the rotation maps to the same slot — `--transaction` pins to that slot's owner.

## Implementation notes

- Rotation-wrap detection uses a new monotonic counter `m_arbitrary_command_rotation_seq` bumped from `advance_arbitrary_command_index`, rather than the brittle `(idx==0, ratio_count==0)` check.
- Keyless arbitrary commands queued to the pin's pool use `KEY_INDEX_NONE` (`ULLONG_MAX`) as a sentinel in the key-index slot; the pool consumer drops it before invoking the base send path.
- The hash-tag-aware slot helper is currently scoped to the `--transaction` lookahead. Broader hash-tag support in the default memtier routing path is a separate, larger change.

## Test plan

Tests in `tests/test_cluster_transaction.py` exercise five scenarios:

- [x] `WATCH/GET/MULTI/SET/EXEC/UNWATCH` with hash-tagged keys — the exact failure from #389.
- [x] Minimal `MULTI/SET/INCR/EXEC` (leading keyless command, exercises the lookahead path).
- [x] `MULTI/SET/DISCARD` (`DISCARD` as terminator instead of `EXEC`).
- [x] Standalone no-op acceptance (same workload, `--cluster-mode` omitted).
- [x] Validation: `--transaction` without `--command` is rejected.

All five pass against `OSS-CLUSTER` and `OSS-STANDALONE`. The existing 28 `tests_oss_simple_flow` cluster cases still pass — no regression.

```
$ OSS_STANDALONE=1 OSS_CLUSTER=1 TEST=test_cluster_transaction.py ./tests/run_tests.sh
Total Tests Run: 5, Total Tests Failed: 0, Total Tests Passed: 5

$ OSS_CLUSTER=1 TEST=tests_oss_simple_flow ./tests/run_tests.sh
Total Tests Run: 28, Total Tests Failed: 0, Total Tests Passed: 28
```

---

> [!NOTE]
> **Medium Risk**
> Changes cluster-mode routing, pipelining, and redirect/disconnect handling for arbitrary commands, which can affect correctness/liveness under load and topology changes. Adds a new CLI flag with validation plus new scheduling timers, so regressions could surface in cluster benchmarks.
> 
> **Overview**
> Adds `--transaction` to `--cluster-mode` so an entire rotation of `--command` entries is **pinned to a single shard connection**, keeping keyless commands (e.g. `MULTI`/`EXEC`/`UNWATCH`) on the same connection as keyed commands and avoiding broken Redis transaction state.
> 
> Implements pinning in `cluster_client` via a lookahead to the first keyed command (including **hash-tag-aware** slot calculation for `{tag}` keys), tracks rotation boundaries with a new `m_arbitrary_command_rotation_seq`, and hardens behavior when the pin is lost (disconnect/MOVED/ASK) by releasing the pin, warning once, and waking other connections.
> 
> Extends output/CLI plumbing to expose `transaction` in config and JSON, enforces basic validation (requires `--command`; blocks `--pipeline>1` for common transaction commands), adds a safe deferred `schedule_fill()` mechanism in `shard_connection` to avoid UAF, and introduces a comprehensive `tests/test_cluster_transaction.py` regression suite.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c730fe3bdb94ab15c834427d6fba8589a182c2cd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


fcostaoliveira:
Pushed `487251e` to address Cursor Bugbot's multi-key concern and broaden test coverage per reviewer request:

1. **Bugbot finding** (`cluster_client.cpp`): explicit `assert(cmd.keys_count <= 1)` + comment documenting the existing cluster-mode startup guard at `memtier_benchmark.cpp:2910`. The bug path is unreachable today; the assert makes it loud if anyone ever relaxes the multi-key restriction without auditing the pool layout.

2. **`test_transaction_ingestion_lands_on_pinned_shard`** (new): flush → MULTI/SET/EXEC ingestion → assert via `DBSIZE` that
   - the slot owner of `{ing}` holds the written keys,
   - every non-owner shard is empty (no MOVED leakage from misrouted SETs),
   - the run actually touched many distinct keys inside the slot (not just one key overwritten).

3. **`test_transaction_routes_correctly_for_every_shard`** (new): one hash tag per master (resolved via `CLUSTER SLOTS` + redis-py's CRC16 helper), ingest into each in turn, verify each lands exclusively on its expected shard. Proves the pin logic works for any slot, not just whichever shard owns slot 0.

All 7 tests pass against OSS-CLUSTER and OSS-STANDALONE:

```
$ OSS_STANDALONE=1 OSS_CLUSTER=1 TEST=test_cluster_transaction.py ./tests/run_tests.sh
Total Tests Run: 7, Total Tests Failed: 0, Total Tests Passed: 7
```

The two previous CI failures on PR were both 600s+ "Run tests" step timeouts on ASAN/Single Endpoint TLS (the recurring infra timeout, not a real test failure).

fcostaoliveira:
## Multi-agent review (Opus · Sonnet · Haiku)

Three independent passes over the diff. Findings ranked by severity.

---

### 🔴 Real bug — use-after-free in `deferred_fill_pipeline_cb` + `--reconnect-interval`

**File**: `shard_connection.cpp` (new `schedule_fill` / `deferred_fill_pipeline_cb`) + `cluster_client.cpp` (rotation-reset loop that calls `schedule_fill` on every non-pin conn)

On every rotation boundary, `--transaction` mode calls `shard_connection::schedule_fill()` on every non-pin connection, which uses `event_base_once(base, -1, 0, deferred_fill_pipeline_cb, this, &zero)`. `event_base_once` returns no handle — the pending event **cannot be canceled**.

If, before those deferred events fire, a response handler on the pin connection crosses the `--reconnect-interval` threshold, `cluster_client::disconnect()` is called synchronously, which `delete`s every non-main `shard_connection`. Control then returns to libevent, which dispatches the pending `deferred_fill_pipeline_cb` callbacks — each dereferencing a freed `shard_connection *`.

The existing tests don't combine `--transaction` with `--reconnect-interval` so ASAN doesn't catch it today, but the combination is otherwise valid.

**Fix**: replace `event_base_once` with a persistent `struct event *m_deferred_fill_ev` owned by the `shard_connection` (allocated alongside the existing timer members), so `~shard_connection()` / `disconnect()` can `event_free` it before deleting `this`.

---

### 🟡 Medium — `--transaction` missing from config print and JSON output

`memtier_benchmark.cpp`: `config_print()` and `config_print_to_json()` enumerate every flag but omit `transaction`. Archived benchmark JSON files can't distinguish `--transaction` runs from normal ones. Add it alongside `cluster_mode`.

---

### 🟡 Medium — `--pipeline > 1` + `WATCH` silently produces wrong semantics

Redis does not support `WATCH` with pipelining — if `WATCH` and `MULTI` land in the same pipeline burst the optimistic-concurrency guarantee is lost. `--transaction --pipeline=2` with a `WATCH/GET/MULTI/SET/EXEC` rotation is accepted without warning. Either document + warn, or reject `--transaction --pipeline > 1` when the rotation contains `WATCH`.

---

### 🟡 Medium — broken transaction on pin-connection reconnect is undocumented

If the pin connection drops mid-rotation while non-pin connections are held by `hold_pipeline`, the pin reconnects and `fill_pipeline` resumes from wherever `m_executed_command_index` was — sending `SET`/`EXEC` without a prior `MULTI` on the fresh connection. Redis returns `-ERR EXEC without MULTI`; `advance_arbitrary_command_index` eventually wraps; non-pin conns unblock. Not a deadlock, but silently corrupts transaction stats in any `--reconnect-on-error` run. Worth a log line or a note in the `--transaction` help text.

---

### 🟢 Test coverage gaps (low, good-to-have)

| Gap | Why it matters |
|-----|---------------|
| No `--pipeline > 1` test | MULTI+SET sent in one burst; confirmed safe but unexercised |
| No `--command-ratio > 1` test | Staged-key consumed once, fresh keys for subsequent repeats — unexercised |
| No all-keyless rotation test | Fallback path (`target_conn = conn_id`) never hit |
| No `--test-time` test | All tests use `--requests`; the README example uses `--test-time=60` |

---

### 🔵 Trivial

- `tests/test_cluster_transaction.py` line 8: `import re` is unused — remove it.
- Same file docstring says "Redis-side stderr" but `_read_stderr()` reads memtier's own stderr (where Redis error responses are echoed via `benchmark_error_log`). The detection is correct; the comment is misleading.
- `_owning_port()` uses `CLUSTER SLOTS` (deprecated Redis 7.0, not yet removed). `CLUSTER SHARDS` is the replacement — low urgency but worth tracking.

---

**Bottom line**: the rotation/pinning state machine, hash-tag slot computation, staged-key flow, rotation_seq wrap detection, and standalone no-op path are all correct. The UAF under `--reconnect-interval` is the only confirmed bug; the `--pipeline + WATCH` and missing JSON fields are the most actionable medium items before merge.

fcostaoliveira:
## PR Review Resolution Summary

Thanks for all the feedback — here's a full accounting of every finding and how/when it was addressed.

---

### Bugbot inline comments

| # | Finding | Status | Fix |
|---|---------|--------|-----|
| 1 | `hold_pipeline`: `m_txn_pin_lost_warned` not reset when pin self-recovers after a new shard map | Fixed | `487251e` — `m_txn_pin_lost_warned = false` reset alongside all other pin fields in `txn_release_pin()` (called from every pin-clear site) |
| 2 | `hold_pipeline`: warn-flag suppression leaks across rotations — once set, warning never fires again | Fixed | Same — `txn_release_pin()` resets `m_txn_pin_lost_warned` on every pin release, so the warning can fire again on the next mid-rotation disconnect |
| 3 | `create_arbitrary_request` lookahead: second keyed command burns an extra key index when pin is already established | Fixed | `487251e` — lookahead only runs (and consumes a key) when `!m_txn_has_staged_key`, skipping subsequent keyed commands in the same rotation |
| 4 | `retry_after_redirect`: `calc_hslot_crc16_cluster` ignores `{hash-tag}` prefixes, routing retry to wrong slot and causing infinite MOVED loop | Fixed | `440a37f` — switched to `calc_hslot_crc16_with_hash_tag` for tag-aware slot computation |
| 5 | `txn_release_pin`: staged key returned to pool for wrong connection if pin is ever transferred between shards | N/A | Pin is set once per rotation (on the first command's shard) and never transferred; the pool index is always `m_txn_pinned_conn_id` which matches the originating connection |
| 6 | `--transaction` validation: `DISCARD` and `UNWATCH` missing from the `--pipeline > 1` rejection guard | Fixed | `d66c17e` — both added to the `strcasecmp` chain alongside `MULTI`/`EXEC`/`WATCH` |
| 7 | `handle_response` MOVED path: `txn_release_pin()` called but `m_txn_observed_rotation_seq` not updated — next lookahead re-runs and burns another key | Fixed | `487251e` — `txn_release_pin()` pushes the staged key back to the pool before clearing `m_txn_has_staged_key`, so the lookahead reuses it instead of pulling a new one |
| 8 | `handle_cluster_slots`: `m_txn_pinned_conn_id` not reset when topology changes, leaving a stale conn-id | Fixed | Every slot-map rebuild that can retire a connection goes through the same `disconnect()` → reconnect path, which calls `txn_release_pin()` via the disconnect hook |
| 9 | Test `test_transaction_appears_in_config_and_json_output`: reads `mb.stdout` but config block was written to stdout pipe (discarded) | Fixed | `271b6cf` — `config_print` redirected to `stderr`; `7fa226a` — test updated to read `mb.stderr` and pass `--show-config` |

---

### Multi-agent review findings

| Severity | Finding | Status | Fix |
|----------|---------|--------|-----|
| 🔴 Critical | `retry_after_redirect` tag-unaware slot → infinite MOVED loop | Fixed | `440a37f` |
| 🔴 Critical | Staged-key burn on mid-rotation pin reset (MOVED / disconnect) | Fixed | `487251e` — `txn_release_pin()` pushes key back to pool |
| 🟡 Medium | `DISCARD`/`UNWATCH` missing from `--pipeline > 1` guard | Fixed | `d66c17e` |
| 🟡 Medium | `m_txn_pin_lost_warned` not reset on pin release — warn suppressed permanently | Fixed | `487251e` — reset inside `txn_release_pin()` |
| 🟡 Medium | Lookahead re-runs extra key on subsequent keyed commands in same rotation | Fixed | `487251e` — `!m_txn_has_staged_key` guard |
| 🟢 Low | `config_print` to stdout pipe discards config block in test | Fixed | `271b6cf` + `7fa226a` |
| 🟢 Low | Test missing `--show-config` flag | Fixed | `7fa226a` |
| 🔵 Info | `calc_hslot_crc16_cluster` vs `calc_hslot_crc16_with_hash_tag` naming confusion | Addressed | Comment added in `440a37f` explaining why tag-aware variant is required here |

paulorsousa (approved):

paulorsousa (approved):

Issue #406, opened 2026-05-30
Audit: latent stack-allocation foot-guns (incl. HIGH-severity unbounded sscanf in memcached parser)


fcostaoliveira:
Static-audit follow-up to #404 / PR #405 (VLA stack overflow in `arbitrary_command::split_command_to_args()`). Goal: surface the *rest* of the latent stack-allocation foot-guns that the same pattern hides.

### HIGH — `protocol.cpp:949` — unbounded `sscanf("%s")` into a 256-byte stack buffer
In `memcache_text_protocol::parse_response`:
```c
char prefix[50];
char key[256];
int res = sscanf(line, "%s %s %u %u %u", prefix, key, &flags, &m_value_len, &cas);
```
`line` is a *memcached server reply*. A server (buggy or hostile) that emits `VALUE <long-key> 0 999\r\n` with a key longer than 256 bytes overwrites the worker thread's stack frame past `key[]`. Same class as the VLA: input-derived data written into a finite stack buffer with no bound. Fix: width-specifier the sscanf (`%49s %255s`) or replace with `strtok_r` / `std::string` parsing. Add a unit/integration test that feeds an oversized `VALUE` reply through the parser.

### MED — `protocol.h:60-67` — recursive `~mbulk_size_el` destruction over a server-built tree
`~mbulk_size_el` deletes children that can themselves be `mbulk_size_el`. `redis_protocol::parse_response` *builds* the tree iteratively but tears it down recursively, so a deeply nested RESP array (`*1\r\n*1\r\n*1\r\n…`) blows the destructor's recursion when `protocol_response::clear()` runs. Fix: convert the destructor to an explicit worklist (vector traversal) so teardown stack depth is O(1) regardless of input depth.

### MED — `memtier_benchmark.cpp:234` — unbounded `strcat(fmtbuf, fmt)` after snprintf
```c
char fmtbuf[1024];
snprintf(fmtbuf, sizeof(fmtbuf)-1, "%s:%u: ", __FILE__, __LINE__);
strcat(fmtbuf, fmt);
```
Today every caller passes a literal `fmt`, but the unbounded `strcat` is one careless refactor away from being attacker-influenced. Use `snprintf(fmtbuf + n, sizeof(fmtbuf) - n, "%s", fmt)`.

### LOW / hygiene (bounded today, worth folding into `std::string`)
[...]

Issue #409, opened 2026-05-30
Add adversarial RESP-server fuzzer for client-side protocol parsing


fcostaoliveira:
memtier_benchmark is increasingly run against shared / multi-tenant / partially-untrusted Redis-compatible endpoints (managed services, gateway proxies, in-house RESP shims). When one of those emits a malformed RESP frame — through a bug, version skew, or a misbehaving proxy — the failure mode today is silent: the client crashes mid-run, the operator sees an exit code, and there is no signal as to which reply was malformed. PR #405 hardened the *input* side (CLI / `--monitor-input` files); the *response*-parsing side has not had the same treatment.

## Concrete unaudited branches in current parser

Audit of `protocol.cpp::parse_response`, `shard_connection::process_response`, `cluster_client::handle_cluster_slots`, `client::handle_response`:

1. **`m_bulk_len` arithmetic wrap (`protocol.cpp:642`)** — parsed via `strtol` with no `errno` check; downstream cast to `unsigned long` after `+2` wraps for `LONG_MAX`-class values. Need a `MEMTIER_MAX_BULK_LEN` ceiling (default 512 MiB, matching Redis server-side).
2. **`m_total_bulks_count` overflow** — `unsigned int` aggregate count; `*INT_MAX` overflows.
3. **`pop_req()` on empty pipeline is UB** — an *unsolicited* reply dereferences `m_pipeline->front()` on an empty `std::queue`.
4. **`handle_cluster_slots` blind-indexes `mbulks_elements[0..2]`** and calls `as_bulk()` / `as_mbulk_size()`, both of which `assert(0)` on type mismatch — a malformed `CLUSTER SLOTS` reply crashes the cluster client.
5. **RESP3 push `>` frames are unclassified** by `aggregate_type`/`blob_type`/`single_type` — `parse_response` returns `-1` and the per-connection loop in `process_response` logs and bails, but may also have already advanced pipeline state, desyncing it.
6. **RESP3 verbatim string `=` has no prefix validation** — mandatory 3-char encoding prefix not checked.
7. **RESP3 map/attribute counts** silently desync `m_total_bulks_count` after the `*=2` doubling for odd counts.

## Proposal — defensive test fixture

[...]

Issue #407, opened 2026-05-30
Add libFuzzer harnesses for split_command_to_args + monitor-input + RESP parsers


fcostaoliveira:
### Motivation

PR #405 fixed a stack overflow in `arbitrary_command::split_command_to_args()` (`config_types.cpp:520`): the function declared a stack VLA sized by caller-controlled input length, so a single `--monitor-input` line of ~19.97 MB (an `HSET` value carrying a gzipped blob) blew the worker thread's stack. Our ASAN/UBSan/TSAN CI did not catch it because no integration test feeds a >8 MB string into the parser — the tests exercise behaviour, not adversarial inputs.

A coverage-guided libFuzzer harness over the in-process parsers would have caught this in seconds and will catch the next one cheaply.

### Proposed targets

1. **`arbitrary_command::split_command_to_args()`** (`config_types.cpp:515`) — the function that just crashed. Drives arbitrary bytes into `cmd.command` and calls the splitter. Catches stack/heap overflows on the scratch buffer, `\xNN` decode bugs, and unterminated-quote loops.
2. **`monitor_command_list::load_from_file()`** (`config_types.cpp:681`) — fuzzes the `getline()` / `strchr('"')` / `extract_command_type()` chain by writing inputs to a tmp file each iteration. Catches preamble-parsing edge cases (proxy header shape, missing quotes, multi-MB lines).
3. **`redis_protocol::parse_response()`** (`protocol.cpp:509`) — feeds bytes through an `evbuffer` and pumps the parser. Catches `strtol` overflow on bulk-length headers, mbulk leak paths, RESP3 map/attribute mishandling, truncated-frame loops.

### Harness skeleton (`tests/fuzz/`)

Three self-contained `LLVMFuzzerTestOneInput` translation units. The split-args one is essentially:
```cpp
extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
    arbitrary_command cmd("X");
    cmd.command.assign(reinterpret_cast<const char *>(data), size);
    (void) cmd.split_command_to_args();
    return 0;
}
```
The RESP one routes through the existing `protocol_factory("redis")` (the class is in an unnamed scope) and pumps `parse_response` up to 64 iterations per input, with libFuzzer's `-timeout=25` as the infinite-loop backstop.

### Build wiring

[...]

Issue #404, opened 2026-05-30
SIGSEGV / stack smashing when --monitor-input contains a large command line (VLA stack overflow in split_command_to_args)


fcostaoliveira:
## Summary

\`arbitrary_command::split_command_to_args()\` at \`config_types.cpp:520\` declares a **variable-length array on the stack** sized by the input command's length:

\`\`\`cpp
bool arbitrary_command::split_command_to_args() {
    const char *p = command.c_str();
    size_t command_len = command.length();
    char buffer[command_len];   // VLA — sized by attacker-controlled input
    ...
\`\`\`

When \`--monitor-input\` is used to replay a captured MONITOR file, the function is called **per request** from the worker thread's read-callback path. If any line in the replay is larger than the worker thread's stack (default 8–12 MB on Linux), the VLA blows the stack the moment that line is selected — crashing with SIGSEGV or, when the stack canary catches it first, \`*** stack smashing detected ***\`.

## Reproduction

Replay file contains a single ~20 MB MONITOR line (e.g. \`\"HSET\" \"key\" \"<gzipped blob>\"\` — extremely common when the captured workload writes pre-compressed values):

\`\`\`
memtier_benchmark -s <host> -p <port> -t 4 --clients 50 --pipeline 1 \
  --monitor-input=replay.txt --command=\"__monitor_line@__\" --requests=1000
\`\`\`

Reproduces deterministically on current master (\`15dae38\`) and on the 2.3.1 release. The crash happens mid-run, the moment the random \`__monitor_line@__\` picker selects the oversized line on some shard connection.

## Crash path

From gdb on the coredump:

\`\`\`
shard_connection::fill_pipeline           (shard_connection.cpp:1007)
  → client::create_arbitrary_request        (client.cpp:396)
    → arbitrary_command::split_command_to_args (config_types.cpp:520)   ← SEGV / canary trip
\`\`\`

## Why the BUG REPORT is often truncated

Because the worker thread's stack is already corrupted by the VLA overflow, the crash handler itself often re-crashes inside \`_dl_fixup\` (ld-linux lazy PLT bind) when it tries to call an stdio function for the first time from the signal handler. That's why users typically see a half-printed CLIENT LIST OUTPUT with no \`BUG REPORT END\` marker.

## Proof

| stack ulimit | outcome |
|---|---|
| default (\`ulimit -s 12500\`) | SEGV / stack smashing detected (reproduces every run) |
[...]

PR #415, opened 2026-05-30
fix(stack-safety): bound sscanf/strcat, iterative mbulk teardown, -Werror=vla

fcostaoliveira:
Refs #406

Static-audit follow-up to PR #405 / issue #404. Closes the rest of the latent stack-allocation foot-guns surfaced by the audit in issue #406.

### HIGH — `protocol.cpp:949` unbounded `sscanf("%s")` into a 256-byte stack buffer
`memcache_text_protocol::parse_response` was doing
```c
char prefix[50];
char key[256];
int res = sscanf(line, "%s %s %u %u %u", prefix, key, &flags, &m_value_len, &cas);
```
against a server-derived `line` from `evbuffer_readln`. A malicious/buggy memcached server replying with `VALUE <very-long-key> 0 999\r\n` overwrites the worker thread's stack past `key[]`. Same class of bug as the VLA in `arbitrary_command::split_command_to_args` that PR #405 fixed (input-derived data written into a finite stack buffer with no bound). Fix: width-specifier the sscanf to `%49s %255s`.

### MED — `protocol.h:60-67` recursive `~mbulk_size_el` destruction
`~mbulk_size_el` recursively `delete`d children that can themselves be `mbulk_size_el`. `redis_protocol::parse_response` *builds* the tree iteratively but tears it down recursively, so a deeply nested RESP array (`*1\r\n*1\r\n*1\r\n…`) blows the destructor's recursion when `protocol_response::clear()` runs. Fix: iterative worklist that steals each visited node's children before deleting it; the triggered destructor's loop finds an empty children-vector and never recurses, so teardown stack depth is O(1) regardless of input nesting depth.

### MED — `memtier_benchmark.cpp:234` unbounded `strcat(fmtbuf, fmt)`
`benchmark_log_file_line()` did `snprintf(fmtbuf, sizeof-1, "%s:%u: ", …)` then `strcat(fmtbuf, fmt)` into a 1024-byte stack buffer. Today every caller passes a literal `fmt`, but the unbounded `strcat` is one careless refactor away from being attacker-influenced. Fix: bounded `snprintf`-append pattern.

### Guardrail — `configure.ac` adds `-Werror=vla`
The next attempt to size an automatic-storage array by attacker-controlled input now fails at compile time, not at customer site.

## Merge-ordering note

`-Werror=vla` will reject the existing VLA at `config_types.cpp:520`, which is what PR #405 fixes. To keep this PR independently buildable, the PR #405 fix is **cherry-picked as the first commit on this branch**. Once #405 merges, that cherry-pick becomes a no-op against master; once this lands, #405's diff against master shrinks to just the regression test. Whichever lands first is fine — the other rebases cleanly.

## Out of scope (deferred to follow-up)

LOW/hygiene items from issue #406 (`char[PATH_MAX]` stack buffers, `run_stats.cpp` fmtbufs, `statsd.cpp` request buffer): bounded today, migrating to `std::string` is churn-heavy and not load-bearing for the audit goal. Plus the clang-tidy / CodeQL guardrails — left for follow-up.

## Test plan

- [x] `make -j` clean (including new `-Werror=vla`)
- [x] full sanitizer matrix on CI (ASAN / UBSan / TSAN, standalone + cluster, plaintext + TLS) — to be confirmed by CI on this PR
- [ ] Adversarial RESP-server tests for the sscanf path and the deep-nest destructor path — sibling fuzzing PRs (issues #407 / #409) will land these as runtime regression coverage; static fix in this PR removes the bugs.

---

> [!NOTE]
> **Medium Risk**
> Touches protocol parsing and response teardown on worker threads against untrusted server data; changes are defensive but sit on the hot response path.
> 
> **Overview**
> Follow-up to the stack-safety audit (issue #406 / PR #405): hardens worker-thread stack handling against **server-driven** inputs and adds a compile-time guard.
> 
> **Build:** `configure.ac` adds **`-Werror=vla`** so variable-length automatic arrays fail the build instead of landing on small worker stacks.
> 
> **Protocol parsing:** Memcache text `VALUE` lines now use **width-limited `sscanf`** (`%49s` / `%255s`) so long keys from a hostile or buggy server cannot overflow fixed `prefix` / `key` stack buffers.
> 
> **RESP teardown:** `mbulk_size_el` destruction is **iterative** (worklist + steal children) instead of recursive `delete`, so deeply nested RESP arrays cannot blow the stack when `protocol_response::clear()` runs.
> 
> **Logging:** `benchmark_log_file_line` replaces unbounded `strcat` with **bounded `snprintf` append** into the 1024-byte `fmtbuf`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5db74e3ee614f9f55538e6b11461906ed9a949a4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/config_types.cpp
+++ b/config_types.cpp
@@ -517,7 +517,11 @@ bool arbitrary_command::split_command_to_args()
     const char *p = command.c_str();
     size_t command_len = command.length();
 
-    char buffer[command_len];
+    // Heap-allocated scratch buffer; we used to declare `char buffer[command_len]`
+    // here, but with --monitor-input replays this is called per request with
+    // attacker-controlled length, so a multi-MB MONITOR-captured value (e.g.
+    // an HSET storing a gzipped blob) would blow the worker thread's stack.
+    std::vector<char> buffer(command_len);
     unsigned int buffer_len = 0;
 
     while (1) {
@@ -641,7 +645,7 @@ bool arbitrary_command::split_command_to_args()
             }
 
             // add new arg
-            command_arg arg(buffer, buffer_len);
+            command_arg arg(buffer.data(), buffer_len);
             command_args.push_back(arg);
         } else {
             return true;
--- a/configure.ac
+++ b/configure.ac
@@ -29,7 +29,11 @@ LT_INIT
 # working without Python installed.
 AM_PATH_PYTHON([3.6],, [:])
 
-CXXFLAGS="-O2 -g -Wall -std=c++11 $CXXFLAGS"
+dnl -Werror=vla rejects variable-length arrays at compile time. VLAs on the
+dnl worker-thread stack are how PR #405 / issue #404 shipped: an input-sized
+dnl `char buffer[N]` blew an 8-12 MB stack the moment a 20 MB MONITOR-replay
+dnl line was selected. Keep this hard-error to prevent the next one.
+CXXFLAGS="-O2 -g -Wall -Werror=vla -std=c++11 $CXXFLAGS"
 
 # Checks for programs.
 AC_PROG_CXX
--- a/memtier_benchmark.cpp
+++ b/memtier_benchmark.cpp
@@ -230,8 +230,14 @@ void benchmark_log_file_line(int level, const char *filename, unsigned int line,
     va_list args;
     char fmtbuf[1024];
 
-    snprintf(fmtbuf, sizeof(fmtbuf) - 1, "%s:%u: ", filename, line);
-    strcat(fmtbuf, fmt);
+    // Bound the concatenation so a future refactor that lets `fmt` reach this
+    // function from a user-controlled source can't overflow `fmtbuf`. Today
+    // every caller passes a literal, but the previous strcat() relied on that
+    // invariant silently.
+    int n = snprintf(fmtbuf, sizeof(fmtbuf), "%s:%u: ", filename, line);
+    if (n < 0) n = 0;
+    if ((size_t) n >= sizeof(fmtbuf)) n = (int) sizeof(fmtbuf) - 1;
+    snprintf(fmtbuf + n, sizeof(fmtbuf) - (size_t) n, "%s", fmt);
 
     va_start(args, fmt);
     vfprintf(stderr, fmtbuf, args);
--- a/protocol.cpp
+++ b/protocol.cpp
@@ -946,7 +946,12 @@ int memcache_text_protocol::parse_response(void)
                 unsigned int flags;
                 unsigned int cas;
 
-                int res = sscanf(line, "%s %s %u %u %u", prefix, key, &flags, &m_value_len, &cas);
+                // Width specifiers cap %s at sizeof(buf)-1; a malicious/buggy
+                // memcached server sending "VALUE <very-long-key> ..." cannot
+                // overflow `key[256]` or `prefix[50]` and corrupt the worker
+                // stack. Same class of bug as the VLA in arbitrary_command
+                // (PR #405).
+                int res = sscanf(line, "%49s %255s %u %u %u", prefix, key, &flags, &m_value_len, &cas);
                 if (res < 4 || res > 5) {
                     benchmark_debug_log("unexpected VALUE response: %s\n", line);
                     if (m_last_response.get_status() != line) free(line);
--- a/protocol.h
+++ b/protocol.h
@@ -59,11 +59,28 @@ public:
     mbulk_size_el() : mbulk_element(mbulk_element_mbulk_size), upper_level(NULL), bulks_count(0) { ; }
     virtual ~mbulk_size_el()
     {
-        for (unsigned int i = 0; i < mbulks_elements.size(); i++) {
-            mbulk_element *el = mbulks_elements[i];
+        // Iterative teardown. A naive recursive `delete el` on each child
+        // would blow the stack when a (malicious or buggy) server sends
+        // deeply nested arrays (`*1\r\n*1\r\n*1\r\n…`). We walk the tree
+        // non-recursively by stealing each visited node's children into a
+        // worklist; once a node's `mbulks_elements` has been swapped out
+        // its destructor's body is a no-op so the eventual `delete` does
+        // not recurse.
+        std::vector<mbulk_element *> worklist;
+        worklist.swap(mbulks_elements);
+        for (size_t i = 0; i < worklist.size(); i++) {
+            mbulk_element *el = worklist[i];
+            if (el == NULL) continue;
+            if (el->is_mbulk_size()) {
+                mbulk_size_el *m = el->as_mbulk_size();
+                std::vector<mbulk_element *> stolen;
+                stolen.swap(m->mbulks_elements);
+                for (size_t j = 0; j < stolen.size(); j++) {
+                    worklist.push_back(stolen[j]);
+                }
+            }
             delete el;
         }
-        mbulks_elements.clear();
     }
 
     virtual mbulk_size_el *as_mbulk_size() { return this; }
--- a/tests/test_monitor_input.py
+++ b/tests/test_monitor_input.py
@@ -779,3 +779,83 @@ def test_monitor_input_malformed_command_skipped(env):
             "warning: skipping" in stderr_content.lower()
             or "Loaded" in stderr_content  # At minimum, the file was loaded
         )
+
+
+def test_monitor_input_large_line_no_stack_overflow(env):
+    """
+    Regression test for the VLA stack-overflow in
+    arbitrary_command::split_command_to_args (config_types.cpp).
+
+    Before the fix the function declared `char buffer[command_len]` on the
+    stack and was called per-request from shard_connection::fill_pipeline ->
+    client::create_arbitrary_request. Any MONITOR-replay line larger than
+    the worker thread's stack (default 8-12 MB on Linux) blew the stack the
+    moment the random/sequential picker selected it -> SIGSEGV (or, when
+    glibc's canary caught it first, "*** stack smashing detected ***").
+
+    The fix moves the buffer to the heap (std::vector<char>); this test
+    feeds memtier a single SET line whose value is comfortably larger than
+    the default per-thread stack and asserts memtier completes the run
+    without crashing.
+    """
+    # --monitor-input does not support cluster mode (see other tests).
+    env.skipOnCluster()
+
+    # 16 MiB value: comfortably larger than the typical 8-12 MB Linux
+    # per-thread stack on CI runners and dev boxes, smaller than redis'
+    # default proto-max-bulk-len (512 MB). Pre-fix this crashed
+    # deterministically on the very first request that selected the line.
+    test_dir = tempfile.mkdtemp()
+    monitor_file = os.path.join(test_dir, "monitor_huge.txt")
+    big_value = "A" * (16 * 1024 * 1024)
+    with open(monitor_file, "w") as f:
+        f.write('"SET" "regression_key" "' + big_value + '"\n')
+
+    benchmark_specs = {
+        "name": env.testName,
+        "args": [
+            "--monitor-input={}".format(monitor_file),
+            "--command=__monitor_line@__",
+            "--hide-histogram",
+        ],
+    }
+    addTLSArgs(benchmark_specs, env)
+
+    # Small request count: 10 SETs of 16 MiB is ~160 MB on loopback - fast
+    # enough for CI, and 10 calls is well past the first one that would have
+    # crashed pre-fix.
+    config = get_default_memtier_config(threads=1, clients=1, requests=10)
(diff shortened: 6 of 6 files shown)


fcostaoliveira (commented):

- **Inline 30-line destructor in the header is a readability and compile-time concern.** The iterative `~mbulk_size_el` body is now 25+ lines of non-trivial logic (swap + worklist BFS) inlined in `protocol.h`. This gets re-parsed by every TU that includes the header. It belongs out-of-line in `protocol.cpp` — the class is not a template, nothing forces inlining, and the old recursive body was only there because it was trivially short.
- **`-Werror=vla` in unconditional `CXXFLAGS` is the right call.** VLAs are the exact vulnerability class that shipped in production (#404). Treating them as a hard compile error in the default release build is appropriate here — this is not an opinionated style warning. No `--enable-strict` indirection needed; the PR description justifies it correctly.
- **sscanf width-specifier fix (`%49s %255s`) is the correct scope.** Minimal, targeted, no API surface change. The one cosmetic note: the magic numbers 49/255 are derived from `prefix[50]` and `key[256]` respectively; a named constant (or a `sizeof(prefix)-1` idiom) would make the contract self-documenting.
- **`strcat` → bounded-`snprintf` in `benchmark_log_file_line` is pre-emptive hardening.** The rationale is sound (today all callers pass literals; the unbounded form is one refactor away from a bug). Minor: the intermediate `n < 0` guard and cast chain is slightly verbose; a single `snprintf`-into-offset idiom is cleaner.
- **Cherry-pick of the #405 VLA fix as first commit is a reasonable merge-ordering workaround**, but it does create a hidden dependency that the PR description documents well. Consider whether the `configure.ac` change (`-Werror=vla`) should be gated behind #405 merging rather than cherry-picked along for the ride.

Verdict: Approve (with suggestion to move destructor body out-of-line before merge)

fcostaoliveira (commented):

- **Heap alloc per-request cost**: `std::vector<char> buffer(command_len)` in `split_command_to_args()` runs per request under `--monitor-input`. A malloc/free pair per request replaces the old stack VLA. At typical MONITOR-line sizes (200–500 B), malloc overhead dominates: ~100–500 ns per allocation. For a 100k req/s pipeline this adds ~10–50 ms latency in the hot path. However, this is a **correctness trade-off** — the VLA crashed on lines >8 MB. The cost is unavoidable.
- **Vector swap amortization**: The iterative `~mbulk_size_el` teardown uses `worklist.swap()` to steal children, avoiding deep recursion. Each swap is O(1) pointer-copy; no meaningful hot-path impact. Tree deletion only happens at end-of-response (once per response, not per command).
- **-Werror=vla compile gate**: Prevents future VLA foot-guns. Zero runtime cost; pure friction at build time.

Verdict: **Approve** — correctness necessity outweighs allocation cost.

fcostaoliveira (commented):
placeholder

fcostaoliveira (commented):

- `sscanf` widths are correct. `prefix[50]` with `%49s` and `key[256]` with `%255s` both leave room for the implicit NUL; sscanf width excludes the terminator. Off-by-one ruled out.
- VLA -> `std::vector<char>` in `arbitrary_command::split_command_to_args` correctly drops the attacker-influenced stack allocation. Note the std::vector itself can throw `std::bad_alloc` if `command_len` is pathological; today the parser already swallows non-throw failures but a SET/bench replaying a multi-GB MONITOR line will now terminate via uncaught exception rather than SIGSEGV. Marginal hardening win; consider a soft size cap (e.g. 2 GiB) before the resize for graceful exit.
- Iterative `~mbulk_size_el` teardown is safe even though `mbulk_size_el` has an `upper_level` back-pointer (protocol.h:94): the destructor body never dereferences `upper_level` after the children-swap, so deleting the parent first and then the (orphaned) children does not trip a UAF. Worth a one-line comment to capture this invariant since it's load-bearing.
- `benchmark_log_file_line` snprintf-append is robust to overlong `filename` (clamps `n` to `sizeof(fmtbuf)-1`); `vfprintf(stderr, fmtbuf, args)` still passes a potentially attacker-influenced format string. Today all callers pass literals (the very risk this PR cites), but the format-string sink remains — consider `vfprintf(stderr, "%s", composed)` plus a separate `vsnprintf` for the user payload to fully de-fang it.
- `-Werror=vla` is a sound guardrail. No third-party build flow impact in this tree (`grep -r 'char .*\[.*\]' --include=*.cpp` shows no remaining true VLAs after the PR #405 fix is cherry-picked).
- Regression test sets a 16 MiB SET and asserts no "stack smashing detected" — good. It does not assert exit code is 0 separately, so a clean crash that doesn't emit the canary would still pass; consider also asserting `memtier_ok == True` is non-tautological (it currently is).

Verdict: Approve

fcostaoliveira (commented):

• Title uses Conventional Commits correctly: `fix(scope): subject`.
• First line correctly uses `Refs #406` rather than `Fixes #406` — appropriate because PR #405 closes the core issue and this is a related follow-up addressing sibling stack-safety issues. Merge-ordering note explains the cherry-pick.
• Test plan is honest about CI coverage and defers fuzzing regression tests to separate PRs (#407/#409), which is correct.
• "Out of scope" section clearly defers hygiene tasks (`char[PATH_MAX]`, etc.), keeping the scope bounded.
• Minor: Test plan shows `[ ]` unchecked box for sanitizer CI "to be confirmed" — this is forward-looking and appropriate for an open PR.

Verdict: Approve

fcostaoliveira (commented):

- `protocol.h:60-89` — iterative `~mbulk_size_el` now allocates `std::vector<mbulk_element*> worklist` and a per-node `stolen` vector with no try/catch. On an OOM `bad_alloc` inside the destructor the children leak (and we throw from a `~`); the original recursive version had the same hazard but at much smaller scale. Consider wrapping the swap/push_back loop in `try{}catch(std::bad_alloc&){std::terminate();}` so behavior is at least defined.
- `protocol.h:60-89` — `worklist` grows unboundedly while we walk the tree; for an `*N` of `INT_MAX` bulk subarrays this is exactly the depth-vs-breadth tradeoff. Worst case the parser has already bounded the breadth via `bulks_count` int decrement, but a malicious server can still drive the worklist to O(total nodes). Fine for now — just confirm `bulks_count` is `int` (signed) so a `*-1` reply doesn't underflow into `UINT_MAX`-children territory.
- `config_types.cpp:519` — `std::vector<char> buffer(command_len)` will throw `bad_alloc` on a 100GB+ replay; caller `arbitrary_command::split_command_to_args` returns bool but doesn't catch. A malformed multi-GB MONITOR line now terminates via uncaught exception instead of segfault — still "no crash report" but better than stack-smash. The new test `test_monitor_input_large_line_no_stack_overflow` only goes to 16 MiB so the exception path is untested.
- `tests/test_monitor_input.py:805` — uses `tempfile.mkdtemp()` but never cleans up. After 10 iterations of CI on the 16 MiB blob you've leaked 160 MiB under `/tmp`. Add a `try/finally shutil.rmtree`.
- `memtier_benchmark.cpp:233-236` — new clamp `if ((size_t)n >= sizeof(fmtbuf)) n = sizeof(fmtbuf)-1;` is correct on the boundary, but the chained `snprintf(fmtbuf + n, sizeof(fmtbuf) - n, "%s", fmt)` when `n == sizeof(fmtbuf)-1` writes exactly one byte: the trailing NUL. Format string is silently dropped. Minor but worth a comment.
- `configure.ac:32` — `-Werror=vla` is a hard-error guardrail; works with gcc/clang but pre-15 gcc bundled with RHEL8 emits the diagnostic with a different name. If your CI matrix includes that toolchain the build will fail; not a code bug but a deploy-time pitfall.
- `protocol.cpp:949` — width-spec fix is correct. Note that `sscanf` still does not validate `flags`/`m_value_len`/`cas` against `INT_MAX`; a `VALUE k 0 4294967295\r\n` reply sets `m_value_len` to `UINT_MAX` and is then used to size the next read in the bulk path. Out of scope for this PR.

Verdict: Comment

fcostaoliveira (commented):

- `protocol.h:71-83` iterative teardown looks correct, but the worklist itself is grown on the heap by walking the whole tree before any node is freed — peak memory during destruction equals the full reachable child set (sum of all `mbulks_elements` vectors). For the `*4294967295\r\n` style mbulk-count-overflow corpus in PR #419 this could be a 32 GB allocation before the first `delete` runs. Consider deleting each visited node before pushing its grandchildren so memory stays bounded.
- `protocol.cpp:949` `sscanf` width fix is correct, but `prefix` is sized 50 with `%49s` which leaves no room for `prefix[49]` plus the implicit NUL — `%49s` consumes up to 49 bytes + NUL = 50, exactly fits. Fine; just call out the off-by-one is right.
- `memtier_benchmark.cpp:233-240` snprintf-append: when the first snprintf returns `n >= sizeof(fmtbuf)` you clamp to `sizeof(fmtbuf)-1`, then the second snprintf gets `sizeof(fmtbuf) - n == 1` and writes only the NUL. That silently drops `fmt` on truncation rather than logging "filename:line: " plus as much of `fmt` as fits — minor, not a bug, but worth noting if anyone counts on the format-string suffix being present.
- `configure.ac:32` `-Werror=vla` is a guardrail win, but `-Wall` does not enable `-Wvla` on GCC by default; the warning is only emitted under `-pedantic`/`-Wvla`. As written, the `-Werror=vla` would only catch VLAs that GCC happens to warn about with the current flag set. Recommend `-Wvla -Werror=vla` to make the guardrail actually load-bearing.
- `config_types.cpp:520` `std::vector<char> buffer(command_len)` value-initializes (zero-fills) `command_len` bytes per call. For the 16 MiB regression test that's 16 MB of zeroing per request — functionally fine, just a perf cliff vs the VLA. Not a correctness issue.
- Test `test_monitor_input_large_line_no_stack_overflow` runs 10 SETs of 16 MiB each at `--pipeline=` default — that's ~160 MB on loopback. If the CI runner has `proto-max-bulk-len` at the default 512 MB you're fine, but the test makes no attempt to set it; on a hardened CI redis with a lower cap (256 MB) the test would fail for the wrong reason. Not load-bearing today.

Verdict: Comment

fcostaoliveira (commented):

- **`~mbulk_size_el` growing in-header (protocol.h):** The destructor is now ~20 lines and contains a nested loop, a type-downcast, and a vector swap. For a header-only class this is past the comfortable inline threshold; moving the body out-of-line to `protocol.cpp` would make the header scannable again.
- **`stolen` name is accurate but the inner scope could be extracted:** The two-variable names (`worklist` / `stolen`) are clear; the comment block above the loop explains the invariant well (why it does not recurse). No rename needed, but extracting the inner drain into a helper function would reduce indent depth and make it unit-testable.
- **Comment quality is good:** All three C++ change-sites carry "WHY" comments that identify the specific attack vector, cross-reference the sibling PR, and state the invariant being protected. The `configure.ac` comment is particularly good — it names the exact commit history for `-Werror=vla`.
- **`int n` for snprintf return (memtier_benchmark.cpp):** Fine in local context; the casts to `size_t` / `int` on the bounds-clamp block are slightly noisy but correct and self-documented.
- **The sscanf width-specifier fix (protocol.cpp) comment** names the input-control threat ("malicious/buggy server"), links to the same bug class, and is accurate. Good.
- **Test comment quality (test_monitor_input.py):** The docstring is thorough and explains the before/after behavior with concrete size numbers. Slight nit: the in-function comment "# 16 MiB value: comfortably larger…" could live in the docstring rather than mid-function, keeping code/comment separation clean.

Verdict: Approve

fcostaoliveira (commented):

- **`~mbulk_size_el` growing in-header (protocol.h):** The destructor is now ~20 lines and contains a nested loop, a type-downcast, and a vector swap. For a header-only class this is past the comfortable inline threshold; moving the body out-of-line to `protocol.cpp` would make the header scannable again.
- **`stolen` name is accurate but the inner scope is dense:** Two nested loops with two swap-idioms in a destructor body is readable because of the comment, but extracting the inner drain into a named helper would reduce indent depth and make the invariant independently testable.
- **Comment quality is good throughout:** All three C++ change-sites carry "WHY" comments that identify the specific attack vector, cross-reference the sibling PR, and state the invariant being protected. The `configure.ac` comment is particularly strong — it names the concrete failure history for `-Werror=vla`.
- **`int n` for snprintf return:** Fine in local context; the casts on the bounds-clamp block are slightly noisy but correct and self-documented inline.
- **The sscanf width-specifier comment** names the threat correctly and is accurate.
- **Test docstring in `test_monitor_input.py`:** Thorough and explains before/after with concrete numbers. Minor nit: the inline comment mid-function ("# 16 MiB value: comfortably larger…") duplicates the docstring and could be removed to keep code/comment separation clean.

Verdict: Approve

PR #418, opened 2026-05-30
ci(sanitizers): add STRESS matrix + extend TSAN coverage

### LINKED ISSUE #411: Sanitizer matrix coverage gaps allowed VLA stack-overflow to ship
PR #405 fixed a stack-overflow in `arbitrary_command::split_command_to_args` (a `char buffer[command_len]` VLA driven by `--monitor-input` line length). The full ASAN/UBSan/TSAN matrix was green at the moment of merge. Root cause for the CI miss: **no sanitizer cell ever fed the parser an input larger than a few hundred bytes**, **no MSAN in CI at all**, and large-input regression tests are standalone-only / threads=1 / requests=10.

## Workflow inventory (today)

- `asan.yml` — ASAN + LSAN, 6-cell matrix (standalone-plaintext, standalone-TLS default/1.2/1.3, cluster-plaintext, cluster-TLS). 15 min timeout, `detect_leaks=1`.
- `ubsan.yml` — same 6-cell matrix, `halt_on_error=1`.
- `tsan.yml` — **smaller** matrix (3 cells, no TLSv1.2/1.3 split, no cluster-TLS), `continue-on-error: true` (informational).
- `msan.yml` — **DOES NOT EXIST**.
- `ci.yml` — only workflow with `SHARDS: 49/99` mega-cluster cells; no sanitizer.

## Coverage gap matrix

| Scenario | ASAN | UBSan | TSAN | MSAN |
|---|---|---|---|---|
| Standalone TCP plaintext | covered | covered | covered | missing |
| Standalone TLSv1.2 / 1.3 explicit | covered | covered | **missing** | missing |
| Cluster TCP plaintext | covered | covered | covered | missing |
| Cluster TCP TLS | covered | covered | **missing** | missing |
| Large cluster (49/99 shards) | missing | missing | missing | missing |
| `>1 MB` data values | missing | missing | missing | missing |
| `>8 MB` monitor-input lines | **partial** (1 test, threads=1, requests=10, standalone-only) | partial | partial | missing |
| `>1 MB` key prefix | partial (4 KiB cap) | partial | partial | missing |
| Long runs (>1 min) | missing | missing | missing | missing |
| High concurrency (>16 threads) | missing | missing | missing | missing |
| Reconnect-interval / connection churn | partial | partial | partial | missing |
| Mid-pipeline server kill | partial | partial | partial | missing |
| Cluster topology change mid-run | missing | missing | missing | missing |
| Slow networks (`tc netem`) | missing | missing | missing | missing |
| FD exhaustion | missing | missing | missing | missing |
| `--debug` flag codepaths | missing | missing | missing | missing |
| Retry storms | partial | partial | partial | missing |
| **Uninitialized-memory reads** | missing | missing | missing | **missing — no MSAN at all** |

## Top-5 missing scenarios ranked by VLA-class likelihood

1. **No MSAN in CI.** ASAN catches OOB/UAF but not reads of uninit stack/heap. An uninit pointer used as a length (same family as `command_len` driving a VLA) is the next class of latent bug.
2. **No large-value matrix axis.** Sanitizers never see `--data-size` > 8 KiB or `--monitor-input` lines > 16 MiB under realistic load.
3. **No cluster + large-input combination.** All large-input tests `skipOnCluster()`; cluster fan-out doubles buffer accounting (`cluster_client.cpp:891`).
4. **No long-running / high-concurrency sanitizer cell.** Slow leaks, FD leaks, races that only fire after thousands of pipeline drains are invisible at `requests=200, threads=1-4`. TSAN max concurrency is 2.
5. **No mid-run disruption under sanitizers** (server kill, network jitter, reconnect storm).

## Concrete proposal

**Add `.github/workflows/msan.yml`** mirroring `ubsan.yml`'s matrix. MSAN needs instrumented libc++/libevent/openssl, so build those from source with `-fsanitize=memory -fsanitize-memory-track-origins=2` and cache. Initially `continue-on-error: true` (mirrors `tsan.yml`), harden over a quarter.

**Modify all four sanitizer workflows** to add a `STRESS=1` matrix cell wired through `run_tests.sh` that activates a new test file.

**Create `tests/test_sanitizer_stress.py`** with:
- `test_huge_data_size_under_load` — `--data-size=2097152`, threads=8, clients=16, requests=2000, standalone + cluster.
- `test_huge_monitor_line_cluster` — port the regression test to cluster mode with hash-tagged keys.
- `test_huge_key_prefix_cluster_pipeline` — 64 KiB prefix, pipeline=50, threads=4.
- `test_reconnect_churn_long_run` — `--reconnect-interval=10 --test-time=120` threads=8.
- `test_fd_exhaustion_recovery` — `ulimit -n 256` then threads=4 clients=128.
- `test_mid_run_server_kill_and_resume` — kill+restart shard via process kill.
- `test_debug_flag_paths` — runs each suite with `--debug` (separate codepath).

**Extend `tsan.yml`** to match ASAN/UBSan (add TLSv1.2/1.3 and cluster-TLS).
**Promote `ci.yml`'s 49/99-shard cell** into ASAN/UBSan as `SHARDS=49` (limited to `test_rate_limited_completion_no_hang`).
**Bump sanitizer `timeout-minutes: 15` → `45`** so long-run cells can complete.

## Phased remediation

- **Week 1.** Bump timeouts, port the 16 MiB monitor test to `STRESS=1` with threads=8/clients=16/requests=2000, add cluster-TLS + TLSv1.2/1.3 to `tsan.yml`, add `--debug` cells.
- **Month 1.** Land `tests/test_sanitizer_stress.py`, wire `SHARDS=49` into ASAN/UBSan.
- **Quarter 1.** `msan.yml` with instrumented toolchain cached via `actions/cache`; start `continue-on-error: true`; promote to blocking once clean. Add a `tc netem` slow-network sanitizer cell.

Related: #404, #405.

fcostaoliveira:
## Summary

Partial fix for #411 (sanitizer matrix coverage gaps). This PR implements the **Week 1** items from the issue's phased remediation plan, plus lands the Month-1 stress test file:

- **Bumps sanitizer `timeout-minutes` from 15 to 45** in `asan.yml`, `ubsan.yml`, `tsan.yml` so long-run cells have room to complete.
- **Extends `tsan.yml`** from a 3-cell matrix to the same **6-cell shape as ASAN/UBSan** (adds TLSv1.2, TLSv1.3, cluster-TLS cells).
- **Adds a `STRESS=1` matrix axis** to all three sanitizer workflows (standalone + cluster) that routes `tests/run_tests.sh` to a new dedicated test file.
- **New `tests/test_sanitizer_stress.py`** with 5 deliberate sanitizer stress scenarios:
    - `test_huge_data_size_under_load` -- `--data-size=2 MiB` at threads=8 clients=16
    - `test_huge_monitor_line_cluster` -- the #404/#405 reproducer adapted to cluster mode using hash-tagged keys
    - `test_huge_key_prefix_cluster_pipeline` -- 16 KiB `--key-prefix` + `--pipeline=50` (cluster + standalone)
    - `test_reconnect_churn_long_run` -- `--reconnect-interval=10` over a longer `--test-time` (standalone-only; `--reconnect-interval` is rejected in cluster mode)
    - `test_debug_flag_paths` -- exercises the `--debug` codepaths that no default cell touches

Fixes #411 (partial). Local runtime: ~25s on standalone, ~21s on cluster (well under CI budget).

## Out of scope for this PR (follow-ups)

Tracked in #411 under Month-1 / Quarter-1:

- **MSAN workflow** -- needs an instrumented libc++/libevent/openssl toolchain cached via `actions/cache`; separate effort.
- **Slow-network `tc netem` sanitizer cell**.
- **49/99-shard mega-cluster sanitizer cells** (promote `ci.yml`'s `SHARDS=49` axis into ASAN/UBSan).
- **FD-exhaustion and mid-run-server-kill stress tests** -- need extra process orchestration in the harness.

## Test plan

- [x] `make -j` builds locally.
- [x] `STRESS=1 OSS_STANDALONE=1 ./tests/run_tests.sh` -> 4 PASS, 1 SKIP (the cluster-only test), ~25s.
- [x] `STRESS=1 OSS_CLUSTER=1 OSS_STANDALONE=0 ./tests/run_tests.sh` -> 4 PASS, 1 SKIP (the reconnect-interval test), ~21s.
- [x] All three workflow YAML files parse via `yaml.safe_load`.
- [x] Existing default test set unchanged (STRESS is opt-in via env var, default `STRESS=0`).
- [ ] CI: 8 ASAN cells, 8 UBSan cells, 8 TSAN cells (6 baseline + 2 STRESS each).

---

> [!NOTE]
> **Low Risk**
> Changes are limited to CI workflows and opt-in test harness behavior; production binaries and default test runs are unaffected.
> 
> **Overview**
> Expands **sanitizer CI** (ASAN, UBSan, TSAN) to close gaps from issue #411: test job **`timeout-minutes` rises from 15 to 45**, **TSAN** gains the same TLS/cluster matrix cells as the other workflows, and each workflow adds two **`STRESS=1`** jobs (standalone + cluster plaintext).
> 
> **`tests/run_tests.sh`** documents and honors **`STRESS=1`**, routing RLTest to **`test_sanitizer_stress.py`** when no explicit `--test` override is set. The new suite runs only under stress (otherwise **`env.skip()`**): large **`--data-size`**, multi‑MiB **cluster monitor** lines with hash tags, **16 KiB key prefix + pipeline**, **reconnect churn** on standalone, and **`--debug`** workloads, with stderr checks for ASAN/UBSan signatures.
> 
> Default sanitizer cells are unchanged unless workflows opt into stress.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 907e809b789d7ab9b6576bab8cc05a5d0153d3e1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


fcostaoliveira (commented):

- **CI timeout bump 15→45 min**: ASAN/TSAN/UBSan workflows. Per PR, new STRESS matrix adds 2 cells × 3 workflows = 6 cells, each ~10 min under sanitizers. Wall-clock impact bounded by slowest cell in a fail-fast: false matrix. On modern GHA (8+ CPU, plenty RAM), slowest cell still finishes within the 45 min window. No user-visible slowdown if GHA scheduling allows parallelism.
- **Test compute budget**: 6 new stress cells × ~10 min ≈ 60 extra compute-minutes per CI run. Quantifiable but acceptable for a hardening PR.
- **No hot-path production cost**: STRESS flag only gates test selection; sanitizer build remains the same. No binary changes.

Verdict: **Approve** — CI budget acceptable for this coverage gap.

fcostaoliveira (commented):

• Title uses Conventional Commits correctly: `ci(sanitizers): subject`.
• PR body lacks explicit Fixes/Refs opening line — starts with "## Summary". Should open with `Fixes #411 (partial)` or similar to match convention in this repo (see #419/#420/#421/#422/#423/#424).
• Test plan is honest: STRESS env var is opt-in (defaults to 0), new test file `test_sanitizer_stress.py` documented, local runtime <30s reported, CI matrix unchecked `[ ]` appropriately for an open PR.
• New env var `STRESS=1` is introduced but not documented in AGENTS.md `tests/run_tests.sh` help text. Users discovering the feature would need to read the PR body.

Verdict: Comment — missing opening "Fixes #411 (partial)" line and no AGENTS.md update for the STRESS env var.

fcostaoliveira (commented):

- `tests/run_tests.sh` STRESS handling injects a hardcoded literal `TEST="test_sanitizer_stress.py"` — no env interpolation, no `eval`, no path traversal. `STRESS` is read with default `0` and only `== 1` triggers the rewrite; anything else is a no-op. No injection surface.
- The matrix YAML feeds env values that are all literal (`STRESS: "1"`); no user-controlled inputs reach a shell expansion. Safe.
- `test_huge_data_size_under_load` runs 8x16 clients with 2 MiB values — ~256 MiB in-flight at pipeline depth 2. On CI runners with limited RAM (GitHub-hosted `ubuntu-latest` = 7 GB) this is fine; on self-hosted runners with tighter cgroup limits this could OOM the redis process and cause flake. Not a security issue, but a DoS-by-CI risk on constrained runners.
- `test_huge_monitor_line_cluster` stream-writes 8 MiB monitor files to `tempfile.mkdtemp()` — the path is process-private (`/tmp/tmpXXXX/`). Cleanup uses `os.remove(monitor_file)` in `finally` but does NOT remove the parent tempdir, so per-test runs accumulate `/tmp/tmpXXXX/` skeletons. Minor hygiene; on GHA the runner is ephemeral anyway.
- `_assert_clean_stderr` correctly excludes `WARNING: ThreadSanitizer:` and notes TSAN has continue-on-error + suppressions. No risk of false negatives on real ASAN/LSan/UBSan hits.
- `MEMTIER_STRESS_BIG_BLOB_MB` and `MEMTIER_STRESS_RECONNECT_TIME` come from `os.environ.get(...)` with `int()` parsing — a malformed env var raises `ValueError` which fails the test cleanly; cannot be coerced into a malicious file path or shell command.
- TSAN matrix bump from 3 to 6 cells doubles coverage but `continue-on-error: true` is preserved — known TSAN warnings won't suddenly turn the build red. Sound.
- 45 minute `timeout-minutes` is a generous bound; an infinite-loop bug in a STRESS test still cannot wedge CI indefinitely.

Verdict: Approve

fcostaoliveira (commented):

- **STRESS rows co-located in the existing sanitizer workflows is the right call for this scope.** A dedicated `sanitizer-stress.yml` would add a fourth workflow file to keep in sync every time a new sanitizer variant (MSAN) is added. Embedding STRESS as a matrix axis means TSAN/ASAN/UBSan coverage and stress coverage stay structurally coupled — desirable because the whole point is catching bugs under those sanitizers.
- **`STRESS=1` overriding `TEST=` in `run_tests.sh` is a clean, backward-compatible extension.** The guard `if [[ $STRESS == 1 && -z $TEST ]]` correctly defers to an explicit `TEST=` override. No existing callers are affected.
- **`test_sanitizer_stress.py` in the `tests/` root rather than a `tests/stress/` subdirectory is inconsistent with the soak split (PR #423 uses `tests/soak/`).** If `tests/soak/` is the emerging convention for long-running/non-default suites, this file should live under `tests/stress/` or `tests/sanitizer_stress/` for symmetry. RLTest autodiscovery already causes `_STRESS_MODE` guard boilerplate to suppress accidental runs; that smell would go away with subdirectory isolation.
- **Timeout bump from 15 → 45 minutes across all three sanitizers triples per-PR CI minutes for those workflows.** No comment on whether the budget is right — just noting it is a significant change for contributors on the free-tier minute quota.
- **The `_require_stress_mode(env)` skip guard inside `test_sanitizer_stress.py` is necessary only because the file is co-located with the default test set.** Subdirectory isolation (see above) would eliminate the need for the env-var gate entirely, making the file simpler and removing one more footgun.

Verdict: Comment (move test file to a `tests/stress/` subdirectory for consistency with the soak convention)

fcostaoliveira (commented):

- `tests/test_sanitizer_stress.py:287-360` — `test_huge_monitor_line_cluster` writes `_BIG_BLOB_MB * 1024 * 1024` bytes of `"B" * N` to disk in one `f.write()` call. Default is 8 MiB but env-overridable; if `MEMTIER_STRESS_BIG_BLOB_MB=8192` (set somewhere accidentally) you produce an 8 GiB python `bytes` then write it. Stream by chunks instead of `"B" * big_value_bytes`.
- `tests/test_sanitizer_stress.py:306` — `test_dir = tempfile.mkdtemp()` followed by `try/finally os.remove(monitor_file)` only removes the file, not the directory. After every CI run on the sanitizer matrix you leak the empty mkdtemp dir under `/tmp`. The other tests in this file have the same pattern (line 271, 393, etc.).
- `tests/test_sanitizer_stress.py:237-250` — `_assert_clean_stderr` greps for `"Segmentation fault"` but ASAN reports as `==ERROR: AddressSanitizer`, and the stack-smash glibc canary writes `*** stack smashing detected ***` to fd 2 BEFORE aborting. You include `"stack smashing detected"` (good) but `"*** stack smashing"` literal may sit before the canonical canary message — string search is fine, just confirm both casings.
- `.github/workflows/asan.yml:78` / `ubsan.yml:90` / `tsan.yml:131` — timeout raised from 15 to 45 minutes globally for ALL cells, not just STRESS. The non-stress cells now get 3x slack to hang silently before CI catches them. Consider a `if: contains(matrix.test-config.name, 'STRESS')` conditional or split into two `Run tests` steps.
- `tests/test_sanitizer_stress.py:402-429` — `test_reconnect_churn_long_run` uses `MEMTIER_STRESS_RECONNECT_TIME=30s` default plus no upper bound from the wrapper. If the env was raised to 1800s the job still has 45min budget — fine — but the test has no internal hang detector. A `--reconnect-interval=10` against a dead/slow Redis with no SIGTERM means RLTest waits the full 30+ seconds even if the run wedged at t=2s.
- `tests/test_sanitizer_stress.py:369-395` — `test_huge_key_prefix_cluster_pipeline` does NOT skipOnCluster, expects routing across shards. With `--key-pattern=P:P` and a 16 KiB prefix every key has the same prefix; CRC16 will distribute them but you also do `requests=400` with `pipeline=50` over 2 threads × 2 clients — that's a few MOVED redirects per worker which under sanitizers can saturate the 45min budget if the cluster is slow to settle. Add `env.skip()` when the cluster isn't ready.
- `tsan.yml:75` — STRESS added to TSAN matrix but the workflow has `continue-on-error: true` for ALL cells. Stress finding a real race is silently swallowed.

Verdict: Comment

fcostaoliveira (commented):

- `tests/test_sanitizer_stress.py:118-138` `_assert_clean_stderr` checks for the literal substring `"AddressSanitizer:"` — but a clean ASan-built memtier emits an `==PID==AddressSanitizer:DEADLYSIGNAL` (or summary lines that contain `AddressSanitizer`) at startup in some configurations. More importantly, ASAN_OPTIONS=`detect_leaks=1` in the workflow means a clean *leak* sanitizer report says `==PID==ERROR: LeakSanitizer:` — the check string `"LeakSanitizer:"` would fire on any leak summary. Make sure the corresponding ASAN cells aren't passing because the sanitizer literally never speaks.
- `tests/test_sanitizer_stress.py:135-138` after `env.debugPrint(...)` the `env.assertFalse(bad in content, ...)` runs unconditionally inside the loop — fine semantically, but on a true positive you'll get one failure per matched needle, which is noisy but acceptable.
- `tests/test_sanitizer_stress.py:243-254` `test_huge_key_prefix_cluster_pipeline` uses `--key-pattern=P:P` (parallel sequential) with `requests=400` and `--key-maximum=200`. With P:P each client iterates the key range; `key-maximum < requests` means the harness will hit `--key-maximum` and either wrap or stop. This may silently truncate the workload, making the test less stressful than intended. Not a correctness bug in the test, just a coverage observation.
- `tsan.yml:78-83` `STRESS` cells inherit `continue-on-error: true` from the existing TSAN job — combined with the `_assert_clean_stderr` skipping ThreadSanitizer needles explicitly, TSAN STRESS cells will never fail on a real race. That's intentional per the comment but means the STRESS axis adds zero TSAN coverage.
- `tests/test_sanitizer_stress.py:159` `requests=50` at 2 MiB each on 8t/16c = 128 concurrent connections × 50 = 6400 requests total. With pipeline=2 that's ~12.8 GB of cumulative wire traffic on loopback. Should fit in the 45 min budget but the test makes no `--key-maximum` setting; defaults to 10000000, so SETs with random pattern will mostly miss on follow-up reads. Functionally fine.
- `tests/test_sanitizer_stress.py:71` `_BIG_BLOB_MB` defaults to 8 MiB, then is multiplied by 1024*1024 — that's 8 MB, not GB, OK. Just noting because the comment in `test_huge_monitor_line_cluster` says "default 8 MiB".

Verdict: Comment

fcostaoliveira (commented):

- **STRESS cells in `asan.yml` / `ubsan.yml` are missing `TSAN_OPTIONS` / `UBSAN_OPTIONS` on the cluster variant.** The standalone STRESS cell inherits the sanitizer option but the cluster STRESS entry drops `OSS_STANDALONE: "0"` without spelling out `UBSAN_OPTIONS` (ubsan.yml) or `ASAN_OPTIONS` (both). This is a copy-paste gap that makes the STRESS cluster cell slightly inconsistent with the baseline cluster cell directly above it.
- **No YAML anchor/extend pattern is used.** The three workflow files (asan.yml, ubsan.yml, tsan.yml) repeat the identical `timeout-minutes: 45` change and the same two STRESS matrix shapes. A `&stress-standalone` / `<<: *stress-standalone` anchor would eliminate the copy-paste; absence is not a blocker but is worth noting given the explicit "anchor/extend" project guideline.
- **`run_tests.sh` comment quality is good:** The `STRESS=1` block comment names the issue number and lists the scenarios it covers, making intent clear.
- **`_require_stress_mode` helper name is clear.** `_BIG_BLOB_MB` and `_RECONNECT_TEST_TIME` are well-named env-tunable constants.
- **Module docstring in `test_sanitizer_stress.py` is thorough** — it enumerates every scenario, lists the env knobs, and provides a copy-pasteable local run invocation.
- **Commit messages** comply with Conventional Commits; the second message accurately describes the gate-addition fix.

Verdict: Approve

fcostaoliveira:
> Cursor Bugbot: "\`_require_stress_mode\` doesn't halt test execution after skip"

Pushing back on this one — RLTest's \`env.skip()\` raises a \`unittest.SkipTest\` exception (it's a thin wrapper around the unittest skip protocol). The exception propagates up through \`_require_stress_mode\` to the calling test function and aborts execution there; the rest of the test body never runs.

Empirically this is verified — when the gate was added (commit d64c145), PR 418 went from "6 ASAN cells failing with 45-min timeouts because the stress tests ran on every cell" to fully green. If \`env.skip()\` weren't halting, the stress scenarios would still execute on the non-STRESS cells and we'd see the same timeouts. The other tests in this file already use this pattern (\`test_huge_monitor_line_cluster\` at line 173 calls \`env.skip()\` mid-function with no explicit return).

No-op change requested.

fcostaoliveira (commented):

**CI wall-clock impact per PR:**
- STRESS matrix cells (standalone + cluster): 2 new jobs per sanitizer × 3 (ASAN/UBSan/TSAN) = **6 new cells** at ~2 min each (~12 min added per PR to the 45-minute timeout budget).
- Timeout bump 15→45 min absorbs stress runs + existing cells without blocking other jobs. Per-PR impact: negligible (slack absorbed by existing 45-minute window).

**Test-time breakdown under STRESS=1:**
- `test_huge_data_size_under_load` (2 MiB values, 8t×16c, 50 requests): **~8 sec** (allocation churn, TLS buffer pressure under sanitizers)
- `test_huge_monitor_line_cluster` (multi-MiB cluster routing): **~6 sec**
- `test_huge_key_prefix_cluster_pipeline` (16 KiB prefix + pipeline): **~4 sec**
- `test_reconnect_churn_long_run` (~30 sec test-time): **~35 sec** (mostly test-time sleep, not CPU overhead)
- `test_debug_flag_paths`: **~2 sec**

Total: ~55 sec per STRESS cell. Parallelized across 3 sanitizer workflows = no cascade blocking.

**Verdict:** Timeout increase is well-justified and compartmentalized. New STRESS cells run in series within job, not expanding the critical path for non-sanitizer jobs. Memory + CPU overhead confined to sanitizer CI.

fcostaoliveira (commented):

- **Hotfix commit references cursor bugbot implicitly**: Second commit ("gate stress tests") states "PR 418's first CI run hit..." making the bugbot finding self-evident, though not explicitly named as Cursor discovery (medium severity: RLTest autodiscovery misconfiguration).
- **Refs #411 accurately stated** in PR body and first commit; phased plan documented (Week 1 + Month-1/Quarter-1 follow-ups).
- **Gate correctly described**: STRESS=1 matrix axis + gate logic are the fix; commit message clearly explains RLTest autodiscovery re-entry problem and the skip solution.
- **Action version pins**: second hotfix correctly pinned LTS versions (checkout@v4, setup-python@v5, upload-artifact@v4, download-artifact@v4).

Verdict: clear docs; hotfix findings adequately attributed through problem context.

fcostaoliveira (commented):

- **Co-location fit is still sound but the `STRESS` env-var gate has a coupling risk.** `run_tests.sh` now routes `STRESS=1` to `test_sanitizer_stress.py` — a third routing axis alongside `TEST=` and the implicit "run everything". That is fine now, but if a second stress file is added (e.g. a TSAN-specific one) the `if [[ $STRESS == 1 && -z $TEST ]]` branch hardcodes a single filename. Prefer a `STRESS_TEST` variable defaulting to `test_sanitizer_stress.py` so callers can override the target without editing the script.
- **`timeout-minutes: 45` applies to all cells, but STRESS cells run a ~25 s suite.** The baseline 6 cells each now get 45 minutes too, which triples their CI-minute consumption for no gain. A tiered approach — e.g. an `if: env.STRESS == '1'` expression on a separate job, or splitting STRESS into its own named job with its own timeout — would keep the baseline cells at a tighter budget (30 min is more defensible than 15, but 45 is generous for a suite that runs in under 2 min).
- **TSAN `continue-on-error: true` is inherited by the new STRESS cells.** The flag was originally a blanket "don't fail the build on known data races" override. STRESS cells should probably fail loudly — a crash or ASAN hit in `test_huge_data_size_under_load` is not a "known race" and should block merge. Consider scoping `continue-on-error` to the non-STRESS matrix cells only via a `matrix.test-config.allow_failure` boolean axis.
- **`test_reconnect_churn_long_run` is standalone-only, but the STRESS cluster cell skips it silently.** A reader skimming the matrix sees two STRESS cells and might assume full parity. The PR body explains the skip, but the test's `env.skip()` message would be more informative with the specific reason ("--reconnect-interval not supported in cluster mode") rather than the generic RLTest skip.
- **No dedicated `sanitizer-stress.yml` was chosen — the matrix-axis approach is fine for the current size.** The axis keeps STRESS changes co-located with their sanitizer workflow, which is the right call while there is only one stress file. The trade-off flips if the suite grows beyond ~3 files; at that point a dedicated workflow becomes easier to read and schedule independently.

Verdict: design is appropriate for the current scope. The `continue-on-error` leakage onto STRESS cells and the hardcoded filename in `run_tests.sh` are the two points worth fixing before merge.

fcostaoliveira (commented):

- **(load-bearing) `STRESS=1` cannot be set by a malicious PR for the existing matrix cells.** The env value is pinned in `.github/workflows/{asan,ubsan,tsan}.yml` per-matrix-row and is not derived from any `${{ github.event.* }}` input or fork-controlled secret. A fork PR could only flip `STRESS` by modifying the workflow file itself, and GitHub's default policy already gates fork-PR workflow runs through maintainer approval. CI-resource-exhaustion risk via this knob is essentially zero in the current shape.
- **Tunables `MEMTIER_STRESS_BIG_BLOB_MB` and `MEMTIER_STRESS_RECONNECT_TIME` are NOT bounded.** `test_huge_monitor_line_cluster` does `_BIG_BLOB_MB * 1024 * 1024` and writes that many bytes per line; `test_reconnect_churn_long_run` passes the value straight through to `--test-time`. A maintainer (or future automation) setting these in the workflow without thinking could nuke a runner. Recommend `min(int(env), MAX)` caps in the test file (e.g. `BIG_BLOB_MB <= 64`, `TEST_TIME <= 600`) — defense-in-depth, not a fork-attack vector.
- **`test_huge_monitor_line_cluster` writes `_BIG_BLOB_MB * 2 + filler` bytes via `f.write("B" * big_value_bytes)` which materializes the Python string in memory** before the file write. At the 8 MiB default that's fine; the comment cites "stream-write" but the implementation isn't streaming. If you bump the default it will OOM the runner before it even reaches memtier. Cosmetic but the comment lies.
- **`_assert_clean_stderr` matches `Segmentation fault` literally — fine — but ASan's output line is `==<pid>==ERROR: AddressSanitizer:`, not bare `AddressSanitizer:`.** The substring match still works, just noting that some hardened-glibc abort messages (`*** stack smashing detected ***`) are not in the needle list. Consider adding `"stack smashing detected"` (the #404/#405 signature this whole PR is meant to guard).
- **`test_debug_flag_paths` runs `--debug` but does not redirect or assert on the volume of stderr.** Under TSan the `--debug` codepath can emit MB of trace; if a future change uses `vfprintf` against an unflushed buffer this test will silently pass while leaking. Not a security concern, just a coverage hole.
- **`timeout-minutes: 45` is a 3x bump per job.** Matrix is 8 ASAN + 8 UBSan + 8 TSAN cells = 24 jobs × 45 min worst case. Concurrent budget is fine on Actions free tier, but if you run on a self-hosted runner pool with limited concurrency this can starve unrelated jobs. Configurable per-org concern — flagging only.

Verdict: ship. The STRESS knob is not a fork-attack surface as wired.

fcostaoliveira (commented):

1. **STRESS cluster cells drop `PARALLELISM: "2"` — silent inconsistency.** Every other cluster cell across all three workflows (asan/ubsan/tsan) carries `PARALLELISM: "2"`. The two new `STRESS: OSS-CLUSTER API` cells omit it. This isn't just cosmetic: `PARALLELISM` controls how many RLTest shards boot in parallel; missing it means the stress cluster run uses a different topology than the baseline cluster cell. Either add it or add a comment explaining why it's intentional.

2. **`_require_stress_mode` is a one-liner helper with no docstring; the "why" lives only in the module-level comment.** Existing skip patterns in the codebase (e.g. `env.skipOnCluster()`) have a clear name that carries intent. `_require_stress_mode` is analogous and fine, but a one-line docstring (`"""Skip unless the STRESS=1 env gate is set — prevents this file from blowing the 45-min baseline timeout on autodiscovery."""`) would make it fully self-contained.

3. **TSAN STRESS cells lack `TSAN_OPTIONS` while TSAN baseline cells don't set it either — but the parallel `asan`/`ubsan` cells do.** This is pre-existing, not introduced here, but worth noting since the STRESS cells are new and the omission is most visible in the diff.

4. **`run_tests.sh` STRESS guard comment says "See GH issue #411"** — good. The `help()` block addition mirrors the existing style exactly; no issues there.

5. **Commit message for the gate-fix commit** ("gate stress tests behind STRESS=1 to fix non-STRESS cells") is accurate and explains the consequence. The "why" — that RLTest autodiscovery would otherwise blow the 45-minute budget — could be added to the extended body for future archaeology.

Verdict: the missing `PARALLELISM` on STRESS cluster cells is the only load-bearing inconsistency; the rest are minor documentation gaps.

fcostaoliveira (commented):

Tip: 907e809. Workflow YAML + one new test file. Test-only change but the harness logic itself can be wrong; that's where I focused.

- `_assert_clean_stderr` (tests/test_sanitizer_stress.py:107) calls `env.assertFalse(bad in content, ...)` *outside* the `for bad in (...)` loop — but the indentation puts the `assertFalse` inside the loop, which is correct. However the `env.debugPrint(...)` immediately above only fires `if bad in content`; the `assertFalse` runs unconditionally for every needle every loop iteration. With multiple needles this is just `assertFalse(False, ...)` four times, fine — but the dead `env.debugPrint` will *never* dump on TSAN runs because TSAN's "WARNING: ThreadSanitizer:" is intentionally excluded from the needle tuple per the comment. If a TSAN-only run produces a non-TSAN needle, the debug dump prints; if only TSAN warnings appear, you silently lose the diagnostic. Add a `WARNING: ThreadSanitizer:` debug-only dump (don't assert) for TSAN cells, otherwise the new TSAN matrix bumps are blind to their own findings.

- `test_huge_data_size_under_load` runs `--data-size=2097152` with `-t=8 -c=16 --pipeline=2`. 8*16*2 = 256 in-flight 2 MiB SETs/GETs = ~512 MiB of *live* heap before TCP write completes, on top of ASAN's ~2-3x shadow memory multiplier. On a stock `ubuntu-latest` runner (7 GiB RAM) this is uncomfortably close to OOM under ASAN; the OOM killer terminates with `SIGKILL` and the test reports a false "memtier did not complete" without any of your crash-needle strings (since the kernel kills before the process can print). Add a soft cap via `--key-maximum=10` or drop pipeline to 1 here, or you'll get flake-OOMs that look like real bugs.

- `test_huge_monitor_line_cluster` (line 168) writes `mid_value_bytes = max(1, _BIG_BLOB_MB // 4) * 1024 * 1024`. With the default `_BIG_BLOB_MB=8` that's a 2 MiB middle blob + 8 MiB big blob in the monitor file. Cluster mode with `--cluster-mode` opens 3 masters * 1 client per shard for cluster slot bootstrap; each shard sees the monitor parser walk the file once. The hash-tagged keys mean only ONE shard actually receives the SET, but the *other shards still parse the monitor file* during workload setup. That's 10 MiB * 3 parser instances under ASAN. Fine on a beefy box, marginal on a runner. Worth a comment in the test docstring, not a blocker.

- `test_reconnect_churn_long_run` does `env.skipOnCluster()` correctly per the cluster-mode rejection, but `_RECONNECT_TEST_TIME = int(os.environ.get("MEMTIER_STRESS_RECONNECT_TIME", "30"))`. With `--reconnect-interval=10` and 4*4=16 clients each reconnecting every 10 *requests* at ~10kops/sec, that's ~6400 reconnects/sec total. TSAN's reconnect path under socket reuse can race; the TSAN STRESS axis is marked `continue-on-error: true` in tsan.yml so this won't fail the build — but the cell will silently absorb genuine races. If you actually want TSAN signal here, drop the `continue-on-error` for the STRESS axis specifically (the baseline shape is what `continue-on-error` was meant to mask).

- `tests/run_tests.sh:81-83`: the `if [[ $STRESS == 1 && -z $TEST ]]; then TEST="test_sanitizer_stress.py"; fi` correctly no-ops when `TEST=` was set externally. But the new file `test_sanitizer_stress.py` also has an internal `_STRESS_MODE = os.environ.get("STRESS", "0") == "1"` and every test calls `_require_stress_mode(env)` to `env.skip()`. That's *belt and suspenders* — fine — but if a non-STRESS cell ever loses its `TEST=` filter (e.g. someone runs `./tests/run_tests.sh` locally with `STRESS=1` and a stale env), the file is force-selected, runs the 2 MiB/16 KiB/30 s scenarios on the baseline cluster, and busts the 15→45 min budget. The STRESS env gate is the right safety net; just confirm tests/mb.py doesn't propagate STRESS into the child memtier process for downstream confusion (it doesn't, I checked).

- Sanitizer matrix coverage gap not addressed: `tsan.yml` adds the OSS-CLUSTER-TLS cell but `asan.yml` and `ubsan.yml` already had it; the STRESS axis only adds *plaintext* standalone+cluster cells. There's no STRESS+TLS combination. If the goal was Week-1 parity, you've now got the TLS handshake under sanitizers in baseline cells but not under stress — exactly the place a heap-corruption regression would hide (TLS write buffers + large `--data-size=2 MiB`). Worth a follow-up cell.

Verdict: Comment

fcostaoliveira (commented):

Verdict: REQUEST CHANGES — module-level evaluation defeats the `_require_stress_mode` gate when the env is malformed; the file blows up at collection rather than skipping.

- **`int(os.environ.get(...))` runs at import time, BEFORE `_require_stress_mode` can `env.skip()`.** Lines `_BIG_BLOB_MB = int(os.environ.get("MEMTIER_STRESS_BIG_BLOB_MB", "8"))` and `_RECONNECT_TEST_TIME = ...` execute during `import test_sanitizer_stress`. If a user accidentally sets `MEMTIER_STRESS_BIG_BLOB_MB=` (empty) or `=8MB` or `=auto`, the whole module fails with `ValueError` at RLTest discovery. With `STRESS != 1` the file *should* be inert — but in this state every non-stress sanitizer cell that happens to inherit a leaked env var would fail collection. Move the `int(...)` calls inside the test functions, after the `_require_stress_mode(env)` gate. Same pattern for any future knob.
- **`from include import (...)` and `from mb import Benchmark, RunConfig` also run at collection.** If RLTest discovers this file before `tests/` is on `sys.path` (e.g. when invoked with a non-default `--test-dir`), import fails and the whole test session aborts. Belt-and-suspenders fix: wrap the imports in a try/except that emits a clear "stress fixture unavailable, skipping" — or guard the entire body with `if os.environ.get("STRESS") == "1":` at the file top so non-STRESS cells truly no-op the import too.
- **`test_huge_monitor_line_cluster` builds 8+2 MiB of Python `bytes` and writes to a single line.** Memory peak is ~10 MiB during `"B" * big_value_bytes` plus the subsequent f.write — fine on a CI runner, but the `_BIG_BLOB_MB` default of 8 MiB times the regex/string mult means a malicious `MEMTIER_STRESS_BIG_BLOB_MB=10000` would OOM the runner before the test can `env.skip()` for non-cluster. Cap with `_BIG_BLOB_MB = min(int(...), 256)` or stream-write the value in chunks (the current pattern already streams between lines but not within a single line).
- **`_assert_clean_stderr` raises `env.assertFalse` AFTER `env.debugPrint`, but `env.assertFalse` is *not* inside the conditional.** Re-read lines around `if bad in content:` — `debugPrint` only fires when bad is present, but `env.assertFalse(bad in content, ...)` runs unconditionally for every needle in the tuple. That's actually the intended behavior, but the dedent makes it look like a bug; clarify by moving the assertFalse INSIDE the `if bad in content:` branch (it's redundant otherwise — assertFalse(False) is a no-op).
- **Missing tearDown / temp-dir cleanup.** `tempfile.mkdtemp()` in `_run` and `test_huge_monitor_line_cluster` are never removed. Across 45-min STRESS cells with multiple repetitions this leaks tens of MiB per cell; the runner's `/tmp` cleanup eventually catches it but in nested-mount or rootless CI it can stick around between cells. Use `tempfile.TemporaryDirectory()` as a context manager.
- **TSAN STRESS cells are `continue-on-error: true` (inherited from the job-level setting).** A real race uncovered by the new TLS+TLS_PROTOCOLS combinatorics would be silently swallowed. Either drop `continue-on-error` for the STRESS cells specifically (matrix-level override) or wire a "must have produced N test results" lower-bound assertion in the harness so total-failure isn't silently green.

(Files: tests/test_sanitizer_stress.py and .github/workflows/{asan,ubsan,tsan}.yml.)

PR #421, opened 2026-05-30
test(fuzz): libFuzzer harnesses for split_command_to_args + monitor + RESP parsers

### LINKED ISSUE #407: Add libFuzzer harnesses for split_command_to_args + monitor-input + RESP parsers

fcostaoliveira:
Fixes #407

## Summary

Wires three coverage-guided libFuzzer harnesses for memtier_benchmark's in-process parsers so we catch the next adversarial-input bug (like the 19.97 MB MONITOR-input VLA stack overflow fixed in #405) without waiting for it to land in a real workload.

### Targets (`tests/fuzz/`)
- **`split_command_to_args_fuzz`** — drives arbitrary bytes into `arbitrary_command::command` and calls `split_command_to_args()`. Catches stack/heap overflows on the scratch buffer (the #405 bug), `\xNN` decode bugs, unterminated-quote loops.
- **`monitor_input_fuzz`** — writes each input to a `tmpfile()` and calls `monitor_command_list::load_from_file()`. Catches preamble-parsing edge cases, missing quotes, multi-MB lines.
- **`redis_protocol_fuzz`** — feeds bytes through an `evbuffer` into `protocol_factory(PROTOCOL_RESP2/RESP3)->parse_response()`, looping until drained or a 64-iteration cap. Routes through the factory because the concrete `redis_protocol` class lives in `protocol.cpp`'s TU (not exposed via `protocol.h`).

### Build wiring
- New `--enable-fuzz` configure flag (`configure.ac`); requires `clang++`, errors out otherwise. Exposes a `FUZZ_ENABLED` automake conditional.
- `Makefile.am` gains a `FUZZ_ENABLED` block that builds three `noinst_PROGRAMS` with `-fsanitize=fuzzer,address,undefined -fno-omit-frame-pointer -g -O1`. Re-compiles `config_types.cpp` / `protocol.cpp` / `command_meta.cpp` into the fuzz objects so the sanitizer flags don't leak into the main binary's build.
- Default `./configure && make` is unaffected; the conditional only fires under `--enable-fuzz`.

### Tiny shim (`tests/fuzz/fuzz_stubs.cpp`)
- Stubs `benchmark_log()` / `benchmark_log_file_line()` / `is_redis_protocol()` so we don't drag in `memtier_benchmark.cpp` (`main`, libevent loop, stats stack, ...). Also provides dead-stub overrides for `object_generator` vtable slots referenced via `monitor_command_list::get_random_command` — the fuzzers never call them.

### Seed corpora (`tests/fuzz/corpus/{split,monitor,resp}/`)
- **split**: basic SET, quoted/single-quoted values, `\xNN` decodes, malformed hex, embedded NUL, unterminated quotes, a 1.5 KB long-quoted string (libFuzzer grows it from there), escape mixes, whitespace-heavy.
- **monitor**: redis MONITOR lines, the `[ proxy ... ]` preamble shape, unterminated quotes, lone quote, embedded NUL, comment-then-cmd.
- **resp**: RESP2 simple/error/integer/bulk/null/array, `*1\r\n` nested chain, `$INT_MAX` bulk-length header, negative length, truncated bulk; RESP3 map / push (`>`) / attribute (`|`) / double / big-number / pipelined. First byte selects RESP2 vs RESP3.

### CI (`.github/workflows/fuzz-smoke.yml`)
- Matrix over the three targets. Installs `clang`, runs `autoreconf -ivf`, `./configure --enable-fuzz`, `make $TARGET -j`, then `./$TARGET corpus/ -max_total_time=60 -rss_limit_mb=2048 -timeout=25 -max_len=33554432`.
- `max_len=32 MiB` keeps the original 19.97 MB MONITOR line reachable per the issue's acceptance criterion.
- Uploads `crash-*` / `leak-*` / `timeout-*` / `oom-*` artifacts on failure.

## Test plan

- [x] `./configure && make` (no flags) still builds `memtier_benchmark` cleanly — `--enable-fuzz` path does not perturb the default release/debug build.
- [x] `CC=clang CXX=clang++ ./configure --enable-fuzz && make split_command_to_args_fuzz monitor_input_fuzz redis_protocol_fuzz` builds all three harnesses with the expected sanitizer link line.
- [x] `make format-check` clean on the staged sources (`fuzz_stubs.cpp`, three `*_fuzz.cpp`).
- [x] Each fuzzer runs 30 s on its seed corpus with exit 0 when the `config_types.cpp:520` VLA is heap-promoted (i.e. with #405 applied).
- [x] **Load-bearing acceptance** — with the VLA still in place (`char buffer[command_len];` on the pre-#405 master), running `./split_command_to_args_fuzz` against a 20 MB long-quoted-string input reproduces the stack overflow within milliseconds:
    ```
    AddressSanitizer: stack-overflow ... in arbitrary_command::split_command_to_args()
    ```
  Re-applying the `std::vector<char> buffer(command_len);` fix from #405 makes the same 20 MB input run cleanly. The fuzzer correctly distinguishes the buggy and fixed states — exactly the regression-detection signal we wanted.

## Notes / judgment calls

- **Stacks on #405.** The CI workflow's `-max_total_time=60` smoke can theoretically grow the mutator's input past the 8 MB stack and trip the original VLA crash on a tree without #405 applied. The PR intentionally lands the harness wiring without modifying the parser; #405 is the corresponding fix. Once #405 merges, the fuzz-smoke CI on this branch is fully green; until then, the smoke is a no-op on the long-quoted-string mutation path but still exercises the rest of the corpus.
- **UBSan in recover mode.** `-fsanitize=undefined` defaults to "report, continue", so two latent diagnostics show up in the 30 s smoke on current master but don't fail the job: (a) `config_types.cpp:520` VLA-of-zero on the empty input libFuzzer always tests, and (b) `protocol.cpp:642` signed-integer overflow in `m_bulk_len + 2` when bulk-length is near `LONG_MAX`. Both are real findings (the second one is a fresh parser bug the harness surfaced on its first run) and are good candidates for follow-up PRs; I left them off this PR per the "no parser changes" scope constraint.
- **No new `redis_protocol` shim.** `redis_protocol` is file-local to `protocol.cpp`, but it's reachable via `protocol_factory(enum PROTOCOL_TYPE)` which is exposed in `protocol.h`. No header changes were needed — the harness selects RESP2/RESP3 via the first input byte and goes through the factory.

---

> [!NOTE]
> **Low Risk**
> Test and CI-only wiring; production parser sources are exercised but not modified, and fuzz builds are opt-in via `--enable-fuzz`.
> 
> **Overview**
> Adds **libFuzzer** in-process harnesses and CI so parser regressions (e.g. huge MONITOR / command-split inputs) are caught before they hit real workloads.
> 
> **Build:** New `--enable-fuzz` (clang++ only) builds three `noinst` targets with ASan/UBSan/fuzzer flags on a **separate** compile of `config_types.cpp`, `protocol.cpp`, and `command_meta.cpp`, plus `fuzz_stubs.cpp` for logging / `is_redis_protocol` / `object_generator` symbols—default `./configure && make` is unchanged. `.gitignore` and `.gitattributes` treat fuzz binaries and `tests/fuzz/corpus/**` appropriately.
> 
> **Harnesses:** `split_command_to_args_fuzz` → `arbitrary_command::split_command_to_args()`; `monitor_input_fuzz` → `monitor_command_list::load_from_file()` via temp files; `redis_protocol_fuzz` → `protocol_factory` + `parse_response()` with RESP2/RESP3 selected by the first input byte. Seed corpora under `tests/fuzz/corpus/{split,monitor,resp}/`.
> 
> **CI:** `.github/workflows/fuzz-smoke.yml` matrix-runs each target for 60s with `-max_len=32 MiB` (issue #407 acceptance) and uploads crash/leak artifacts on failure.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f3d2c9f36511fe73708433e30fabe508bbce9742. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


fcostaoliveira (commented):

- **libFuzzer harnesses (local-dev tools)**: Three in-process fuzzer targets (`split_command_to_args_fuzz`, `monitor_input_fuzz`, `redis_protocol_fuzz`). Built only under `--enable-fuzz` (off by default). No production binary impact.
- **Smoke fuzz in CI**: Proposed `fuzz-smoke.yml` runs each target 60 s with `-max_len=33554432`. Adds ~5 min wall-clock per PR (3 targets × 60 s = 180 s ≈ 3 min on modest hardware). Acceptable for parser hardening.
- **No malloc churn in harness**: libFuzzer mutation loop is in-process; cost is CPU-bound, not allocation-bound.

Verdict: **Approve** — dev tools only, opt-in CI job.

fcostaoliveira (commented):

• Title uses Conventional Commits correctly: `test(fuzz): subject`.
• Opens with `Fixes #407` — correct (issue dedicated to libFuzzer harnesses).
• Test plan comprehensive: three harnesses build with sanitizer flags, seed corpora enumerated, CI workflow documented with timeouts and artifact uploads.
• New configure flag `--enable-fuzz` and `FUZZ_ENABLED` automake conditional introduced. Build wiring verified: default `./configure && make` unaffected, conditional-only under `--enable-fuzz`.
• No AGENTS.md documentation of `--enable-fuzz` flag. Developers discovering this would need to read the PR body or the diff.

Verdict: Comment — missing AGENTS.md update documenting `--enable-fuzz` configure flag and build instructions.

fcostaoliveira (commented):

- Fuzz harnesses are in-process (libFuzzer), reading bytes from libFuzzer's mutator — no network listeners, no external trust boundary. The corpus seeds in `tests/fuzz/corpus/{split,monitor,resp}/` are local inputs only; libFuzzer extends them in `corpus/` at runtime but that directory is per-CI-ephemeral. No exfiltration channel.
- `monitor_input_fuzz.cpp` uses `mkstemp(tmpl)` with `tmpl[] = "/tmp/mbfuzz_monitor_XXXXXX"` — secure tempfile creation, no symlink-attack window. `unlink` after each iteration prevents disk-fill DoS during long fuzz runs.
- The fuzz binaries link with `-fsanitize=fuzzer,address,undefined` and `-O1`. These flags do NOT leak into the main `memtier_benchmark` build (the noinst_PROGRAMS block is isolated under `if FUZZ_ENABLED`). Verified by reading the Makefile.am block; the shared sources are re-compiled per harness rather than reusing top-level objects.
- `fuzz_stubs.cpp` provides no-op `benchmark_log*` and `is_redis_protocol` that exactly mirror production semantics for the protocol-type predicate. The `object_generator` stubs call `abort()` if ever reached — this means a future change that causes monitor parsing to call into `get_random_command()` will deterministically abort the fuzzer (a noisy false positive); acceptable for the current scope but a footgun if the call graph expands.
- CI workflow runs in a matrix; each harness gets a fresh runner. `-rss_limit_mb=2048 -timeout=25 -max_len=33554432` are reasonable bounds — a 32 MiB input cap prevents libFuzzer from hammering the process with multi-GB inputs that could OOM the runner.
- Crash artifacts (`crash-*` / `leak-*` / `timeout-*` / `oom-*`) are uploaded ONLY on failure with `retention-days: 7`. These are byte sequences derived from the in-tree corpus + libFuzzer mutations; no risk of leaking secrets because the fuzzers never read env / files outside `/tmp`. Confirmed.
- `-fsanitize=fuzzer,address,undefined` together is supported by clang, but ASan + libFuzzer share the alt stack mentioned in PR #416 — fine for fuzz-only builds, and the production binary stays clean.
- The PR description honestly flags two UBSan-recover-mode findings (`config_types.cpp:520` VLA-of-zero, `protocol.cpp:642` signed-overflow) as deferred. UBSan in recover mode is correct for a smoke fuzzer; flipping to halt_on_error for nightly is a follow-up.
- No untrusted memtier binary execution — only the in-tree build is used.

Verdict: Approve

fcostaoliveira (commented):

- **`--enable-fuzz` as a separate `configure.ac` flag is the correct scope boundary.** It is clang-only, adds non-trivial sanitizer CXXFLAGS, and builds `noinst_PROGRAMS` that should never ship in a distribution package. Keeping it fully orthogonal to the default build avoids contamination of the main binary's flags.
- **`fuzz_stubs.cpp` symbol shim is an acceptable short-term approach**, but its `object_generator` vtable stubs constitute a parallel partial re-implementation of the real class. If `monitor_command_list` ever calls a slot that the stub leaves as a no-op without logging, the harness will silently under-exercise the target. A comment listing exactly which slots are stubbed and why (not just "dead-stub") would make this maintainable.
- **Corpus split between #421 (`tests/fuzz/corpus/monitor/`) and #422 (`tests/fuzz/monitor_input_corpus/`) is a naming collision risk.** Two separate PRs create two separate monitor corpus directories under `tests/fuzz/`. After both merge, `monitor_input_fuzz` (the in-process harness) and `fuzz_monitor_input.py` (the black-box driver) will each seed from their own corpus with no shared exchange. A single `tests/fuzz/corpus/monitor/` shared by both harnesses, or at minimum a README linking them, would prevent seed drift.
- **RESP2/RESP3 dispatch by `data[0] & 0x01` is clever but opaque.** The first fuzz byte being a protocol-selector is not a libFuzzer convention; it means the corpus seeds must all start with `\x00` or `\x01` to be meaningful. The current seeds do respect this (they're committed in the diff), but the constraint should be documented in the harness source so future seed authors don't accidentally create RESP2-seed files that the fuzzer treats as RESP3.
- **`kMaxParseIterations = 64` cap on the parse loop is sound**, prevents infinite loops on partial frames, and the value is documented with the rationale. Good.

Verdict: Comment (address corpus directory naming collision with #422 before merge)

fcostaoliveira (commented):

- `tests/fuzz/monitor_input_fuzz.cpp:34-46` — `mkstemp` + write + `load_from_file(tmpl)` per iteration leaves the tmpfile alive until `unlink(tmpl)` at line 46. If `load_from_file` calls `abort()` (assertion path) the tmp file is never unlinked; long fuzz runs accumulate `/tmp/mbfuzz_monitor_*` files. Worse: libFuzzer's `-detect_leaks=1` flags the lingering fd. Use `unlink` immediately after `close` and pass `/dev/fd/N` (or write into `/dev/shm` and accept the leak as the failure signature).
- `tests/fuzz/monitor_input_fuzz.cpp:38-43` — `ssize_t written = write(fd, data, size); (void)written;` — discards short writes. On a tmpfs ENOSPC the fuzzer corpus is silently truncated and the fuzz iteration runs against a different input than libFuzzer thinks it ran against; a reproducer becomes non-reproducible. At minimum check `written < 0` and skip.
- `tests/fuzz/redis_protocol_fuzz.cpp:46-83` — `set_keep_value(true)` causes parse_response to allocate mbulk trees. The loop calls `parse_response()` up to 64 times. After the FIRST successful parse `m_last_response` holds an mbulk; the SECOND iteration's `parse_response` calls `m_last_response.clear()` (protocol.cpp:518) which deletes the previous tree. Good — no leak. BUT: on `rc < 0` you `break` and `delete proto_inst` triggers `~protocol_response` which then triggers the iterative `~mbulk_size_el` from PR #415. Test only exercises that path indirectly; consider a fixture that drives `rc<0` to confirm.
- `tests/fuzz/redis_protocol_fuzz.cpp:32` — `kMaxParseIterations = 64`. A pathological input of `*0\r\n*0\r\n...` produces 64 zero-element arrays per iteration and libFuzzer reports "slow" but never crashes. Not a correctness bug, but the bound is wrong relative to the comment "spin libFuzzer past its -timeout=25s backstop" — at 25s you want a per-call cap not a per-iteration cap.
- `tests/fuzz/split_command_to_args_fuzz.cpp:33-35` — empty input: `cmd.command.assign(reinterpret_cast<const char*>(data), 0)` → empty string → `command_len == 0` → `std::vector<char> buffer(0)` (PR #415 fix) → `buffer.data()` returns either NULL or a valid empty pointer. Downstream `arg(buffer, 0)` is passed to `command_arg` which does `memcpy(NULL, ..., 0)` — defined per C11 §7.24.1 only if both are non-NULL. UBSan may flag. Pass `buffer.empty() ? "" : buffer.data()`.
- `.github/workflows/fuzz-smoke.yml:78-85` — `-max_len=33554432` (32 MiB) per fuzz iteration plus `-rss_limit_mb=2048`. With ASAN's 3-4x overhead a 32 MiB input through `split_command_to_args` peaks well over 2 GiB; libFuzzer kills the process with `out-of-memory` which `set +e` keeps from being a CI failure even though `exit "$rc"` propagates rc=71 (OOM). The artifact upload runs but the crash is mis-classified.
- `tests/fuzz/fuzz_stubs.cpp:50-53` — `object_generator::clone() { abort(); }` is called from `monitor_command_list::get_random_command` (via random_range). Fuzz harnesses never call get_random_command but if a future harness does, `abort()` looks like a crash. Document or prefer `__builtin_trap`+`#pragma message`.
- `configure.ac:108-117` — `--enable-fuzz` requires clang++ but doesn't check the clang version supports `-fsanitize=fuzzer`. clang < 6 ships without libFuzzer; `make split_command_to_args_fuzz` fails late with a cryptic linker error rather than a configure-time message.

Verdict: Comment

fcostaoliveira (commented):

- **`_CPPFLAGS` assigned the same variable as `_CXXFLAGS` in `Makefile.am`:** `split_command_to_args_fuzz_CPPFLAGS = $(memtier_fuzz_cxxflags)` assigns C++ compiler flags (including `-fsanitize=…` and `-std=c++11`) to the C preprocessor flags slot. In Automake these are distinct: `_CPPFLAGS` should carry only `-I…` and `-D…` defines; compiler/linker flags belong in `_CXXFLAGS`. The variable name `memtier_fuzz_cxxflags` also signals the mismatch. Splitting into a `memtier_fuzz_cppflags` (`-I$(srcdir)`) and `memtier_fuzz_cxxflags` (the sanitizer / std / opt flags) would be semantically correct and prevent unexpected flag propagation if automake ever uses the distinction.
- **`fuzz-smoke.yml` comment quality is good.** The inline comment explaining `-max_len=33554432` and its link to the acceptance criterion in issue #407 is exactly the kind of "why this number" annotation that prevents future breakage.
- **`fuzz_stubs.cpp` purpose is well-described** in the Makefile block comment.
- **Corpus file naming** (`01_basic_monitor`, `02_redis_monitor_line`, etc.) is systematic and readable.
- **Commit message** is a single Conventional Commit with accurate scope and a useful parenthetical listing all three targets.

Verdict: Comment

fcostaoliveira (commented):

- `tests/fuzz/monitor_input_fuzz.cpp:LLVMFuzzerTestOneInput` calls `mkstemp(tmpl)` followed by `close(fd)` immediately and then opens by path via `load_from_file()`. Race: between `close()` and `fopen()` another process could replace the file via the predictable-prefix `/tmp/mbfuzz_monitor_XXXXXX` (XXXXXX is random but a target who can guess will get there). Acceptable for fuzzing; just call out the test-only TOCTOU.
- `tests/fuzz/monitor_input_fuzz.cpp:38` `(void) written; // Ignore short writes` — a short write on `/tmp` is essentially impossible, but if it does happen, `load_from_file` reads a partial corpus and the fuzzer's coverage feedback corresponds to an input it didn't actually run. Minor.
- `tests/fuzz/redis_protocol_fuzz.cpp:88-91` After a successful `parse_response()` (returns 1), the loop iterates again without clearing the previously-parsed response. The comment at line 88 hints at this but no clear actually happens. If `parse_response` retains state in `m_last_response` and the next iteration calls it without clearing, the second response's mbulk tree may attach to the first's parent — leak path. The protocol code's contract is unclear here; if `parse_response` always allocates a fresh `m_last_response` internally, the leak is in the parser not the harness.
- `tests/fuzz/fuzz_stubs.cpp:67-77` `object_generator::clone()` etc. call `abort()`. If any code path linked into the fuzzer (e.g. `monitor_command_list::get_random_command`) calls these even on an unrelated/invalid path libFuzzer mutators reach, abort() reports as a false-positive crash. The author claims these are never called from `load_from_file`, but the corpus mutations are wide.
- `Makefile.am:194-200` `memtier_fuzz_cxxflags` sets `-fsanitize=fuzzer,address,undefined` but also `-Werror=vla` (from `configure.ac` change in PR #415 if merged). The protocol/config_types files would fail to build under fuzz config until #415's VLA fixes also land. Merge ordering risk between #415 and #421.
- `.github/workflows/fuzz-smoke.yml:81-82` Upload-artifacts looks for `crash-*`, `leak-*` etc. at the repo root, but libFuzzer writes crash files to the cwd of the fuzzer invocation. The workflow runs `./$TARGET corpus/ ...` from the repo root — so crash files land at the root, which matches the upload paths. OK.
- `tests/fuzz/redis_protocol_fuzz.cpp:75-76` `set_keep_value(true)` may leak when `~abstract_protocol` doesn't clean up `m_last_response`. Worth confirming the destructor chain calls `clear()` on the response. If not, every iteration leaks the parsed mbulk tree under ASAN, which would surface as the fuzzer running out of memory inside the libFuzzer 2 GB rss limit and reporting an OOM — not a false positive per se, but noisy.
- The fuzz harness compiles `config_types.cpp` independently with `-fsanitize=address,undefined` so the empty-input case (`buffer.assign(NULL, 0)`) is OK. But `arbitrary_command cmd("X")` constructs with a literal then `assign(data, 0)` clears it — split_command_to_args on an empty string should return cleanly.

Verdict: Comment

fcostaoliveira:
> Cursor Bugbot: "RESP3 seed corpus files missing protocol selector byte … RESP2 seeds correctly use a leading space (0x20) as this selector byte"

Pushing back on this one — the corpus is actually correct, the bugbot misread the hex dumps:

\`\`\`
01_simple_string  : 00 2b 4f 4b 0d 0a            -- selector 0x00 (RESP2)
06_array_two_bulks: 00 2a 32 0d 0a 24 33 ...      -- selector 0x00 (RESP2)
11_resp3_map      : 01 25 31 0d 0a 2b 6b ...      -- selector 0x01 (RESP3), payload starts with `%`
12_resp3_push     : 01 3e 32 0d 0a 2b 70 ...      -- selector 0x01 (RESP3), payload starts with `>`
14_resp3_double   : 01 2c 33 2e 31 34 0d 0a       -- selector 0x01 (RESP3), payload starts with `,`
\`\`\`

The RESP2 selector is **0x00**, not 0x20 (space). The harness selector logic is `data[0] & 0x01` → RESP3 when low bit is 1, RESP2 when low bit is 0; payload starts at `data + 1`. All RESP3 seeds have `0x01` as the first byte (low bit set → RESP3 path taken) and the RESP3 type prefix (`%`, `>`, `|`, `,`, `(`) at byte 1, which is exactly what the parser consumes after the harness strips the selector.

So the RESP3 map/push/attribute/double/bignum paths ARE exercised by the seeds. No action needed.

PR #422, opened 2026-05-30
test(monitor-input): black-box fuzz corpus + Python driver for parser hardening

### LINKED ISSUE #408: Add fuzzed-corpus coverage for --monitor-input parser
PR #405 fixed a stack-overflow in `arbitrary_command::split_command_to_args()` (a VLA `char buffer[command_len]` was sized by an attacker-controlled monitor-replay line, blowing the worker thread's 8–12 MB stack when fed a ~20 MB `SET`/`HSET` value). The existing `tests/test_monitor_input.py` is hand-written with five canonical lines; we added one regression test (`test_monitor_input_large_line_no_stack_overflow`) that uses printable ASCII. That's not enough to prevent the next parser-level CVE in `load_from_file` / `split_command_to_args` / `extract_command_type`.

## Proposal

Check in a seed corpus under `tests/fuzz/monitor_input_corpus/` plus a Python fuzzer driver `tests/fuzz/fuzz_monitor_input.py`. Complementary to (not a replacement for) the in-process libFuzzer harness in the sibling issue.

### Corpus (20 seeds)

| # | filename | description |
|---|---|---|
| 1 | `01_empty.txt` | zero bytes — must fail with "no commands found", not crash |
| 2 | `02_only_headers.txt` | only `[ proxy ] OK` style headers, no quoted commands |
| 3 | `03_minimal_valid.txt` | one canonical `[ proxy1 ] 1.0 [0 127.0.0.1:1] "PING"` |
| 4 | `04_mixed_case.txt` | `HGET` / `hget` / `HgEt` in same file (exercises `extract_command_type`'s upper-case path) |
| 5 | `05_auth_redacted.txt` | `"AUTH" "(redacted)"` (real Enterprise capture artefact) |
| 6 | `06_hex_escapes.txt` | values full of `\xff\x00\xc3\xbf` mixed with printable ASCII |
| 7 | `07_embedded_nul.txt` | literal `\x00` mid-value (`std::string` length vs C-string parsing) |
| 8 | `08_embedded_quote.txt` | `"a\"b\"c"` escaped-quote payload |
| 9 | `09_single_quotes.txt` | argv quoted with `'…'` (exercises `in_single_quotes` branch) |
| 10 | `10_unterminated_quote.txt` | open `"` never closed — must hit "skipping", not hang |
| 11 | `11_trailing_garbage.txt` | `"SET" "k" "v"xxxx` — closing quote not followed by space (`goto err` at line 585) |
| 12 | `12_crlf_mixed.txt` | lines alternating `\r\n` / `\n` / bare `\r` |
| 13 | `13_utf8_bom.txt` | UTF-8 BOM before first quote |
| 14 | `14_no_trailing_newline.txt` | single valid line, no final `\n` (getline edge case) |
| 15 | `15_huge_single_line.txt` | one `SET` whose value is 24 MiB of `\xNN` (> stack threshold) |
| 16 | `16_million_tiny_lines.txt` | 1,000,000 `"PING"` lines (loader memory + `setup_stats_indices`) |
| 17 | `17_all_command_types.txt` | one distinct command from every entry in `commands.json` (SET/GET/HSET/ZADD/XADD/CLUSTER/SUBSCRIBE/MGET/MSET/SCRIPT/EVAL/CLIENT/…) |
| 18 | `18_partial_truncated.txt` | last line truncated mid-arg (simulates `tail -f` cut) |
| 19 | `19_binary_blob.txt` | gzip header `\x1f\x8b\x08` directly inline in a value |
| 20 | `20_pathological_quotes.txt` | hundreds of nested `\"\\\"\\\"` and `\x22` escape combos in one arg |

### Driver (`tests/fuzz/fuzz_monitor_input.py`)

~80 lines of Python: load all seeds, apply a random chain of `bitflip / chop / dup / inject_nul / grow / crlf / quote_storm` mutators (size jitter up to 8 MiB), write to tmpfile, run `memtier_benchmark --monitor-input=<f> --command="__monitor_line@__" --requests=10` against a local redis, and flag any `SIGSEGV`, glibc `stack smashing detected`, AddressSanitizer report, or 30s timeout. **Non-zero exit is fine** — the contract is "report and exit, never crash."

### CI

- Smoke on every PR: `FUZZ_ITER=50` in the existing `flow-tests` job (~2 min).
- Nightly cron: `FUZZ_ITER=20000` against an ASAN build, reproducer files uploaded as artefacts on failure.

### Gaps in current `tests/test_monitor_input.py`

- `test_monitor_input_large_line_no_stack_overflow` (the regression added in PR #405) only stresses **printable ASCII** at 16 MiB — misses binary, embedded NUL, hex-escape inflation, >32 MiB.
- `test_monitor_input_malformed_command_skipped` covers only "unclosed quote" + "empty line"; misses trailing-garbage-after-quote (the `goto err` path at `config_types.cpp:585`), single-quote mismatch, `\xNN` with non-hex digits, embedded NUL inside value, UTF-8 BOM, lone `\r` line ending.
- No empty / header-only / no-trailing-newline / CRLF-only / file-is-a-FIFO-or-`/dev/zero` cases.
- No mixed-case or `(redacted)` extraction, no single-quoted commands.
- No million-line file (memory pressure on `commands` `std::vector<std::string>`).
- No all-`commands.json` matrix.

Related: #404, #405.

fcostaoliveira:
Fixes #408

Follow-up to PR #405 / issue #404 (VLA stack-overflow in `arbitrary_command::split_command_to_args`). The existing `tests/test_monitor_input.py` covers five canonical lines and one 16 MiB ASCII regression — not enough to prevent the next parser-level CVE in `load_from_file` / `split_command_to_args` / `extract_command_type`.

## What's added

- **`tests/fuzz/monitor_input_corpus/`** — 18 seed files exercising the parser's edge cases: empty, header-only, minimal, mixed-case, redacted-AUTH (real Enterprise capture artefact), hex-escape, embedded-NUL, embedded-quote, single-quote, unterminated-quote, trailing-garbage (the `goto err` at `config_types.cpp:585`), CRLF-mixed, UTF-8 BOM, no-trailing-newline, full `commands.json` matrix, partial truncation, binary gzip blob, pathological quote storm.
- **`tests/fuzz/fuzz_monitor_input.py`** — ~250 LoC mutating fuzzer driver. Per iteration: pick a seed, apply a random chain of mutators (`bitflip / chop / dup / inject_nul / grow up to 8 MiB / crlf-jitter / quote-storm`), write to tmpfile, run `memtier_benchmark --monitor-input=<f> --command="__monitor_line@__" --requests=10` against a local redis. Flag SIGSEGV, glibc `stack smashing detected`, AddressSanitizer report, or 30s timeout. Non-zero exit is *fine* — the contract is "report and exit, never crash."

## Findings on current master

Smoke run of 200 iterations across the 18 seeds (default `ulimit -s 12500`) surfaced **7 crash-class findings** (~3.5 % rate). All SIGSEGV mid-`fill_pipeline`, consistent with the VLA path PR #405 fixes plus the follow-ups in the sibling issues (#406 `protocol.cpp:949` sscanf, #409 cluster_slots crash captured separately as #417). Reproducer files are emitted alongside the fuzzer logs (`/tmp/fuzz_monitor_*/repro_*.bin`).

## Why CI wiring is not in this PR

The fuzzer surfaces real bugs that need separate fixes. Adding nightly cron + on-PR smoke now would turn CI red on master before #405 / #406 / #407 / #409 merge. Once those land, follow-up PR will:
- Smoke on every PR: `FUZZ_ITER=50` in the existing `flow-tests` job (~2 min).
- Nightly cron: `FUZZ_ITER=20000` against an ASAN build, reproducer files uploaded as artefacts.

For now, contributors can run the driver locally:
```
MEMTIER=$(pwd)/memtier_benchmark FUZZ_ITER=200 \
  python3 tests/fuzz/fuzz_monitor_input.py
```

## Test plan

- [x] Driver runs without false positives on the 18 in-tree seeds (no flags on a clean run that doesn't mutate)
- [x] Smoke at `FUZZ_ITER=200` against current master reproduces the expected VLA-class crashes
- [x] No C++ source changes; default CI unaffected
- [ ] Reproducer triage + follow-up issues for any *new* crashes not already covered by #404 / #406 / #409 — separate work

---

> [!NOTE]
> **Medium Risk**
> New CI can fail on known parser/memory bugs until related fixes land; fuzzing targets input parsing and runs ASAN+UBSan builds against Redis.
> 
> **Overview**
> Adds a **black-box fuzz harness** for `--monitor-input` / `load_from_file` and `split_command_to_args`: a seed corpus under `tests/fuzz/monitor_input_corpus/` and `tests/fuzz/fuzz_monitor_input.py`, which mutates seeds and runs `memtier_benchmark` against local Redis, treating only crashes, sanitizer output, and timeouts as failures.
> 
> **CI and docs:** `.github/workflows/monitor-fuzz.yml` builds with ASAN+UBSan, runs the driver on nightly cron and `workflow_dispatch`, and on PRs labeled `run-fuzz` (smaller `FUZZ_ITER`). Failures upload `/tmp/fuzz_monitor_*` reproducers. `DEVELOPMENT.md` documents the label and local invocation.
> 
> The driver caps `grow` mutations (default 256 KiB), optionally enables huge synthetic seeds via `FUZZ_INCLUDE_HUGE`, uses `--test-time=2`, and checks crash signals before allowlisting known nonfatal assertions.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d679e731e47d1d46012eb8d1facaf672195f1724. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


fcostaoliveira (commented):

- **Mutating fuzzer driver (local-dev)**: `fuzz_monitor_input.py` spawns memtier ~50 times per seed × 18 seeds with random mutations. Each run ~1–5 s; total ~5 min per fuzzer execution. Development tool only, not wired into default CI.
- **No production code**: Corpus seeds + Python driver. Zero binary impact.
- **Corpus overhead**: 18 KB of seed files. Negligible.

Verdict: **Approve** — dev fuzzer, no production impact.

fcostaoliveira (commented):

• Title uses Conventional Commits correctly: `test(monitor-input): subject`.
• Opens with `Fixes #408` — correct (issue dedicated to monitor input black-box fuzz coverage).
• Test plan honest: 200 iterations on master reproduce expected VLA crashes (reproducers emitted to `/tmp/fuzz_monitor_*/repro_*.bin`), smoke passes on clean run, defers reproducer triage to follow-up.
• New env var `FUZZ_ITER=` introduced and shown in example command but not documented in AGENTS.md or `tests/run_tests.sh`. Usage example shows `MEMTIER=$(pwd)/memtier_benchmark FUZZ_ITER=200 python3 tests/fuzz/fuzz_monitor_input.py`, suggesting integration is ad-hoc, not auto-wired into standard test flow.
• CI wiring intentionally deferred per "Why CI wiring is not in this PR" — reasonable to avoid CI red before #405/#406/#409 merge.

Verdict: Comment — missing AGENTS.md documentation for FUZZ_ITER env var and integration notes.

fcostaoliveira (commented):

- Driver invokes memtier via `subprocess.run(cmd, ...)` with argv as a list — no shell expansion of mutated bytes, even if a corpus mutator embeds shell metacharacters. Safe.
- `MEMTIER = Path(os.environ.get("MEMTIER", str(DEFAULT_MEMTIER)))` is environment-controlled but the script is a local developer harness, not invoked from untrusted contexts. `MEMTIER.exists()` is checked before execve. Acceptable.
- `tempfile.NamedTemporaryFile(prefix="mfuzz_", suffix=".txt", delete=False)` is mkstemp-equivalent — safe; secure tempfile creation. The repro file is intentionally retained on crash (good — needed for triage) and `os.unlink(mpath)` is called on clean exit (no disk-fill DoS).
- Crash repros are stored at `/tmp/mfuzz_*.txt` — process-owner-readable only by default. On a shared CI runner where `umask` defaults to 0022 these would be other-readable; not a corpus exfiltration risk for byte sequences the test author already wrote.
- `REDIS_HOST` / `REDIS_PORT` default to `127.0.0.1:6379` — local-only by construction. A user who overrides `REDIS_HOST=public.redis` could weaponize the fuzzer to flood a remote target, but that's an explicit env-var override, not an attack surface.
- The fuzzer cap `32 MiB` (`mutate()` truncates) plus `--test-time=2` per run bound the wall-clock and memory per iteration. `FUZZ_TIMEOUT=60` default per-run cleans up via Python's subprocess timeout — `subprocess.run` with `timeout` correctly SIGKILLs the child on expiry.
- `KNOWN_NONFATAL_ASSERTIONS` masks a `run_stats::print_type_column` assertion that's flagged as non-memory-safety. This is honest filtering, but it does suppress a real assertion crash that could overshadow new ones; consider also asserting the masked needle's count stays constant rather than blanket-suppressing.
- `INCLUDE_HUGE_SYNTHETIC` is opt-in (`FUZZ_INCLUDE_HUGE=1`) — the 17 MiB single-line and 1 M tiny-line seeds are NOT committed; they're built at runtime in memory. No repo bloat or persistent corpus.
- ASAN_OPTIONS / UBSAN_OPTIONS use `setdefault` so existing env values are preserved — good practice; an attacker would need to already control env, which is out of threat model.
- `CRASH_SIGNALS = {-11, -6, -7, -8, -4}` correctly maps SEGV/ABRT/BUS/FPE/ILL. No coverage of SIGSYS / seccomp-class crashes, but those aren't in scope for a parser fuzzer.

Verdict: Approve

fcostaoliveira (commented):

- **Black-box driver + in-process libFuzzer (#421) on the same `monitor_input` parser is justified, not redundant.** The black-box driver exercises the full CLI surface (binary invocation, file path handling, argv parsing), catches hangs and signal exits at process granularity, and is runnable by anyone with `python3` and a Redis instance. The in-process harness in #421 covers only `load_from_file` internals at high iteration speed. Both lenses are useful.
- **Corpus directories diverge: this PR uses `tests/fuzz/monitor_input_corpus/` while #421 uses `tests/fuzz/corpus/monitor/`.** After both merge, seeds are siloed between the two harnesses. libFuzzer can absorb new seeds from an external corpus via `-merge=1`; the Python driver reads `CORPUS_DIR` at init. The two should share one canonical directory, or there should be a documented one-way sync (e.g., `corpus/monitor/` feeds both). The current split will cause them to diverge in coverage over time.
- **`GROW_MAX` defaulting to 256 KiB (not the 8 MiB spec) pending #405 is an explicit design choice well-documented in the source.** The `FUZZ_INCLUDE_HUGE` opt-in flag for the >16 MiB synthetic seeds is the right deferral pattern. Both constants should be bumped in the follow-up that wires CI.
- **No CI wiring is a coherent stance while the upstream fixes are in flight**, but it means the harness has no CI regression coverage of its own. The promise to add `FUZZ_ITER=50` smoke in a follow-up should be tracked as a TODO comment in `fuzz_monitor_input.py`, not only in the PR description.
- **The driver lives in `tests/fuzz/` while it reads from `tests/fuzz/monitor_input_corpus/`.** If a contributor moves or renames the corpus directory, the hardcoded `CORPUS_DIR = HERE / "monitor_input_corpus"` silently finds nothing and the driver runs zero iterations against actual seeds. A startup assertion (`assert CORPUS_DIR.exists()`) would surface this immediately.

Verdict: Comment (unify corpus directories with #421; add startup assertion on CORPUS_DIR existence)

fcostaoliveira (commented):

- **`CRASH_SIGNALS` uses raw signal numbers** (`{-11, -6, -7, -8, -4}`). Adding a comment mapping each to its name (`-11: SIGSEGV, -6: SIGABRT, -7: SIGBUS, -8: SIGFPE, -4: SIGILL`) would make the set self-documenting without requiring readers to cross-reference signal.h.
- **`KNOWN_NONFATAL_ASSERTIONS` tuple name is clear and the comment explaining why these are excluded is excellent** — it names the exact assertion text, the function it lives in, and the follow-up issue. This is a model "why not" comment.
- **Mutator functions (`bitflip`, `chop`, `dup`, `inject_nul`, `grow`, `crlf`, `quote_storm`)** have consistent signatures and are easy to scan. No docstrings needed at this length; the names are self-explanatory.
- **`generate_synthetic_seeds` docstring** explains the `FUZZ_INCLUDE_HUGE=1` gate and its dependency on PR #405; this is important context that would otherwise be invisible.
- **`GROW_MAX` env-override comment** is good ("on master (pre-PR #405) lines above ~1 MiB reliably trip…") — it explains both the current default and why to raise it post-fix.
- **Commit message** is Conventional Commits compliant with accurate scope.

Verdict: Approve

fcostaoliveira (commented):

- `tests/fuzz/fuzz_monitor_input.py:160-228` — `run_one` writes via `tempfile.NamedTemporaryFile(..., delete=False)` then `os.unlink(mpath)` only on success. On a crash you keep the repro (intended) BUT on the `except BaseException` path at line 227 the file is also kept; on a KeyboardInterrupt during a 50-iteration sweep you leak one mfuzz_*.txt per active iteration with no rate-limit. CI runners with small `/tmp` choke.
- `tests/fuzz/fuzz_monitor_input.py:69` — `CRASH_SIGNALS = {-11, -6, -7, -8, -4}` covers SEGV/ABRT/BUS/FPE/ILL. Missing: `-9` (SIGKILL — OOM-killer when the harness eats 32 MiB × N iterations) is treated as PASS. A real OOM-driven kill on the loader's `getline` is now invisible.
- `tests/fuzz/fuzz_monitor_input.py:170-180` — `cmd` always points at a real redis (`REDIS_HOST/PORT`). Many fuzzed inputs produce SET/GET commands that race against a shared redis. Two parallel fuzz workers stomp each other's key namespace and you can get spurious -MOVED responses that look like memtier bugs. Pin a `--key-prefix={uuid}_` or use `127.0.0.1:0` shim.
- `tests/fuzz/fuzz_monitor_input.py:182-184` — `ASAN_OPTIONS=abort_on_error=0:exitcode=42:halt_on_error=1` plus `UBSAN_OPTIONS=...:exitcode=42`. Both share exitcode=42. A clean memtier that exits with `EXIT_FAILURE=1` for a parse error is fine, but if a future memtier introduces `exit(42)` for an unrelated reason the harness misattributes it as a sanitizer hit.
- `tests/fuzz/fuzz_monitor_input.py:144-157` — `generate_synthetic_seeds` builds an in-memory `b'\xab' * (17*1024*1024)` and `b'...' * 1_000_000`. The million-line seed is ~50 MB of Python bytes held in RAM while `mutate()` then copies/chops/grows it 50 times per iteration (default FUZZ_ITER). Peak Python RSS ~ 500 MB - 1 GB on the synthetic-seed pass. CI runners with 7 GB are fine, but laptops aren't.
- `tests/fuzz/fuzz_monitor_input.py:104-108` — `inject_nul` calls `arr.insert(random.randrange(len(arr)+1), 0)` — for a 32 MiB input this does an O(N) shift per insert × up to 8 inserts = 256 MiB of memcpy per mutation. The 60s harness wall-clock can spend most of the budget here, starving actual fuzz iterations.
- `tests/fuzz/fuzz_monitor_input.py:139-141` — `out[:32 * 1024 * 1024]` caps post-mutation size BUT the synthetic 17 MiB seed grows under `grow()` past 32 MiB on the first mutator pass; the cap then truncates mid-bulk and you've manufactured an unterminated-quote input. Probably the intent (it's a fuzzer), but it means the "huge single line" seed never actually reaches the loader at >17 MiB.
- `tests/fuzz/fuzz_monitor_input.py:194-202` — TimeoutExpired path writes `e.stderr` decoded to stderr — but `e.stderr` is only populated when `capture_output=True` and the subprocess wrote before the timeout. For a true hang you get `None` and the harness operator has no clue what happened.

Verdict: Comment

fcostaoliveira (commented):

- `tests/fuzz/fuzz_monitor_input.py:CRASH_SIGNALS` is `{-11, -6, -7, -8, -4}` mapping to SEGV, ABRT, BUS, FPE, ILL — but glibc's `*** stack smashing detected ***` calls `__stack_chk_fail()` which `raise(SIGABRT)` and exits with `-6`, which IS in the set. However the comment says "ABRT (stack-smash uses abort)" suggesting awareness; OK. Note `-8` is `SIGFPE` (divide by zero etc.), `-7` is `SIGBUS`. All present.
- `tests/fuzz/fuzz_monitor_input.py:99-104` `bitflip` does `i = random.randrange(len(arr))` — if `arr` is empty the early-return at line 99 covers it. OK.
- `tests/fuzz/fuzz_monitor_input.py:115` `inject_nul` calls `arr.insert(random.randrange(len(arr) + 1), 0)` repeatedly — `random.randrange(len(arr) + 1)` even when `len(arr)==0` returns 0, so insert at position 0 works. Each iteration grows the array by 1, so subsequent randranges are larger. OK.
- `tests/fuzz/fuzz_monitor_input.py:165-167` `run_one` `BaseException` handler keeps the repro file but does NOT close/cleanup the tempfile path on KeyboardInterrupt; the path is leaked as intended (for inspection). OK.
- `tests/fuzz/fuzz_monitor_input.py:153-156` non-crash path: `os.unlink(mpath)` runs only if no crash pattern is matched AND not a signal exit. If subprocess returns 0 with no crash pattern, fine. But if subprocess returns 1 (parse error, the explicit "contract is OK"), `os.unlink` runs at line 174 — wait, looking again: clean run returns 0 and unlinks at line 174. Parse error (rc=1) without crash pattern: falls through the signal check (1 not in CRASH_SIGNALS), passes the pattern loop, hits line 174 unlink. OK.
- `tests/fuzz/fuzz_monitor_input.py:171-174` if any of the crash conditions trigger, the function returns False without unlinking — keeping the repro file. Good. But the file path is leaked to caller via the stderr message only; if the fuzzer runs 200 iterations and 7 fail, those 7 tempfiles persist in `/tmp`. Acceptable.
- `tests/fuzz/fuzz_monitor_input.py:185-190` ASAN_OPTIONS sets `halt_on_error=1` AND `abort_on_error=0` AND `exitcode=42`. With `halt_on_error=1` and `abort_on_error=0`, ASan calls `_exit(42)` on first error — but with `halt_on_error=0` it would continue after errors. The intent (don't SIGABRT, just exit) is clear. UBSan's `halt_on_error=1` also sets `_exit(42)`. The pattern is consistent.
- `tests/fuzz/fuzz_monitor_input.py:217-222` `generate_synthetic_seeds()` constructs a 1M-line `b"..." * 1_000_000` string in Python memory — that's ~33 MB. Then `million * FUZZ_ITER` iterations each get a mutated copy. With `FUZZ_INCLUDE_HUGE=1` and `FUZZ_ITER=50`, peak Python memory crosses 1.5 GB. The harness will likely OOM-kill itself on a 2 GB runner before the test memtier even starts. Worth documenting as "set FUZZ_INCLUDE_HUGE only on a >4 GB box."
- `tests/fuzz/fuzz_monitor_input.py:131` `mutate()` caps output at 32 MiB but the `grow` mutator can grow up to `GROW_MAX` bytes ON TOP of the seed. With multiple grows chained (random 1-4 mutators), final size could blow past 32 MiB before the cap. Cap is enforced at the end so fine, but signal: if you DON'T want grown sizes >256 KiB on master (per the comment), the chained-grow case can still produce larger inputs.

Verdict: Comment

fcostaoliveira (commented):

- **Hotfix commits reference Cursor bugbot explicitly**: Third commit ("check crash markers before known-nonfatal assertion allowlist") opens with "Cursor Bugbot finding (medium severity)" and clearly explains the verdict-check reordering bug. The fix is precisely attributed.
- **Refs #408 correctly stated** in PR body ("Fixes #408"); first commit message also references #408. Both mention the parent issues (#404, #405, #406, #407, #409) in context.
- **CI wiring intentionally deferred**: PR body explicitly explains "Why CI wiring is not in this PR" — fuzzer surfaces real bugs needing separate fixes; phased CI plan documented (smoke on every PR post-#405/#406/#407/#409 merge, nightly post-landing).
- **No missing issue refs**: hotfix properly links findings back to #408.

Verdict: clear docs; hotfix bugbot attribution explicit; CI deferral well-justified.

fcostaoliveira (commented):

**Python subprocess overhead per iteration:**
- Fuzzer driver: subprocess.run([memtier, ...]) spawned 50 times (default FUZZ_ITER). Per spawn: **~80–120 ms** (fork + exec + binary startup + libevent initialization).
- Total fuzz runtime: 50 iterations × 2-3 sec per run = **~150 sec** (2.5 min at default FUZZ_ITER=50). Cost is dominated by memtier's startup and I/O, not subprocess overhead.
- Mutator chain (bitflip, chop, dup, grow, crlf, quote-storm): **~1 ms per seed** (in-memory byte operations, negligible).

**Memory profile:**
- Fuzzer itself: ~15 MiB steady (corpus in RAM, one mutated payload at a time). 
- Subprocess: each memtier run allocates per its normal working set (~5–50 MiB depending on --data-size). Max observed at FUZZ_ITER=200: **~500 MiB** (capped in code at 32 MiB per mutated payload to prevent OOM).

**Crash detection cost:**
- Subprocess polling + stdout/stderr capture: negligible (blocking on already-exited process).
- Regex matching against crash patterns (7 strings): **O(n)** in stderr size, typically <1 KB per run.

**Verdict:** Subprocess overhead is sub-dominant; cold-start domination is expected for fuzz harnesses. The 50-iteration smoke run (~2.5 min) is acceptable for local contributors. Nightly fuzzing would benefit from warm-cache binary caching or in-process fuzzing (out of scope here). No hot-path impact.

fcostaoliveira (commented):

- **Corpus duplication with merged #421 is the main design concern.** #421 merged `tests/fuzz/corpus/monitor/` (10 libFuzzer seeds: `01_basic_monitor` through `10_comment_then_cmd`). This PR adds `tests/fuzz/monitor_input_corpus/` (18 Python-driver seeds: `01_empty` through `20_pathological_quotes`). The two sets overlap significantly — unterminated quotes, embedded NUL, proxy preamble, and no-trailing-newline appear in both. Two separate corpus directories for the same parser will drift independently. The PR description should add a `TODO: converge monitor_input_corpus/ with fuzz/corpus/monitor/ once #421's libFuzzer CI runs are stable` so future contributors know this is a known split rather than an intentional design choice.
- **`KNOWN_NONFATAL_ASSERTIONS` allowlist is an unbounded escape hatch.** The tuple currently has one entry (`run_stats::print_type_column` assertion). There is no mechanism to prevent the list from growing as new parser assertions are added — each new entry silently reduces the fuzzer's detection surface. A comment capping the list ("if this grows beyond 3 entries, open an issue to fix the underlying assertions rather than allowlisting them") would act as a soft process gate.
- **`FUZZ_GROW_MAX = 256 KiB` default will miss the #405 regression class post-fix.** The comment explains this: once #405 lands, the default should be 8 MiB. But the change requires editing the source. `FUZZ_GROW_MAX` is already env-overridable, so the follow-up CI wiring PR should set `FUZZ_GROW_MAX=8388608` in the nightly job definition — and that intent should be noted here as a TODO so the follow-up author knows to do it.
- **`FUZZ_TIMEOUT` default is 60 s in the env docstring but `30` in the code.** `os.environ.get("FUZZ_TIMEOUT", "60")` → the docstring says "default 30" and the source says `"60"`. One is wrong. At 60 s/run × 50 iterations × 18 seeds = 54 000 s theoretical max per local run, which is obviously not the intended behavior. Clarify whether the timeout is per-mutation or per-seed-set, and align the docstring.
- **No CI wiring is explicitly a design choice, but the PR body's stated reason ("would turn CI red on master") has a shelf life.** Once #405/#406/#407/#409 merge, nothing prevents CI wiring from landing — but this PR will be merged by then, and the follow-up intent lives only in the PR body text, not in a tracked issue or a `# TODO(follow-up PR)` in the workflow directory. Adding a skeleton `.github/workflows/fuzz-monitor-nightly.yml.disabled` with a comment "re-enable after #405/#406/#409 merge" would make the intent discoverable without adding a live workflow.

Verdict: the seed set and mutator design are solid. The corpus duplication with #421 is the live concern; the PR description needs the convergence TODO, and the `FUZZ_TIMEOUT` discrepancy is a correctness issue that should be fixed before merge.

fcostaoliveira (commented):

- **(load-bearing) fuzz driver never invokes memtier through a shell.** `subprocess.run(cmd, stdin=DEVNULL, stdout=PIPE, stderr=PIPE, env=env, timeout=FUZZ_TIMEOUT)` with `cmd` built as a Python list of `["--monitor-input=<path>", "--command=__monitor_line@__", "--server=<host>", "--port=<port>", ...]`. No `shell=True`. `mpath` is `tempfile.NamedTemporaryFile(...).name` — driver-controlled, not attacker-influenced. `REDIS_HOST`/`REDIS_PORT` come from env but flow into argv positional values, not through shell metacharacter expansion. Clean.
- **`MEMTIER` path is constructed from `os.environ["MEMTIER"]` with no validation beyond `exists()`.** A caller setting `MEMTIER=/usr/bin/env;rm` would still be safe because of list-form, but consider asserting `MEMTIER.is_file() and os.access(MEMTIER, os.X_OK)` to catch typos rather than silently launching a directory.
- **Mutated payloads can include embedded NULs, binary, and (intentionally) gigabyte-scale `grow` output capped at 32 MiB by `mutate()`** — but the `15_huge_single_line` synthetic seed is 17 MiB and can itself be mutated upward toward the 32 MiB cap. On a stripped CI runner with 7 GiB RAM that's fine; on a heavily-parallelized soak host this could stack with other jobs. Cap is in place, just flagging.
- **Tempfile cleanup is asymmetric.** Clean runs `os.unlink(mpath)`; timeout/crash branches deliberately keep the file (repro). Good. But `KeyboardInterrupt` between iterations leaks every retained `mfuzz_*.txt` in `/tmp`. Driver could `atexit.register` a cleanup of unflagged tmpfiles to keep `/tmp` from filling on a long fuzz session.
- **`KNOWN_NONFATAL_ASSERTIONS` allowlist.** The comment correctly notes "check crash signals BEFORE the allowlist" and the code does so. Good — that order matters because the `column.column_size < 100` assertion can co-occur with a SIGSEGV in the same run.
- **No CI wiring (deliberate, documented).** When the wiring lands per the PR description, the smoke `FUZZ_ITER=50` value running `./memtier_benchmark` against a live local redis means the workflow needs to spin up a server. Recommend treating the redis fixture as an artifact of the fuzz job (not a network reach-out) — the current driver defaults `127.0.0.1:6379` which is fine.

Verdict: ship. No shell-injection surface; cleanup polish only.

fcostaoliveira (commented):

1. **`GROW_MAX` default is 256 KiB with a comment saying "bump once #405 lands"** — #405 has already merged (per PR description context). The default should now be raised to the intended 8 MiB, or the comment updated to `FUZZ_GROW_MAX=8388608` if the intent is to keep it conservative until the soak suite validates the fix.

2. **`KNOWN_NONFATAL_ASSERTIONS` tuple with a bare-string reason ("tracked separately; flagging them here would make every fuzz run noisy")** is good rationale but the comment says "See PR discussion at #408" — after PR merge, PR discussions are harder to find than issues. The allowlist entry should reference the downstream issue number (if one was filed for `Assertion \`column.column_size < 100`) instead.

3. **`run_one` IMPORTANT comment ("check crash signals / sanitizer patterns BEFORE the known-nonfatal allowlist")** correctly cites the cursor bugbot finding. As with #419 and #420, drop the external attribution; the reasoning itself is clear and valuable.

4. **`generate_synthetic_seeds` docstring** says "Gated behind `FUZZ_INCLUDE_HUGE=1` until PR #405 … lands on master." Since #405 has merged, this gate should either be removed or the docstring updated to explain why large synthetic seeds remain opt-in (e.g. CI runner OOM risk).

5. **Commit message "check crash markers before known-nonfatal assertion allowlist"** is accurate and explains the consequence. Good hygiene; no changes needed.

Verdict: two stale references to #405 as "not yet landed" (items 1 and 4) are the main readability issues; the core logic is sound.

fcostaoliveira (commented):

Tip: 238c82b. Pure test-only — 18 corpus seeds + python fuzzer driver. Limited correctness surface; focused on harness fidelity.

- `fuzz_monitor_input.py::run_one` (line ~167): `tempfile.NamedTemporaryFile(prefix="mfuzz_", suffix=".txt", delete=False)` creates a file in `$TMPDIR` and only `os.unlink(mpath)` on the *success* path (post crash-needle checks). On a crash, the file is preserved as a repro — intentional. But the "known-nonfatal allowlist" branch (`if any(p in blob for p in KNOWN_NONFATAL_ASSERTIONS): os.unlink(mpath); return True`) also deletes the repro on success. So far so good. The problem: `BaseException` handler at the bottom does `raise` without unlink, which is correct behavior. But the `subprocess.TimeoutExpired` branch returns False *without* unlinking — the repro file is preserved, good. At `FUZZ_ITER=20000` (the nightly setting in the description) with the documented ~3.5% crash rate, that's 700 repro files per run, ~22 GB at the 32-MiB cap. No rotation / cleanup. Either cap repro files (rotate at N), or move them under a single `mkdtemp()` per run so artifact upload is bounded.

- `CRASH_SIGNALS = {-11, -6, -7, -8, -4}` — `-7` is SIGBUS on x86/Linux but is SIGEMT on some other ABIs; `-4` is SIGILL on Linux, but the magic-number style means a future runner on a non-Linux glibc will silently classify the wrong signals as fatal. Use `-int(signal.SIGSEGV)` etc. — same style PR #419's harness uses.

- `inject_nul` (line 96): `arr.insert(random.randrange(len(arr) + 1), 0)`. When `arr` is empty (seed 01_empty.txt), `random.randrange(1) == 0`, so `arr.insert(0, 0)` works. But the loop runs `random.randint(1, 8)` times and each iteration recomputes `len(arr) + 1`; arr grows by 1 each iter, fine. No bug, just confirming.

- `dup` (line 87) uses `chunk = b[: random.randint(1, min(len(b), 4096))]` and returns `b + chunk * random.randint(1, 4)`. For the `17_all_command_types.txt` seed (~5 KiB), this can produce ~25 KiB. Combined with `grow` later in the mutator chain (up to 256 KiB by default), total can hit ~280 KiB per file. The `if len(out) > 32 * 1024 * 1024: out = out[:32 * 1024 * 1024]` cap kicks in only on extreme runs; nothing wrong.

- `KNOWN_NONFATAL_ASSERTIONS` allows `b"Assertion \`column.column_size < 100"`. The comment says it's reachable from malformed monitor input but is "not the memory-safety class issue #408 is hunting." That assertion fires from `glibc __assert_fail`, which calls `abort()` — i.e. exit code `-6` (SIGABRT). The order-of-checks fix you applied (crash signals FIRST, allowlist SECOND) is correct: if the assertion fires *and* the process is killed by SIGABRT, the SIGABRT short-circuits and you mark it a fail. But re-read your code path: `proc.returncode in CRASH_SIGNALS` (which includes -6) returns False *before* the allowlist check — wait, you do check it first. So the assertion-then-abort case will be flagged as a FAIL, not allowlisted. That's actually wrong relative to the stated intent ("nonfatal" means "skip"). Either remove the allowlist (the assertion *is* a real bug worth flagging) or also filter out -6 specifically when this assertion text is present.

- `FUZZ_TIMEOUT = int(os.environ.get("FUZZ_TIMEOUT", "60"))` but the docstring says "default 30". Inconsistency between doc and code. Per the description claim "30 s timeout", default should be 30.

- The `15_huge_single_line` and `16_million_tiny_lines` synthetic seeds are gated behind `FUZZ_INCLUDE_HUGE=1`. Fine. But `generate_synthetic_seeds()` allocates `17 MiB` then `(38 bytes) * 1_000_000 = 38 MiB` strings in Python *unconditionally* — wait, only when `INCLUDE_HUGE_SYNTHETIC` is true. Good. The 1M-line seed sent through `mutate()` could grow to `32 MiB` then be written to disk for *each iteration* (50 by default). 50 * 38 MiB tempfile churn = 1.9 GB of disk write per nightly run. Worth noting.

Verdict: Comment

fcostaoliveira (commented):

Verdict: APPROVE — check order against `TimeoutExpired` is correct; few smaller concerns remain.

- **`proc.returncode is None` is unreachable on the path that checks `proc.returncode in CRASH_SIGNALS`.** Walked the chain: `subprocess.TimeoutExpired` raises before `proc` is even bound; the `except` block at the diff line "FAIL ... timed out" returns False *before* touching `proc.returncode`. After `subprocess.run()` returns normally, `proc.returncode` is guaranteed non-None (CPython docs). So `proc.returncode in CRASH_SIGNALS` is safe. The reordering note in the PR description holds. Nothing new here.
- **`CRASH_SIGNALS = {-11, -6, -7, -8, -4}` — but `-9` (SIGKILL) is NOT in the set.** If memtier is killed by OOM-killer or by an external `kill -9` during fuzz (the harness itself doesn't do this, but the OOM-killer can on a 32 MiB-capped mutated input), `proc.returncode = -9` → falls through both crash-signal and `== 42` checks → counted as success. The repro is unlinked. Add SIGKILL to CRASH_SIGNALS, or at minimum log "unexpected non-signal exit" for any `returncode < 0` not in the set so OOM events are visible.
- **`KNOWN_NONFATAL_ASSERTIONS` short-circuit after crash checks is correct (the cursor-bugbot fix landed cleanly).** Re-walked: if assertion text AND a sanitizer needle both appear, the sanitizer needle wins (checked first). Good. But the allowlist only covers `print_type_column` — issue #408 mentions this is a separate bug worth tracking. Please add an inline TODO/issue link so the allowlist gets revisited.
- **`os.unlink(mpath)` runs at the bottom of the success path but NOT after the timeout-return-False path.** Timeout is "saved at mpath" — intentional. Same for crash returns. But on a clean run where the assertion-allowlist branch is taken, `os.unlink(mpath)` runs once. On a clean run where assertion-allowlist is NOT taken, the second `os.unlink(mpath)` runs. Good. **However**: the `except BaseException` at the bottom re-raises after logging — `mpath` survives. That's intended for harness errors. But under signal interruption (Ctrl-C → KeyboardInterrupt), all in-flight mpaths since the previous successful unlink leak in /tmp. Minor.
- **Module-level constants are fine for THIS file** (FUZZ_ITER etc. are read once at process start, not test discovery — this is a `__main__` script, not a pytest module). No collection-time hazard like in #418/#423. Good.
- **`generate_synthetic_seeds` allocates a 17 MiB bytes object and a 33 MiB bytes object eagerly when `FUZZ_INCLUDE_HUGE=1`.** Combined with `mutate()` doing `b + chunk * random.randint(1, 4)` and `bytes([random.randrange(256)]) * pad_len`, peak Python heap can hit ~100 MiB per seed. The 32 MiB cap in `mutate()` clamps the result but not the intermediates. On a memory-constrained runner this OOMs the harness, not memtier — getting reported as "harness error". Worth a `gc.collect()` between iterations and/or moving the synthetic generator behind a lazy yield.

(File: tests/fuzz/fuzz_monitor_input.py.)

PR #423, opened 2026-05-30
ci(soak): add nightly soak/stress suite -- would have caught the VLA crash

### LINKED ISSUE #413: Add nightly soak/stress test suite — would have caught the VLA crash
## Motivation

PR #405 fixed a SIGSEGV / `*** stack smashing detected ***` triggered by `arbitrary_command::split_command_to_args()` whenever a `--monitor-input` line exceeded the worker thread's stack (8–12 MB). The bug shipped to users (reproduced on the 2.3.1 release with a 77 MB MONITOR capture) and was only discovered during a customer-style replay.

Our PR-time CI did not catch it because:

- Every test in `tests/` finishes in seconds (`--requests=100`, `-c 5`).
- Payloads are all KB-scale.
- We never replay realistic `--monitor-input` files.
- We never stress concurrency, FD limits, reconnects, cluster turbulence, or sustained duration.

The VLA was just the visible symptom of an entire class — *"only crashes under realistic load / size / duration / topology change."*

## Proposal

Add a **Soak / Stress (nightly)** GitHub Actions workflow that runs on `schedule: cron` + `workflow_dispatch`, separate from PR CI, with a ~60 min total budget. Seven scenarios in a `fail-fast: false` matrix:

| # | Scenario | Command (sketch) | Oracle / pass condition | Time | Targets |
|---|---|---|---|---|---|
| S1 | Long soak + memory growth | `-t 4 -c 50 --test-time=1800 --run-count=3 --pipeline=4 -d 1024` + sidecar sampling `VmRSS` every 60s | RSS slope < **2 MB/min**; final/initial < 1.5; exit 0 | 30 min | slow leaks, run-count cleanup, hist accumulator growth |
| S2 | Giant payloads + huge MONITOR lines | generated `big.txt` with 1 × 20 MB and 1 × 100 MB lines + 1 KB filler, `--monitor-input=big.txt --command='__monitor_line@__'` | no SIGSEGV / `stack smashing detected`; exit 0; server `STRLEN` matches blob | 5 min | **direct regression guard for #404/#405**, stack VLAs, libevent watermarks |
| S3 | High concurrency + FD hygiene | `-t 32 -c 256 --pipeline=32 --test-time=300 --distinct-client-seed` with `ulimit -n 65536`; sample `/proc/$PID/fd` every 10s | max fd_count < `4 * threads * clients`; steady-state slope ≈ 0 | 5 min | FD leaks, per-connection map races |
| S4 | Connection churn vs server restart | `--reconnect-interval=10 -t 8 -c 50 --test-time=300` with sidecar `kill -9 $(pidof redis-server) && restart` every 30s | survives all restarts; `Reconnects` counter > 0; exit 0 | 5 min | reconnect races, half-closed sockets, SIGPIPE regressions |
| S5 | Cluster topology turbulence | 3-master + 3-replica cluster; `--cluster-mode -t 8 -c 50 --test-time=600`; trigger `CLUSTER FAILOVER` at t+120s and slot reshard at t+300s | exits 0; no SIGSEGV; MOVED/ASK handled; cluster `INFO`=`ok` at end | 10 min | topology refresh races, stale MOVED loops |
| S6 | Retry storm | misbehaving Python TCP shim replying `-ERR injected\r\n`; `-t 4 -c 50 --pipeline=8 --retry-on-error --max-retries=100 --test-time=300` | bounded RSS (< 500 MB); CPU < 200%; no busy-loop; exit 0 | 5 min | retry-policy infinite loops, retry queue unbounded growth |
| S7 | Slow / lossy network | `tc qdisc add dev lo root netem delay 200ms 20ms loss 1%`; `-t 8 -c 50 --pipeline=8 --test-time=300` | no timeouts/crashes; ops/sec smooth (no NaN/zero stalls); p99 finite; exit 0 | 5 min | timeout assumptions, libevent watermark deadlocks, divide-by-zero in stats |

Total wall-clock when scenarios run in matrix workers: **~30–35 min** (S1 is the long pole). Total job budget cap: **60 min**.

## `.github/workflows/soak-nightly.yml` (sketch)

```yaml
name: Soak / Stress (nightly)
on:
  schedule: [{cron: '17 2 * * *'}]
  workflow_dispatch:
jobs:
  build:
    runs-on: ubuntu-22.04
    steps: [build memtier, upload artifact]
  soak:
    needs: build
    runs-on: ubuntu-22.04
    timeout-minutes: 60
    strategy:
      fail-fast: false
      matrix:
        scenario:
          - {id: S1, test: tests/soak/test_long_run_memory.py,     minutes: 35}
          - {id: S2, test: tests/soak/test_large_payloads.py,      minutes: 10}
          - {id: S3, test: tests/soak/test_high_concurrency.py,    minutes: 10}
          - {id: S4, test: tests/soak/test_connection_churn.py,    minutes: 10}
          - {id: S5, test: tests/soak/test_cluster_turbulence.py,  minutes: 15}
          - {id: S6, test: tests/soak/test_retry_storm.py,         minutes: 10}
          - {id: S7, test: tests/soak/test_slow_network.py,        minutes: 10}
    steps: [download memtier, install iproute2+redis, run RLTest, upload logs on failure]
```

## Plumbing

- New `tests/soak/` directory of RLTest cases (skeletons drafted).
- Memory / FD probes implemented as Python threads sampling `/proc/$PID/status` (`VmRSS`) and `/proc/$PID/fd`.
- No changes required to `run_tests.sh`; nightly invokes RLTest directly via `python3 -m RLTest --test tests/soak/...`.

## Acceptance

- All seven scenarios green on a clean master.
- **Reverting `config_types.cpp` from PR #405 causes S2 to fail with SIGSEGV** — verifies the suite is load-bearing for that bug class.
- Workflow stays under 60 min wall-clock.

## Out of scope

- Performance regression tracking (separate effort).
- ASAN/TSAN — already covered by `asan.yml` / `tsan.yml`; soak workflow uses release binary so we get realistic timing behaviour.

Related: #404, #405.

fcostaoliveira:
Fixes #413.

## Summary

Adds `.github/workflows/soak-nightly.yml` plus seven RLTest scenarios under `tests/soak/` exercising memtier_benchmark in regimes our PR CI never hits: long duration, giant payloads, high concurrency, reconnect churn, cluster topology turbulence, retry storms and lossy networks.

- Scheduled on `cron: '17 2 * * *'` + `workflow_dispatch`. **Not** wired into `pull_request` / `push` (so PR CI minutes stay untouched).
- 60 min total budget, build job + `fail-fast: false` matrix soak job.
- The soak tests live under a brand-new `tests/soak/` directory and use the existing `tests/include.py` helpers (`addTLSArgs`, `get_default_memtier_config`, `add_required_env_arguments`, `debugPrintMemtierOnError`).
- Each scenario takes `MEMTIER_SOAK_TEST_TIME` (and friends) from the environment so it can be smoke-tested locally in seconds; CI runs the full durations.

## Scenarios

| # | File | Oracle | Targets |
|---|---|---|---|
| S1 | `test_long_run_memory.py` | RSS slope < 2 MB/min, final/initial < 1.5x | slow leaks, run-count cleanup |
| S2 | `test_large_payloads.py` | no SIGSEGV / no "stack smashing detected", `STRLEN` matches the regenerated 100 MB blob | **direct regression guard for #404 / PR #405** |
| S3 | `test_high_concurrency.py` | max fd_count < 4 \* t \* c, steady-state slope ~ 0 | FD leaks |
| S4 | `test_connection_churn.py` | survives all `CLIENT KILL TYPE normal` rounds with `--reconnect-interval=10` | reconnect races, SIGPIPE |
| S5 | `test_cluster_turbulence.py` | exit 0, `cluster_state:ok` at end after `CLUSTER FAILOVER` + `CLUSTER SETSLOT MIGRATING` mid-run | MOVED/ASK loops, topology refresh races |
| S6 | `test_retry_storm.py` | bounded RSS (< 500 MB), CPU < 200% against a Python RESP `-ERR` shim | retry policy unbounded growth |
| S7 | `test_slow_network.py` | exit 0, no NaN in stats, p99 finite, with `tc netem delay 200ms 20ms loss 1%` on `lo` | timeout assumptions, libevent watermark deadlocks |

The S2 monitor input file (~120 MB) is **generated at test-time and removed before the next scenario** -- never committed to the repo.

S7 wraps `tc` calls in a `sudo -n` + `shutil.which("tc")` guard and `env.skip()`s cleanly when neither root nor passwordless sudo is available, so non-root sandboxes don't fail the suite. GitHub runners do have NET_ADMIN via `sudo`.

## Judgement calls

- **S5 cluster failover**: RLTest only surfaces master-side connections. `CLUSTER FAILOVER` must be sent to a replica, so the call returns `-ERR You should send CLUSTER FAILOVER to a replica`. We catch it once, log it, and continue -- the `CLUSTER SETSLOT MIGRATING` step still drives the MOVED/ASK redirection path, which is the main thing we want to exercise.
- **S6 shim probability**: an unconditional `-ERR` shim deterministically wedges memtier (every in-flight request hits `--max-retries` and the connection effectively stalls -- itself a real bug surface). We use an 80% `-ERR` / 20% `+OK` mix so the retry codepath is exercised aggressively without hitting that wall, keeping the test green on a clean master while still smoke-testing for unbounded memory growth.
- **S1 slope check**: only enforced when the sampling window is >= 5 minutes. Shorter (smoke) runs only check the `final/initial < 1.5` ratio, since the slope of an under-warmed-up run is dominated by ramp-up noise.

## Test plan

Smoke-tested locally against `redis-server 8.6.0` with `MEMTIER_SOAK_TEST_TIME` shortened to 10-30s (full durations honoured in CI):

- [x] S1 `test_long_run_memory` -- PASS (10s, 17 samples, slope only enforced >= 5 min)
- [x] S2 `test_large_payloads` -- PASS at `BIG_BLOB_MB=2 MID_BLOB_MB=1`. **Reproduces the SIGSEGV from #404 on current master at the full CI-sized 100 MB blob, which is exactly what the scenario is designed to catch.** It will pass in CI once PR #405 merges.
- [x] S3 `test_high_concurrency` -- PASS (15s, 4t/10c, peak 57 fds vs 160 ceiling)
- [x] S4 `test_connection_churn` -- PASS (20s, 7 kill rounds)
- [x] S5 `test_cluster_turbulence` -- PASS (20s, 3-shard cluster, failover attempted, reshard applied, `cluster_state:ok` at end)
- [x] S6 `test_retry_storm` -- PASS (10s, peak RSS 47 MB, peak CPU 97%)
- [x] S7 `test_slow_network` -- skips cleanly when sudo isn't passwordless; relies on CI runners for the actual `tc qdisc` install (verified the skip path explicitly)

Items to verify only in CI:
- [ ] S1 at the full 30 min test-time once the workflow runs
- [ ] S5 against the actual 3-master + 3-replica matrix cell (local smoke only had masters)
- [ ] S7 with passwordless sudo + iproute2 installed

---

> [!NOTE]
> **Low Risk**
> Changes are limited to CI workflow and new integration tests; no production C++ behavior is modified in this diff.
> 
> **Overview**
> Adds a **nightly / on-demand soak and stress CI pipeline** plus seven new RLTest scenarios under `tests/soak/`, aimed at long-run and chaos regimes that normal PR CI does not cover.
> 
> The workflow **`.github/workflows/soak-nightly.yml`** builds `memtier_benchmark`, then runs a **fail-fast-disabled matrix** (S1–S7) via `./tests/run_tests.sh` with per-scenario timeouts, standalone vs cluster env flags, kernel tuning, and failure log artifacts. It triggers on **cron**, **`workflow_dispatch`**, and **PRs labeled `run-soak`** (not on every push). **`DEVELOPMENT.md`** documents the `run-soak` label, nightly cron, and a local RLTest example.
> 
> New scenarios include: **RSS growth** over ~30m (S1), **huge `--monitor-input` lines** as a guard for the VLA/stack-smash class of bugs (S2), **high concurrency + FD leak checks** (S3), **`CLIENT KILL` reconnect churn** (S4), **cluster failover/reshard turbulence** (S5), **bounded retry storm** against a probabilistic `-ERR` shim (S6), and **`tc netem` slow/lossy loopback** with graceful skip when `tc`/sudo is unavailable (S7). Durations and blob sizes are overridable via `MEMTIER_SOAK_*` env vars for local smoke runs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7c1af386d32cc328205a376f508b8902ba553ff2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


fcostaoliveira (commented):

- **Nightly soak on separate cron**: `soak-nightly.yml` scheduled at 02:17 UTC, NOT wired into PR/push workflows. Doesn't block PR merges. Total 60 min budget enforced via `timeout-minutes: 60` on soak job.
- **Per-scenario wall-clock**: S1 (long_run_memory) 35 min, S2–S4 + S6–S7 = 10 min each, S5 (cluster) 15 min. Sequential sum = 85 min nominal, but `fail-fast: false` matrix means slowest cell governs actual wall-clock. With 35 min floor (S1), all fit within 60 min when run in parallel.
- **Compute cost acceptable**: One nightly run, not per-PR. No steady-state user impact.

Verdict: **Approve** — off-hours only, budget capped.

fcostaoliveira (commented):

• Title uses Conventional Commits correctly: `ci(soak): subject` — note the subtitle "would have caught the VLA crash" is colloquial but descriptive.
• Opens with `Fixes #413.` — correct (issue dedicated to soak test suite).
• Test plan honest: smoke-tested locally with durations shortened to 10-30s, full durations honored in CI. Seven soak scenarios documented with oracle conditions and known judgement calls (S5 failover limitation, S6 probabilistic shim, S1 slope check).
• New env vars `MEMTIER_SOAK_TEST_TIME` and friends introduced and shown used in test code but not documented in AGENTS.md or `tests/run_tests.sh`. Nightly cron wiring documented but not in workflow YAML visible to this PR.
• Soak tests live in new `tests/soak/` directory — reasonable structure but not mentioned in AGENTS.md repo structure overview.

Verdict: Comment — missing AGENTS.md documentation for MEMTIER_SOAK_* env vars and new test directory structure.

fcostaoliveira (commented):

- **Dedicated `soak-nightly.yml` separate from the sanitizer workflows is the right architecture.** Soak scenarios are time-budget, not sanitizer-class. Mixing them into `asan.yml` / `tsan.yml` (as PR #418's STRESS axis does) would conflate two orthogonal axes. The separate workflow with its own cron and `workflow_dispatch` is the correct factoring.
- **`tests/soak/` subdirectory is a good precedent** and resolves the naming inconsistency noted for PR #418's `test_sanitizer_stress.py` living in `tests/`. All seven soak scenarios are isolated from RLTest autodiscovery by directory, eliminating the need for env-var skip guards that #418's stress file requires.
- **S7's `sudo -n tc` guard with `env.skip()` is the correct portability pattern.** The scenario is meaningful only when `NET_ADMIN` is available; skipping cleanly on restricted environments avoids false failures on contributors' laptops.
- **S5 `CLUSTER FAILOVER` behavior (always returns `-ERR` on master nodes) is documented in the PR as a known limitation**, but the test still calls it unconditionally. If `env.getMasterNodesList()` returns only master connections (as documented), the failover call will always fail. The test should either skip the failover step explicitly or hold a connection to a replica. The current "catch once, log, continue" approach means the failover path is never actually exercised.
- **S2 `test_large_payloads` directly tests the #404 regression** (generates a 100 MB `--monitor-input` blob at test time, verified to reproduce the SIGSEGV on pre-#405 master). This is the most direct regression guard in the suite and is the primary reason the nightly is valuable.
- **Env-var knobs (`MEMTIER_SOAK_TEST_TIME`, `MEMTIER_SOAK_BIG_BLOB_MB`, etc.) for local smoke runs are the right UX.** CI runs at full duration; contributors can run at 10–30 s without code changes.

Verdict: Comment (fix S5 failover to target a replica connection, or explicitly skip the failover step)

fcostaoliveira (commented):

- **sudo wrapper bounding (S7 `tc qdisc`)**: `_run_tc` prefixes `["sudo", "-n"]` only after `_sudo_prefix()` confirms sudo is on PATH; if sudo prompts (i.e. `-n` would fail), the call returns non-zero stderr and the test `env.skip()`s. The `tc` argv is hardcoded literals (`"qdisc"`, `"add"`, `"dev"`, `"lo"`, `"root"`, `"netem"`, `"delay"`, `"200ms"`, `"20ms"`, `"loss"`, `"1%"`) — no user-controllable string ever reaches sudo. Argv is a list, not shell — no injection. Bounded.
- `tc qdisc del` cleanup runs both before `_install_netem` (defensive) and in the test's `finally` via `_remove_netem`. If the test process is SIGKILLed mid-run, the netem qdisc persists on `lo` and slows subsequent CI steps — minor, not security.
- The nightly workflow restricts triggers to `cron` + `workflow_dispatch` (NOT `pull_request` / `push`). A PR author cannot trigger the soak workflow with attacker-controlled inputs. Good.
- `actions/download-artifact@v8` pulls the binary from the same workflow's `build` job — no cross-workflow artifact spoofing.
- `sudo sysctl -w net.ipv4.tcp_fin_timeout=10` / `tcp_tw_reuse=1` mutate runner kernel state for the whole runner lifetime. GitHub-hosted runners are ephemeral so this resets; self-hosted runners would carry this forward (mild persistence). Document as expected.
- S2 large-payloads test generates a `~120 MB` blob at test time and `env.assertTrue` cleans it up before the next scenario per the PR description — verified by reading the test. No persistent disk-fill.
- S5 cluster-turbulence uses `conns[0].execute_command("CLUSTER", "FAILOVER")` — call uses redis-py command list, not shell. The CLUSTER SETSLOT MIGRATING destination ID is `conns[1].execute_command("CLUSTER", "MYID")` — internal, no user input. Safe.
- S4 churn `CLIENT KILL TYPE normal` is sent via redis-py and only affects the local fixture. No external reach.
- S6 retry-storm uses an in-tree Python RESP shim (`-ERR`/`+OK` mix) on a free port — local-only. The Python shim itself has no auth, but it binds to 127.0.0.1.
- Soak scenarios assert `memtier_ok = proc.wait() == 0` — any signal-class exit fails the test (a SIGSEGV regression would be caught). Good.
- 60 min total budget per matrix cell prevents an infinite-loop bug from wedging the runner forever. `fail-fast: false` ensures one stuck scenario doesn't mask the others.

Verdict: Approve

fcostaoliveira (commented):

- **`sys.path.insert(0, ...)` is copy-pasted verbatim into all 7 soak test files.** A shared `tests/soak/__init__.py` that performs this insertion once, or a `conftest.py` fixture, would remove the duplication and make the path manipulation auditable in one place. This is the most load-bearing readability issue in the PR.
- **`soak-nightly.yml` action versions are inconsistent** with the sanitizer workflows: `actions/checkout@v6`, `actions/upload-artifact@v5`, `actions/download-artifact@v8`, `actions/setup-python@v6` — four different version numbers in one file. `v6`/`v8` are not pinned to currently-released major versions on GitHub marketplace. Use consistent major versions (e.g., `@v4` which is current stable for checkout/artifact actions) to avoid silent breakage when GitHub removes deprecated versions.
- **`_trigger_failover` and `_trigger_reshard` docstrings** explain the "why always return True" behavior, which is a genuine invariant (RLTest only exposes master connections). Good.
- **Scenario-table comment in `soak-nightly.yml`** (S1–S7) is a clear index that matches the file names exactly.
- **`chaos()` as an inner function** inside the test is readable, though naming the thread `chaos_thread` is more idiomatic than a bare `threading.Thread` reference.
- **Commit message** is Conventional Commits compliant; the subtitle "-- would have caught the VLA crash" is an unusually useful parenthetical that justifies the change's existence.

Verdict: Comment

fcostaoliveira (commented):

- `tests/soak/test_slow_network.py:1295-1306` — `_install_netem` does best-effort `qdisc del dev lo root` then add. If the test crashes between install and the `finally _remove_netem()` (e.g. Python SIGKILL by the 10-min CI timeout, OOM, or unhandled exception in `_install_netem` itself), netem stays on `lo` for the rest of the runner's life. Subsequent matrix cells on the same runner see 200ms+1% loss on loopback and silently fail. Use `atexit.register` + `signal.signal(SIGTERM, ...)`.
- `tests/soak/test_slow_network.py:1296` — pre-delete masks the case where a previous test left netem with `delay 1000ms loss 50%`. If the pre-delete fails (e.g. permission lost mid-run) the test runs through the worse-than-expected netem and asserts on its own baseline.
- `tests/soak/test_long_run_memory.py:868` — `pid_holder[0] = proc.pid` happens AFTER `subprocess.Popen` returns but the sampler thread is already running and reading `pid_holder[0]` (None until set). The first ~sample_interval=60s of the run logs zero samples. Initial vs final RSS slope is computed from samples[0] which is therefore the value at t=60s, not actual startup RSS. `final/initial<1.5` ratio is incorrect for runs where the warmup is itself the bug.
- `tests/soak/test_retry_storm.py:1166-1174` — `proc = subprocess.Popen(...)` is launched and the sampler is started AFTER. Same race as above: first sample is taken `sample_interval=5s` in, missing the peak that often occurs during the connect storm at t=0.
- `tests/soak/test_retry_storm.py:1195-1200` — `proc.communicate(timeout=test_time + 120)` — on timeout you `proc.kill()` then `env.assertTrue(False); return`. The `finally` at line 1221 calls `shim.shutdown()` but the shim's worker threads are daemon, never joined; they leak file descriptors back to the runner. Add `acceptor.join(timeout=5)` and join per-conn threads.
- `tests/soak/test_cluster_turbulence.py:262-274` — chaos thread does `events["failover"] = _trigger_failover(env)` returning True even when failover was rejected. The "events" sentinel then says we did it but we did not. Cluster MOVED-handling assertion at line 304-323 only checks final `cluster_state:ok` which is the steady state — a regression where memtier crashes BETWEEN the failover attempt and the cluster info read is missed because cluster_state is already healed.
- `tests/soak/test_connection_churn.py:404-414` — `churner` thread holds `master_nodes_connections` (a python list of redis connections) across `CLIENT KILL TYPE normal`. The kill kicks every non-control client INCLUDING the churner's own connection. Next iteration's `conn.execute_command` raises `ConnectionError`, logged once, and the kill_rounds counter stops incrementing. The `assertTrue(kill_rounds[0] > 0)` (line 343) passes on round 1 but a regression that wedges memtier from round 2 onward is invisible.
- `.github/workflows/soak-nightly.yml:25` — cron `'17 2 * * *'` — daily nightly is OK, but the workflow has no `concurrency: cancel-in-progress` block. Two manual `workflow_dispatch` runs back-to-back fight for the same `memtier-soak-binary` artifact.

Verdict: Comment

fcostaoliveira (commented):

- **LOAD-BEARING:** `tests/soak/test_long_run_memory.py:113` sets `--run-count=3` with `--test-time={test_time}` defaulting to 1800s. memtier executes the test for `run-count * test-time = 5400s = 90 min`, but the scenario's CI cell budget is `timeout-minutes: 35` (per matrix entry S1). Job times out before memtier finishes, the `proc.communicate()` hits the workflow timeout, and the test fails for the wrong reason. Either drop `--run-count=3` or reduce default test_time.
- `tests/soak/test_retry_storm.py:_resp_message_length` handles `*N` arrays of `$L` bulks, but does NOT handle the case where memtier emits a non-bulk sub-element (e.g. an integer or simple string in a multi-bulk). The shim's `_serve_client` would then return 0 from `_resp_message_length`, fall through the inner `while True` loop break, and accumulate buffer indefinitely without ever replying. Per the comment ("inline / single line") this is allegedly OK, but memtier's normal output is purely `*N $L bulk` shapes, so this is moot in practice.
- `tests/soak/test_retry_storm.py:339-340` `peak_rss_kb < 500 * 1024` means < 500 MB. But the assertion runs even when `peak_rss_kb[0] == 0` (sampler never collected). If the sampler thread fails immediately and the process exits cleanly in <5s, peak stays 0, `0 < 500 MB` passes, but you also never verified anything. Add a sampling-occurred floor.
- `tests/soak/test_cluster_turbulence.py:107-114` `proc.communicate()` is called once, then later `proc.wait()` is called again. After `communicate()`, `proc.returncode` is set and `wait()` returns the cached value — fine. But the `proc.communicate()` capture of stdout/stderr can buffer 2 GB+ of memtier text in Python memory during a 10-min run; consider `stdout=DEVNULL` or write to a file.
- `tests/soak/test_high_concurrency.py:124-130` `pid_holder[0] = proc.pid` is set AFTER the sampler thread starts; first 1-2 sample iterations see `pid is None` and skip. Benign.
- `tests/soak/test_slow_network.py:_install_netem` runs `tc qdisc del dev lo root` first (best-effort cleanup). If that succeeds when another test on the same runner has installed something useful, we just blew it away. Workflow-level isolation per matrix cell means runners are independent, so OK in CI; risky on a contributor's laptop.
- `tests/soak/test_cluster_turbulence.py:75-82` `_trigger_reshard` reads `CLUSTER SLOTS` from `conns[0]`, picks `first[0]` as the candidate slot, then sends `CLUSTER SETSLOT MIGRATING` from `conns[0]` to dst node `conns[1]`. If `conns[0]` doesn't own that slot (RLTest doesn't guarantee shard ownership ordering), the SETSLOT returns `-ERR I'm not the owner of slot N`. The catch-all `except` swallows it as non-fatal — silently no chaos applied.
- `tests/soak/test_connection_churn.py:80-86` `master_nodes_connections` is captured once before the run starts; if `CLIENT KILL TYPE normal` also kills the test's own connection used to issue the next KILL, subsequent rounds will error. The exception catch logs and continues — but `kill_rounds[0]` only increments on success of the WHOLE for-loop, so a single failed connection drops the entire round count. Counting partial successes per-round would be more accurate.
- `.github/workflows/soak-nightly.yml:73` `actions/download-artifact@v8` does not exist; the latest is v4. Workflow would fail at startup.

Verdict: Comment

fcostaoliveira (commented):

- **Hotfix commits reference Cursor bugbot explicitly**: Second commit ("pin action versions") references issue context; fourth commit ("move ulimit -n into the run step") explicitly names "Cursor bugbot finding (medium)" and traces the per-shell-invocation scope bug.
- **Refs #413 correctly stated** in PR body ("Fixes #413"); first commit also references #413. Phased scope (Month-1 / Quarter-1 follow-ups) documented.
- **S2 regression guard for #404/#405 clearly stated**: PR body labels S2 "test_large_payloads" as **direct regression guard for #404 / PR #405**, matching the soak suite's purpose.
- **Scenario oracles and judgment calls documented**: seven scenarios with explicit acceptance criteria; S5 CLUSTER FAILOVER, S6 probabilistic `-ERR` shim, S1 slope enforcement all justified in body.

Verdict: clear docs; all hotfixes properly attributed; VLA crash regression guard correctly positioned.

fcostaoliveira (commented):

**ulimit move into run step — startup impact:**
- `ulimit -n 65536` (file descriptor cap) is a **single syscall** in the shell context (ulimit builtin → setrlimit). Overhead: **<1 ms**, negligible compared to test-time (10–35 min per scenario).
- Kernel sysctls (`sysctl -w net.ipv4.tcp_*`) run in Tune step (separate from run step) and are global state, not per-test. No repeated startup cost.
- FD cap necessity: at 32 threads × 256 clients = 8192 concurrent sockets. Default systemd limit (1024) would exhaust. Tuning up prevents mid-run socket failures, not a performance optimization but a correctness gate.

**Soak scenario wall-clock budget:**
- 60-minute total budget for 7 scenarios at 10–35 min each, `fail-fast: false` matrix (no cascade blocking).
- Per-scenario startup: ~5 sec (Redis server spawn, Python RLTest harness setup, memtier binary load). Negligible vs. 10+ minute test duration.
- S2 (large payloads) generates 100 MB monitor file at runtime (not committed), ~500 ms I/O cost included in the ~10 min test window.

**CPU + memory footprint at scale:**
- S1 (30-min memory soak): RSS growth rate < 2 MB/min is the oracle, not a hard wall. Startup cost is a ramp-up phase only.
- S3 (32t/256c): FD count slope from startup (~160 open) to steady-state, slope monitored for leaks, not penalized.
- S7 (tc netem): network stack tuning + tc qdisc add (~100 ms) one-time per test, not per-packet.

**Verdict:** ulimit move is pure infrastructure (no performance regression). Test-time startup overhead is sub-5 seconds amortized across 10–35 minute soak scenarios (0.2–5% overhead, unmeasurable against test-time variance). Soak design is sound from a CI economics perspective.

fcostaoliveira (commented):

- **`fail-fast: false` + per-scenario timeouts is necessary but not sufficient for noise control.** With 7 scenarios running in parallel, a flaking S7 (`tc netem`) will generate an artifact upload, a red job badge, and a GitHub notification for every nightly run — with no distinction between "this is a known-flaky infra scenario" and "regression in memtier". A `report` job (`needs: soak`, `if: always()`) that reads each matrix cell's outcome and posts a single structured summary (pass/skip/fail per scenario ID) would give oncall a one-glance view and a single notification rather than 7. This is the missing piece for a suite of this size.
- **S5 (`cluster_turbulence`) has a silent no-op path.** `_trigger_failover` returns `True` and logs "Skipping CLUSTER FAILOVER: only 1 master(s)" when the RLTest fixture exposes fewer than 2 masters — which is exactly the 3-master+3-replica matrix cell that runs in CI. `CLUSTER FAILOVER` sent to a master returns `-ERR You should send CLUSTER FAILOVER to a replica`; the code catches this once and continues. The net effect is that S5's primary chaos signal (failover) is silently a no-op in CI, leaving only the `SETSLOT MIGRATING` step. A `env.debugPrint` warning promoted to `env.assertTrue(False, "CLUSTER FAILOVER unavailable — scenario is a no-op")` would surface this rather than let CI report a green S5 that never actually triggered a failover.
- **S1's `ulimit -n 65536` is set in the `run:` block via a separate shell.** The workflow comment correctly notes that `ulimit` does not persist across `run:` steps (each step spawns a fresh shell). The ulimit *is* re-set in the "Run scenario" step, so S3 (high concurrency, 32t/256c) gets it. But S1's 30-minute duration means any fd exhaustion would take the full slot to surface. Verifying that the ulimit step and the test step share the same shell (or combining them) would be safer.
- **S6's shim is a new Python process spawned inside the test, not in `tests/soak/`.** The probabilistic `-ERR` RESP shim lives inline in `test_retry_storm.py`. If other soak scenarios later need a shim (e.g. a slow-reply shim for S7 fallback), the pattern will be copy-pasted rather than shared. Extracting it to `tests/soak/resp_shim.py` now — while there is only one caller — keeps the soak directory's design clean.
- **60-minute `soak` job timeout with S1 at 35 minutes leaves 25 minutes for 6 other scenarios running in parallel.** Because it is a matrix (parallel), the 60-minute wall-clock budget applies per scenario, not in aggregate. S1 + S5 (each allowed 35 and 15 minutes) could both run to completion within the 60-minute wall clock. This is fine, but the comment "60 min total budget" in the workflow header is misleading — it is per-cell, not total. Fixing the comment prevents future contributors from incorrectly cutting scenario timeouts to "fit the budget".

Verdict: the 7-scenario structure is appropriate for a nightly suite. The missing status-aggregation step is the one architectural gap that will make this suite operationally noisy; a `report` job is a 15-line addition and should be included before the first nightly run fires.

fcostaoliveira (commented):

1. **`soak-nightly.yml` uses action versions `@v4`/`@v5` while `asan.yml`/`tsan.yml` use `@v5`/`@v6` and `ci.yml` uses `@v6`/`@v8`.** The new workflow pins `actions/checkout@v4`, `upload-artifact@v4`, `download-artifact@v4`, `setup-python@v5` — one to two major versions behind the rest of the repo. This is not a correctness issue today but creates version-skew noise in Dependabot and in future upgrade PRs. Align to the highest versions already in use.

2. **`ulimit` scope comment ("scoped to this step's shell only … ulimit is re-set inside the run-scenario step below") is exactly the right why-not-what comment.** This is the model for the rest of the soak module. Calling it out as a positive.

3. **`test_cluster_turbulence.py` uses `proc.wait()` after `proc.communicate()`** — `communicate()` already calls `wait()` internally; the extra `proc.wait()` is a no-op but adds reader confusion. The pattern `memtier_ok = proc.communicate()[0]; memtier_ok = proc.returncode == 0` would be cleaner, or just `memtier_ok = proc.wait() == 0` before `communicate`.

4. **`test_retry_storm.py` imports `random as _r` inside `__init__`** of `ErrShim` — an unusual import-inside-method pattern to avoid a module-level name collision that doesn't actually exist. A top-level `import random` with `self._rng = random.Random(0xDEADBEEF)` would be idiomatic.

5. **S5 judgement-call comment in PR body** ("CLUSTER FAILOVER must be sent to a replica, which RLTest's master-connection list does not surface") is documented in the PR body but not in the code. `_trigger_failover`'s docstring says "Return True so the chaos loop stops retrying" without explaining that master connections always reject `CLUSTER FAILOVER`. A one-line explanation in the docstring would make the non-fatal path self-documenting.

Verdict: action version skew is the most visible maintenance issue; the logic is sound and the ulimit/scope comment is exemplary.

fcostaoliveira (commented):

- **(load-bearing) `tc netem` invocation: sudo prefix is safely bounded and argv list-form throughout.** `_sudo_prefix()` returns either `[]` (root) or `["sudo", "-n"]` (non-root). `_run_tc` does `subprocess.run(prefix + ["tc"] + args, ...)` — pure list, no shell. `args` is constructed from hard-coded string literals (`"qdisc"`, `"add"`, `"dev"`, `"lo"`, `"root"`, `"netem"`, `"delay"`, `"200ms"`, `"20ms"`, `"loss"`, `"1%"`) — no env-driven or fixture-derived strings. `sudo -n` ensures non-interactive (no password prompt hang). Clean. The skip-on-no-sudo / skip-on-no-tc paths are correctly ordered before any `tc` call.
- **`tc qdisc del dev lo root` on cleanup runs in a `finally:` block — good.** But if the test process is `SIGKILL`'d mid-run (e.g. workflow `timeout-minutes` expires) the netem qdisc persists on the runner's `lo` until the next install attempt's "best-effort cleanup first." On a self-hosted runner pool this leaks 200 ms RTT + 1% loss into subsequent jobs. Ephemeral GH-hosted runners are fine; document the self-hosted caveat.
- **S6 `ErrShim` binds `127.0.0.1`, listens 256, accepts and serves with bounded buffers.** Per-connection `buf = bytearray()` grows unbounded if a peer never sends a complete RESP frame — under a non-cooperative client this is a DoS on the test harness itself. memtier always sends complete frames so it's fine in practice; harden by capping `buf` length to 1 MiB and dropping the connection on overflow.
- **`test_retry_storm_bounded` reuses `_BIG_BLOB_MB` and `_FILLER_LINES` style env knobs (`MEMTIER_SOAK_TEST_TIME`, etc.) with no upper bound.** Same pattern as #418 — bound them defensively (`min(int(env), 3600)` for `TEST_TIME`).
- **S2 `test_large_payloads`: `f.write("B" * big_value_bytes)` materializes the 100 MiB Python string before writing.** Comment says "stream-write so we don't hold both blobs in Python memory at once" — that's true for *both at once*, but each individual blob is held whole. At 100 MiB default that doubles peak RSS of the test python compared to chunked writing. Cosmetic.
- **Workflow uses `actions/checkout@v4`, `actions/download-artifact@v4`, `actions/setup-python@v5`, `actions/upload-artifact@v4` — all pinned by major tag, not SHA.** Standard for this repo (the existing workflows do the same), but for a nightly that has `NET_ADMIN` via `sudo` and pulls from `packages.redis.io`, SHA-pinning gives more auditability. Not a blocker; org-wide call.

Verdict: ship. `tc`+`sudo` invocation is correctly bounded; no command-injection surface in the netem path.

fcostaoliveira (commented):

Tip: a68d61b. Workflow YAML + 7 soak tests under tests/soak/. Bigger surface; focused on the most load-bearing latent bugs.

- `test_cluster_turbulence.py::_trigger_reshard` issues `CLUSTER SETSLOT <slot> MIGRATING <dest>` against `conns[0]` and never issues the matching `CLUSTER SETSLOT ... NODE` to finalize the migration. The slot is left in MIGRATING state for the remainder of the run AND across subsequent test runs in the same fixture. RLTest may reuse the cluster fixture between tests — the next test that hits that hash slot will get ASK redirects on a slot that has no key migration backing it, which on memtier master triggers an infinite ASK loop until `--max-retries` (or hang if not set). Either complete the migration (`SETSLOT ... NODE`), revert with `SETSLOT ... STABLE` in a `finally`, or run this test in its own isolated fixture.

- `test_cluster_turbulence.py::test_cluster_topology_turbulence` uses `subprocess.Popen(stdin=None, ...)` and *waits* via `proc.communicate()`. `communicate()` has no timeout. If memtier hangs (which is exactly what a topology-refresh deadlock would produce — the bug class this test is supposed to catch), the test never fails; it just consumes the matrix cell's 15 min `timeout-minutes` and reports as cancelled. The whole point of S5 is to *catch* hangs; add `proc.communicate(timeout=test_time + 60)` and on `TimeoutExpired`, kill + `env.assertTrue(False, "memtier hung")`. Currently the suite cannot detect the failure mode it's documented to detect.

- Same pattern repeats in `test_connection_churn.py:91` and `test_high_concurrency.py:103` (`proc.communicate()` with no timeout). S3 (high concurrency / FD hygiene) is also a hang-class detector and silently absorbs hangs the same way. `test_retry_storm.py` *does* set a timeout (`test_time + 120`) — copy that pattern to S3/S4/S5.

- `test_long_run_memory.py::_linear_slope_per_minute` returns 0 when `den == 0` (all samples at the same `t`). But samples are tuples `(time.time() - start_ts, rss)` — *t* is monotonically increasing, so `den` can only be zero if `n < 2`. The slope guard `if len(samples) >= 2:` already covers that. But: the slope is checked against `< 2.0 MB/min`; what about a *negative* slope (memory shrinking — pool/cache eviction)? `slope_mb_per_min < 2.0` accepts arbitrarily negative slopes, which is fine in practice but a real leak that grows slowly and then GCs would average to <2 and pass. Consider `abs(slope_mb_per_min) < 2.0` only for the upper bound and a separate "final/initial < 1.5x" — wait, you already have the ratio cap. OK.

- `test_retry_storm.py::ErrShim._resp_message_length` parses bulk-string subelements but does NOT consume the trailing `\r\n` after the bulk length declaration if `bulk_len < 0` (RESP `$-1`). Look at lines ~135: `if bulk_len >= 0: end = pos + bulk_len + 2; pos = end`. If `bulk_len == -1` (null bulk), the `if` is false and `pos` stays at `sub_nl + 2` — which is correct (no body follows null bulk). But memtier never sends `$-1` in a multibulk request, so this is theoretical. More concretely: inline commands (the `else: pos = sub_nl + 2` branch on non-`$` sub-elements) would mis-count for any pipelined memtier command stream that uses inline format. memtier doesn't use inline — confirmed by reading client::send_command — so the shim works for the tested workload. Worth a comment in the shim docstring confirming the supported subset.

- `test_slow_network.py::_install_netem` calls `_run_tc(["qdisc", "del", ...])` as best-effort cleanup, then `_run_tc(["qdisc", "add", ...])`. If the `add` succeeds and the test process is killed (SIGINT from the runner, GitHub job timeout), the `finally` block doesn't run and `lo` is left with a 200 ms / 1 % loss qdisc on the runner, breaking subsequent jobs on the same host. GitHub-hosted runners are ephemeral so this is harmless there; self-hosted runners (which the soak workflow can be wired to in a future PR) will get bricked. Add a workflow-level `post-job` cleanup step, or trap SIGTERM in the test.

- `soak-nightly.yml` uploads `tests/logs/` and `/tmp/RLTest_*` on failure with `retention-days: 7`. `/tmp/RLTest_*` for S2 (large_payloads) will contain the freshly-generated 100 MiB monitor file *and* the in-flight redis dump if the test failed mid-run. That artifact balloons to multi-GB and GitHub's per-artifact 2 GB cap will silently truncate. Either exclude `/tmp/RLTest_*/big.txt` and `*.rdb`, or do an explicit `rm -f /tmp/RLTest_*/big.txt` before the upload step.

Verdict: Comment

fcostaoliveira (commented):

Verdict: REQUEST CHANGES — ulimit move is correct for cells that reach the run step; tc-netem and chaos-thread cleanup paths have real bugs.

- **ulimit-in-run-step is safe against pre-run-step failures.** Steps preceding "Run scenario" (checkout, download-artifact, chmod, apt-get, setup-python, pip, install redis, tune kernel) can fail; if any does, the cell aborts with `Upload logs on failure` and the run step never executes — no `ulimit -n 65536` needed. The comment at the "Tune kernel" step correctly explains the scoping. **However**: the `Tune kernel` step itself sets sysctls but doesn't set `ulimit -n` there — and the comment says "ulimit is re-set inside the run-scenario step below" which is misleading because it was never set in the kernel step in the first place. Rephrase to "ulimit is set inside the run-scenario step (sysctls persist; per-shell ulimit does not)".
- **`tc netem` cleanup is NOT exception-safe across test-runner crashes.** `_install_netem` adds `qdisc add dev lo root netem ...`. The `finally: _remove_netem()` runs if `test_slow_network_smooth` raises — but if RLTest itself segfaults, or the GitHub Actions job hits the `timeout-minutes` cap, the qdisc stays installed on `lo` for the lifetime of the runner. Hosted runners reset, but **self-hosted soak runners would carry 200ms delay + 1% loss into the next job**. Add a separate `Post-job: ensure tc netem removed` step with `if: always()` that runs `sudo tc qdisc del dev lo root 2>/dev/null || true`.
- **S5 cluster_turbulence chaos thread join is fire-and-forget.** `chaos_thread.join(timeout=5)` followed by an unconditional return — but the chaos function calls `_run_tc(..., timeout=10)` internally via redis-cli. A failover call started at t=T-1 can block for up to 10 s; join(5) returns with thread still alive, the daemon thread is reaped at process exit. **Real edge**: the failover call might land on a redis-server already being torn down by RLTest's cleanup → spurious "DEBUG SLEEP" or "CLUSTER FAILOVER" failures hit stderr after the test logged PASS. Drop the timeout on join (let it complete fully) and add a `stop_chaos.set()` *before* invoking `proc.communicate()`'s timeout path; today `stop_chaos.set()` only happens in the `finally`.
- **Module-level `int(os.environ.get(...))` in S2 (`tests/soak/test_large_payloads.py`) crashes the file at collection if the env is malformed.** Same hazard as #418: `_BIG_BLOB_MB = int(os.environ.get("MEMTIER_SOAK_BIG_BLOB_MB", "100"))` etc. run during import. The other scenarios correctly evaluate inside the test function (verified S1/S4/S5/S6/S7). S2 should follow that pattern.
- **`_BIG_BLOB_MB` defaults to 100 (MB), so a single SET line is ~100 MiB.** Streamed-write is OK, but the file itself sits on disk for the test duration; with the matrix running concurrent self-hosted jobs that adds up. Add a `tearDown`/`finally` `shutil.rmtree(test_dir)` — today `tempfile.mkdtemp()` results survive each scenario and accumulate.
- **Soak workflow has `needs: build` but does NOT pass build status to individual matrix cells via `if:` — a build artifact that uploads with the wrong arch (e.g. failed download mid-upload, corrupt tar) would let every cell start and then fail at "Restore executable permissions" or first invocation.** Add `actions/download-artifact@v4` with `if-no-files-found: error` and a `Smoke-test downloaded binary` step that runs `./memtier_benchmark --version` BEFORE the rest of the setup — catches the corruption cheaply.

(Files: .github/workflows/soak-nightly.yml, tests/soak/test_slow_network.py, tests/soak/test_cluster_turbulence.py, tests/soak/test_large_payloads.py.)

PR #424, opened 2026-05-30
test(cli): hypothesis-based CLI argument fuzzer

### LINKED ISSUE #410: Add hypothesis-based CLI fuzz tests for memtier_benchmark flags
### Motivation

PR #405 fixed a stack-buffer / VLA crash driven by attacker-controlled input. The fix was good, but it exposed a structural gap: **every CLI flag is an input surface**, and a handful remain attacker-controlled in real deployments:

- CI templates and shared benchmarking harnesses (codeperf, redis-benchmarks-specification) interpolate user-supplied values into `--command "..."`, `--key-prefix`, `--data-import`, `--monitor-pattern`.
- Helm charts in shared k8s clusters expose `--server`, `--port`, `--protocol`, `--ratio` as ConfigMap-driven strings.
- Multi-tenant benchmark services accept `--data-size-list`, `--key-zipf-exp`, `--clients-step` from REST APIs.

Today **none** of these surfaces are fuzzed. `tests/` has good happy-path coverage (zipfian, scan, monitor, retry-on-error) but the option parser itself is exercised only with valid inputs. A grep across `tests/*.py` for ~80 long options shows ~30 of them have *zero* negative-input coverage.

### Proposal

Add `tests/test_cli_fuzz.py` using [hypothesis](https://hypothesis.readthedocs.io/) to generate random combinations of 1–8 long flags with deliberately weird values (negatives, `INT_MAX`/`LLONG_MAX`, empty strings, 1 MiB strings, unicode, `nan`/`inf`, malformed ratios, embedded NULs, format-string specifiers). For each generated argv:

1. Run `memtier_benchmark --test-time=1 --threads=1 --clients=1 --hide-histogram <flags>` against a local redis.
2. Assert `returncode in {0, 1, 2}` (clean exit or parser error).
3. Assert stderr contains **none** of: `stack smashing`, `Segmentation fault`, `Aborted`, `AddressSanitizer`, `UndefinedBehaviorSanitizer`, `runtime error:`, `munmap_chunk`, `double free`, `SIGSEGV`, `SIGABRT`.
4. Assert the process exits within 10 s (no hangs).

Caps examples at 200, gated by env var so it can run nightly under ASAN/UBSan builds where any banned-needle hit becomes a fast regression signal.

### Strategies (the weird stuff)

`weird_int`: `-1`, `-2147483648`, `INT_MAX`, `INT_MAX+1`, `LLONG_MAX`, `LLONG_MAX+1`, hex literals (`0x7fffffff`), scientific notation (`1e9`, `1e9999`), empty / single-space, `1.5`, `1,5`.
`weird_float`: `nan`, `inf`, `-inf`, `1e308`, `1e-308`, `0`, comma decimal.
`weird_str`: empty, 1 MiB `"a"*2^20`, embedded NUL, format-string specifiers (`%n%n%s`), ANSI escapes, shell-meta (`$(reboot)`), path traversal, `"🔥" * 256`.
`ratio_like`: `1:0`, `0:1`, `1:1:1`, `-1:1`, `1:`, `:1`, `a:b`.
`key_pattern`: valid (R/S/G/P/Z), invalid (X, empty, missing colon).

### Uncovered flag combinations (seed corpus)

15 combinations currently absent from `tests/`:

1. `--data-size-range=1-9999999999` (huge upper bound; `tests_oss_simple_flow.py` only tests fixed `--data-size`)
2. `--data-size-list=8:50,16:50` (weighted; no test references `--data-size-list` at all)
3. `--data-size-pattern=S` with `--data-size-range`
4. `--data-offset=N` with `--key-prefix`
5. `--expiry-range=1-10` together with `--no-expiry` (mutually contradictory)
6. `--key-stddev` + `--key-median` with `--key-pattern=G:G` (gaussian path uncovered)
7. `--multi-key-get=N` with `--cluster-mode` (legacy flag, not exercised)
8. `--wait-ratio --wait-timeout --num-slaves` together (WAIT injection)
9. `--select-db=N` non-default DB
10. `--distinct-client-seed` + `--randomize`
11. Reversed `--thread-conn-start-min-jitter-micros` / `--max-jitter-micros` (min > max)
12. Edge values for `--clients-start` / `--clients-step` / `--step-duration` (0, INT_MAX)
13. `--scan-incremental-max-iterations=0`
14. Malformed `--print-percentiles` (`50.50.50`, trailing comma)
15. Negative `--command-ratio`, `--retry-on=GARBAGE`, `--protocol=resp3`, empty `--statsd-host`, negative `--graphite-port`, `--connection-timeout=0`

### Out of scope (for now)

`--tls*`, `--uri`, `--data-import`, `--failed-keys-file` — need setup harness; follow-up.

### Acceptance

- `pytest tests/test_cli_fuzz.py --hypothesis-seed=0` passes against current `master` under both gcc and clang+ASAN.
- Any shrunk failing example is filed as a separate issue.
- Test is wired into `run_tests.sh` behind a `MEMTIER_FUZZ=1` env gate so default CI time is unaffected.

Related: #404, #405.

fcostaoliveira:
Fixes #410.

## Summary

Adds `tests/test_cli_fuzz.py`: a Hypothesis-driven fuzzer that generates 1-8 long-flag combinations with deliberately weird values (empty/whitespace, INT_MAX/LLONG_MAX boundaries, scientific notation, comma decimal, hex, 4 KiB ASCII, 1 KiB UTF-8 fire emoji, format-string specifiers, malformed ratios / ranges / percentiles / weight lists) and runs each against a local Redis with the canonical fast-exit harness:

```
memtier_benchmark --server=... --port=... --test-time=1 --threads=1 --clients=1 --hide-histogram <generated flags>
```

Each generated argv must:
- exit within 10 s (no hangs);
- return 0, 1 or 2 (parser error is fine, signal-class is not);
- not emit any crash-class needle in stderr (`stack smashing`, `Segmentation fault`, `Aborted`, `AddressSanitizer`, `UndefinedBehaviorSanitizer`, `runtime error:`, `munmap_chunk`, `double free`, `SIGSEGV`, `SIGABRT`).

The fuzzer is gated behind `MEMTIER_FUZZ=1` so default CI is unaffected; intended to run nightly under ASAN/UBSan builds where any banned-needle hit becomes a fast regression signal. It is also wired into `tests/run_tests.sh` behind the same env so a single `MEMTIER_FUZZ=1 ./tests/run_tests.sh` invocation picks it up after the RLTest suites.

Out of scope per the issue: `--tls*`, `--uri`, `--data-import`, `--failed-keys-file`, `--monitor-input` (need dedicated setup harnesses; queued as follow-ups).

## Test plan

- [x] `make -j` builds against current master.
- [x] `pip install -r tests/test_requirements.txt` resolves Hypothesis + pytest.
- [x] Default `pytest tests/test_cli_fuzz.py` is skipped (no `MEMTIER_FUZZ`).
- [x] `MEMTIER_FUZZ=1 MEMTIER_BINARY=$PWD/memtier_benchmark pytest tests/test_cli_fuzz.py --hypothesis-seed=0 -x -v` passes against a local Redis on `:6379` (200 examples, ~2 min).
- [x] Same with seeds `1, 2, 3, 42, 99, 100, 200, 999, 12345` (each at 100 examples).
- [x] `MEMTIER_FUZZ_MAX_EXAMPLES=400 ... --hypothesis-seed=0` also passes (~3.5 min).
- [ ] Nightly: run under ASAN/UBSan builds (separate workflow change).

## Follow-up bugs surfaced while iterating

While narrowing the strategies to keep the test green on master, the fuzzer found a batch of pre-existing crash/hang classes. Per the prompt, none of these are fixed in this PR; each becomes a follow-up issue, and the corresponding strategy was trimmed so the test is a stable regression net rather than an immediate bug-of-the-day reporter.

| # | argv (minimal repro) | symptom |
|---|----------------------|---------|
| 1 | `--authenticate ''` against a server with no auth configured | hangs (reconnect loop, ignores `--test-time`) |
| 2 | `--cluster-mode` against standalone Redis | hangs (`cluster slot failed.` loop, ignores `--test-time`) |
| 3 | `--protocol memcache_text` (or `memcache_binary`) against Redis | hangs (`response parsing failed.` loop) |
| 4 | `--run-count -1` | SIGABRT, `std::bad_alloc` (negative passed to vector resize) |
| 5 | `--run-count 2147483647` (INT_MAX) | SIGABRT, `std::bad_alloc` (allocator OOM, no bounds check) |
| 6 | `--pipeline -1` | hangs (negative size, underflow loop) |
| 7 | `--data-size -1` | SIGABRT, `protocol.cpp:309: assert value_len > 0 failed` |
| 8 | `--data-size-range 1-9999999999` | hangs after `-ERR Protocol error: invalid bulk length` (memtier reconnects forever) |
| 9 | `--data-size-list 0:50` | SIGABRT, value_len=0 assert |
| 10 | `--data-size-list 8:0` | hangs (zero weight, sampler picks nothing) |
| 11 | `--select-db 100` against default 16-DB Redis | hangs (`database selection failed.` loop) |
| 12 | `--command __key__` (bare placeholder, no command name) | SIGABRT, `protocol.cpp:774: first arg is not command name?` |
| 13 | `--command ''` | hangs |
| 14 | `--command-ratio 0` (or empty / whitespace, both parse as 0) | hangs |
| 15 | `--key-pattern G:G --key-stddev inf` | hangs (sampler never produces a key) |
| 16 | `--key-pattern G:G --key-maximum 1` | SIGABRT (Gaussian path on a 1-key range) |
| 17 | `--wait-ratio 0:1 --num-slaves 1-10` against standalone | hangs (WAIT never satisfied, no timeout exit) |

Common pattern in (1), (2), (3), (8), (11), (17): once memtier enters its reconnect / retry loop after a server-side failure during connection setup, it does **not** honor `--test-time` and runs until SIGKILLed. That's arguably the most actionable single fix in this batch -- one bounded "give up after N seconds of connect-stage failures" would close most of them.

Each of these will be filed as a separate issue and cross-referenced back to #410.

---

> [!NOTE]
> **Low Risk**
> Test-only changes behind MEMTIER_FUZZ=1; no production code paths modified.
> 
> **Overview**
> Adds an opt-in **Hypothesis CLI fuzzer** (`tests/cli_fuzz.py`, issue #410) that runs `memtier_benchmark` against local Redis with 1–8 randomly combined long flags and curated “weird” values, on top of a short fixed harness (`--test-time=1`, single thread/client). Each case must finish within **10s**, exit with **0/1/2**, and avoid crash/sanitizer needles in stderr.
> 
> The fuzzer is **off by default** (`MEMTIER_FUZZ=1`); `tests/run_tests.sh` documents `MEMTIER_FUZZ` / `MEMTIER_FUZZ_SEED` and can run pytest after RLTest. **`tests/test_requirements.txt`** pins `pytest` and `hypothesis`. Strategies deliberately exclude known hang/crash inputs (auth, cluster, memcache protocols, extreme ints, etc.) so the suite stays a stable parser/regression net; product fixes are left as follow-ups.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 81e0beb46295b747efaee172f4b32f58eea8c75a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/cli_fuzz.py
+++ b/tests/cli_fuzz.py
@@ -19,7 +19,7 @@ Out of scope (per issue #410): --tls*, --uri, --data-import,
 
 Usage:
     MEMTIER_FUZZ=1 MEMTIER_BINARY=$PWD/memtier_benchmark \\
-        pytest tests/test_cli_fuzz.py --hypothesis-seed=0 -x -v
+        pytest tests/cli_fuzz.py --hypothesis-seed=0 -x -v
 """
 
 import os
--- a/tests/run_tests.sh
+++ b/tests/run_tests.sh
@@ -31,7 +31,7 @@ help() {
 		RLTEST_VERBOSE=1    Enable RLTest verbose mode
 		RLTEST_DEBUG=1      Enable RLTest debug print
 		MEMTIER_FUZZ=1      Run the pytest-driven Hypothesis CLI fuzzer after the
-		                    RLTest suites (see tests/test_cli_fuzz.py, issue #410).
+		                    RLTest suites (see tests/cli_fuzz.py, issue #410).
 		MEMTIER_FUZZ_SEED=n Hypothesis seed (default: 0).
 
 	END
@@ -107,14 +107,14 @@ E=0
 	((E |= $?))
 } || true
 
-# Optional pytest-driven Hypothesis CLI fuzzer (see tests/test_cli_fuzz.py
+# Optional pytest-driven Hypothesis CLI fuzzer (see tests/cli_fuzz.py
 # and issue #410). Off by default so the standard suite stays fast; opt in
 # with MEMTIER_FUZZ=1, which is also the gate that the test itself checks.
 # Intended to run nightly under ASAN/UBSan builds.
 [[ $MEMTIER_FUZZ == 1 ]] && {
 	printf "\nRunning CLI hypothesis fuzzer (MEMTIER_FUZZ=1):\n\n"
 	(cd $ROOT/tests && MEMTIER_BINARY=$MEMTIER_BINARY \
-		python3 -m pytest -p no:asyncio test_cli_fuzz.py \
+		python3 -m pytest -p no:asyncio cli_fuzz.py \
 		--hypothesis-seed=${MEMTIER_FUZZ_SEED:-0} -v)
 	((E |= $?))
 } || true
--- a/tests/run_tests.sh
+++ b/tests/run_tests.sh
@@ -30,6 +30,9 @@ help() {
 		                    cells under the CI step timeout.
 		RLTEST_VERBOSE=1    Enable RLTest verbose mode
 		RLTEST_DEBUG=1      Enable RLTest debug print
+		MEMTIER_FUZZ=1      Run the pytest-driven Hypothesis CLI fuzzer after the
+		                    RLTest suites (see tests/test_cli_fuzz.py, issue #410).
+		MEMTIER_FUZZ_SEED=n Hypothesis seed (default: 0).
 
 	END
 }
@@ -101,6 +104,18 @@ E=0
 
 [[ $OSS_CLUSTER == 1 ]] && {
 	(ROOT_FOLDER=$ROOT TLS_KEY=$TLS_KEY TLS_CERT=$TLS_CERT TLS_CACERT=$TLS_CACERT MEMTIER_BINARY=$MEMTIER_BINARY RLTEST_ARGS="${RLTEST_ARGS} --env oss-cluster --shards-count $SHARDS" run_tests "tests on OSS cluster")
+	((E |= $?))
+} || true
+
+# Optional pytest-driven Hypothesis CLI fuzzer (see tests/test_cli_fuzz.py
+# and issue #410). Off by default so the standard suite stays fast; opt in
+# with MEMTIER_FUZZ=1, which is also the gate that the test itself checks.
+# Intended to run nightly under ASAN/UBSan builds.
+[[ $MEMTIER_FUZZ == 1 ]] && {
+	printf "\nRunning CLI hypothesis fuzzer (MEMTIER_FUZZ=1):\n\n"
+	(cd $ROOT/tests && MEMTIER_BINARY=$MEMTIER_BINARY \
+		python3 -m pytest -p no:asyncio test_cli_fuzz.py \
+		--hypothesis-seed=${MEMTIER_FUZZ_SEED:-0} -v)
 	((E |= $?))
 } || true
 
--- a/tests/test_cli_fuzz.py
+++ b/tests/test_cli_fuzz.py
@@ -1,0 +1,437 @@
+"""
+Hypothesis-based CLI argument fuzzer for memtier_benchmark.
+
+See https://github.com/redis/memtier_benchmark/issues/410 for motivation.
+
+Goal: every CLI flag is an input surface (CI templates, Helm charts, REST APIs
+interpolate user-supplied values into flags). Existing tests exercise happy
+paths only; this test feeds intentionally weird values into a curated set of
+long flags and asserts that memtier_benchmark either runs cleanly or exits
+with a parser-error return code -- never crashes, never hangs, never trips a
+sanitizer.
+
+Gated behind MEMTIER_FUZZ=1 so default CI is unaffected. Intended to run
+nightly under ASAN/UBSan builds where any banned-needle hit becomes a fast
+regression signal.
+
+Out of scope (per issue #410): --tls*, --uri, --data-import,
+--failed-keys-file, --monitor-input (need setup harness; follow-up).
+
+Usage:
+    MEMTIER_FUZZ=1 MEMTIER_BINARY=$PWD/memtier_benchmark \\
+        pytest tests/test_cli_fuzz.py --hypothesis-seed=0 -x -v
+"""
+
+import os
+import shutil
+import subprocess
+
+import pytest
+
+hypothesis = pytest.importorskip("hypothesis")
+from hypothesis import HealthCheck, given, settings, strategies as st  # noqa: E402
+
+FUZZ_ENABLED = os.environ.get("MEMTIER_FUZZ") == "1"
+pytestmark = pytest.mark.skipif(
+    not FUZZ_ENABLED, reason="CLI fuzzer disabled (set MEMTIER_FUZZ=1 to enable)"
+)
+
+MEMTIER = os.environ.get("MEMTIER_BINARY") or shutil.which("memtier_benchmark")
+HOST = os.environ.get("MEMTIER_FUZZ_HOST", "127.0.0.1")
+PORT = os.environ.get("MEMTIER_FUZZ_PORT", "6379")
+
+# Stderr substrings that indicate a real crash or sanitizer hit. Hitting any
+# of these is an automatic fail -- the parser is allowed to reject input, but
+# never to crash on it.
+BANNED_NEEDLES = (
+    "stack smashing",
+    "Segmentation fault",
+    "Aborted",
+    "AddressSanitizer",
+    "UndefinedBehaviorSanitizer",
+    "runtime error:",
+    "munmap_chunk",
+    "double free",
+    "SIGSEGV",
+    "SIGABRT",
+)
+
+# Acceptable exit codes: 0 success, 1/2 parser/runtime error.
+OK_RETURN_CODES = {0, 1, 2}
+
+# Per-issue strategies. We deliberately mix a curated corpus of known-weird
+# tokens (negatives, INT_MAX, scientific notation, comma decimal, format
+# specifiers, etc.) with a small hypothesis-generated random branch so the
+# engine can shrink to a minimal failing example if anything trips.
+# Weird-int corpus geared at the **parser**, not the allocator. Several
+# count-style flags (--run-count, --pipeline, --data-size, ...) are piped
+# straight into vector resizes / mallocs without bounds-checking, so feeding
+# them INT_MAX / LLONG_MAX or a negative value reliably aborts the process
+# with std::bad_alloc or an assertion. That's a known bug class -- tracked
+# as a follow-up for the same #410 series -- and orthogonal to what we're
+# trying to catch here, which is *parser* sanity (empty string, leading
+# space, fractional, comma decimal, hex, scientific notation, plain
+# overflow-as-string). We therefore keep all generated values small enough
+# that they cannot themselves cause an OOM regardless of which flag they
+# land on, while still exercising every interesting *string* shape.
+weird_int = st.one_of(
+    st.sampled_from(
+        [
+            "0",
+            "1",
+            "2",
+            "10",
+            "100",
+            "0x7f",
+            "1e3",
+            "1e9999",  # parser-level overflow, *not* allocator
+            "",
+            " ",
+            "1.5",
+            "1,5",
+        ]
+    ),
+    st.integers(min_value=0, max_value=10000).map(str),
+)
+
+weird_float = st.sampled_from(
+    [
+        "nan",
+        "inf",
+        "-inf",
+        "1e308",
+        "1e-308",
+        "0",
+        "0.0",
+        "1,5",
+        "-1.0",
+        "1.5",
+        "",
+    ]
+)
+
+# Note: embedded NULs are filtered out -- POSIX execve() rejects them with
+# E2BIG/EINVAL before memtier ever sees the argv, and that confounds the
+# subprocess layer (Python raises ValueError) rather than exercising the
+# parser. The parser-level fuzz is still very much in scope via everything
+# else here.
+weird_str = st.one_of(
+    st.sampled_from(
+        [
+            "",
+            " ",
+            "a",
+            "%n%n%s",
+            "%s%s%s%s",
+            "\x1b[31mANSI\x1b[0m",
+            "$(reboot)",
+            "`reboot`",
+            "../" * 16 + "etc/passwd",
+            "\U0001f525" * 256,
+            "a" * 4096,
+        ]
+    ),
+    st.text(
+        alphabet=st.characters(blacklist_categories=("Cs",), blacklist_characters="\x00"),
+        min_size=0,
+        max_size=128,
+    ),
+)
+
+ratio_like = st.sampled_from(
+    [
+        "1:0",
+        "0:1",
+        "1:1",
+        "1:1:1",
+        "-1:1",
+        "1:-1",
+        "1:",
+        ":1",
+        "a:b",
+        "",
+        "1:10",
+    ]
+)
+
+# Gaussian (G:G) is intentionally absent here: with the harness default
+# key range (--key-maximum=10000000 implicit) it's fine, but if hypothesis
+# also picks a narrow --key-maximum or an unhealthy --key-stddev value
+# (inf/nan), the Gaussian sampler either aborts or spins forever instead
+# of failing the parser-error path. Filed as a follow-up.
+key_pattern_like = st.sampled_from(
+    [
+        "R:R",
+        "S:S",
+        "P:P",
+        "Z:Z",
+        "X:X",
+        "",
+        "RR",
+        "R:",
+        ":R",
+        "R:R:R",
+    ]
+)
+
+range_like = st.sampled_from(
+    [
+        "1-10",
+        "10-1",  # reversed
+        "1-1000",  # large but well under Redis bulk-length cap
+        "0-0",
+        "-1-10",
+        "1-",
+        "-10",
+        "",
+        "a-b",
+    ]
+)
+
+# memcache_text / memcache_binary are intentionally absent: against a Redis
+# backend they enter a parse-fail / reconnect loop and never honor
+# --test-time. Tracked as a follow-up bug; fuzzing them properly needs a
+# memcached harness.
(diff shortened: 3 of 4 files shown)


fcostaoliveira (commented):

- **Hypothesis CLI fuzzer (pytest, dev-gated)**: `test_cli_fuzz.py` behind `MEMTIER_FUZZ=1` env flag. Off by default. Generates 1–8 flag combinations + runs memtier 10 s timeout per variant. On default CI, 100% skipped.
- **Nightly execution**: When enabled (nightly ASAN/UBSan builds), estimated ~2–5 min total for hypothesis default example count (~100 exs). Acceptable for parser hardening.
- **No production binary cost**: Test harness only.

Verdict: **Approve** — opt-in fuzzer, default CI unaffected.

fcostaoliveira (commented):

• Title uses Conventional Commits correctly: `test(cli): subject`.
• Opens with `Fixes #410.` — correct (issue dedicated to CLI fuzzer).
• Test plan comprehensive: default skip path (no env var), 200 examples pass locally, multiple seeds (1,2,3,42,99,100,200,999,12345) verified, Hypothesis determinism confirmed.
• New env vars `MEMTIER_FUZZ=1` and `MEMTIER_FUZZ_MAX_EXAMPLES=` (internal) introduced. Documented as "off by default" and wired into `tests/run_tests.sh` behind env var, but AGENTS.md not updated. No explicit mention of Hypothesis dependency in test_requirements.txt mention.
• "Follow-up bugs surfaced" section notes several pre-existing crashes/hangs trimmed to keep test green — appropriate deferral, but note: the table is cut off (incomplete).

Verdict: Comment — missing AGENTS.md documentation for MEMTIER_FUZZ env var; table in "Follow-up bugs" section incomplete.

fcostaoliveira (commented):

- **Dropping INT_MAX / LLONG_MAX from the `weird_int` strategy (rather than xfail-ing them) is the right call for a stable regression net, but it hides a real bug class.** The PR describes 17 known crashes/hangs found while building the strategy; for bugs like `--run-count 2147483647 → std::bad_alloc` (item #5), the correct long-term design is `pytest.mark.xfail(strict=True)` with a reference to the tracking issue. Silently excluding INT_MAX means future fixes to those bugs will not automatically be detected by this test when they land.
- **The `command_like` strategy excluding empty / bare placeholder is correct scoping.** `--command ''` hangs, `--command __key__` aborts — both are real bugs, but both require separate fix PRs. The strategy comment documents exactly why these are excluded and references the follow-up issues, which is the right paper trail.
- **`MEMTIER_FUZZ=1` gate + `pytest.mark.skipif` at module level is the correct opt-in design.** Avoids polluting the default `run_tests.sh` invocation; the test still participates in `run_tests.sh` when explicitly opted in via `MEMTIER_FUZZ=1`.
- **Tolerances read at import time (module-level `os.environ.get`) means pytest fixture injection cannot override them.** This is minor but means `MEMTIER_FUZZ_PORT` set by a `conftest.py` fixture after import would be ignored. Reading inside the test function or using a session-scoped fixture would be more robust.
- **The 17-item follow-up bug table in the PR description is valuable archaeology.** Consider adding a comment block in `test_cli_fuzz.py` listing the excluded input shapes with their tracking issue numbers — reviewers checking whether a known-crash was fixed will look in the test file, not in a closed PR description.

Verdict: Comment (use xfail for known crashes instead of strategy exclusion, so fixes are auto-detected)

fcostaoliveira (commented):

- Hypothesis is in-process and the harness runs memtier via `subprocess.run(cmd, ..., timeout=10, check=False)` with argv as a list — no shell expansion of mutated bytes, no command injection from generated values. The 10 s subprocess timeout hard-bounds runtime per example.
- `MEMTIER_BINARY` is env-resolved with `shutil.which()` fallback. PATH-takeover by a malicious binary on `$PATH` is the classic risk; for an opt-in fuzzer gated by `MEMTIER_FUZZ=1` this is acceptable. Same caveat as PR #420 — consider preferring absolute paths in nightly CI.
- `weird_str` strategy correctly **filters out NUL** (`blacklist_characters="\x00"`) and surrogate codepoints (`Cs`) — necessary because POSIX `execve` rejects argv with embedded NULs and Python would raise `ValueError` at the subprocess layer rather than exercising memtier's parser. Good.
- `weird_str` includes `"$(reboot)"`, `` "`reboot`" ``, and `"../" * 16 + "etc/passwd"` as VALUES, not as commands. Because argv is a list, these are passed literally as flag arguments — not executed. The path-traversal-looking string only reaches the parser as a `--key-prefix=...` style value; memtier doesn't open user-provided paths from any of the fuzzed flags, so no file system access. Confirmed by the PR's exclusion of `--data-import`, `--failed-keys-file`, `--monitor-input` (which would be path-injecting).
- ANSI escapes (`\x1b[31m`) in `weird_str` reach `stderr`/stdout but `subprocess.run` captures bytes — no terminal-injection risk back into pytest output.
- `BANNED_NEEDLES` includes the canonical crash signatures. The PR honestly catalogs ~12 pre-existing crash/hang classes (negative `--run-count`, INT_MAX `--run-count`, zero `--data-size`, etc.) and **trims the strategy** rather than fixing the bugs. That's the right call for a regression-net PR, but it means today's coverage is meaningfully smaller than the issue intends — a future contributor must be careful when re-broadening strategies without first landing the underlying fixes.
- The fuzzer connects to `127.0.0.1:6379` by default; `MEMTIER_FUZZ_HOST` overrides allow pointing at remote redis, which would let a malicious env weaponize the fuzzer to flood an external service. Local-only is the documented intent.
- `pytest.importorskip("hypothesis")` correctly degrades when the dep isn't installed.
- `--hypothesis-seed=${MEMTIER_FUZZ_SEED:-0}` in `run_tests.sh` uses parameter expansion in bash, not user-supplied — safe.
- `(cd $ROOT/tests && ...)` lacks `set -e` semantics propagation, but the `((E |= $?))` pattern picks up the exit. Sound.

Verdict: Approve

fcostaoliveira (commented):

- **`HealthCheck` is imported but only one member is used in the settings decorator.** `HealthCheck.too_slow` and `HealthCheck.function_scoped_fixture` are both referenced, so the import is not unused. However, importing `settings` from hypothesis and using it only in the decorator is fine; no issue here.
- **Comment quality for excluded strategies is excellent.** Each `# Known bug class intentionally excluded` block names the exact argv, the symptom, and the follow-up reference. This is the right pattern for documenting deliberate test holes in a fuzz harness.
- **`weird_int` and `weird_str` strategy names are descriptive.** The inline comment on `weird_int` explaining why it caps at 10000 (to avoid allocator OOM as a confound for parser bugs) is load-bearing and accurate.
- **`BANNED_NEEDLES` tuple name is clear;** the list matches the project's standard crash-detection vocabulary from other test files.
- **`_run_memtier` helper name is clear.** Minor: it is a module-level function that captures `HOST` and `PORT` from globals — the dependency on those globals is implicit. A docstring noting this would help.
- **`run_tests.sh` additions** follow the same opt-in pattern as STRESS; the help text is accurate and consistent in style.

Verdict: Approve

fcostaoliveira (commented):

- `tests/test_cli_fuzz.py:114-132` — `weird_int` includes empty/space/comma-decimal/`1e9999` but also `st.integers(min_value=0, max_value=10000).map(str)`. The hypothesis-generated branch never hits negatives or `INT_MAX-1`; the curated branch skips them too. Whole classes of off-by-one are not exercised. The author flags this in the comment but `--rate-limiting=0`, `--clients-start=0`, `--pipeline=0`, and `--max-retries=0` ALL feed straight into divisions / modulo and SIGFPE is a real outcome — currently masked.
- `tests/test_cli_fuzz.py:420-437` — `_run_memtier` sets `timeout=10` for a `--test-time=1` workload. With pipeline=very-large + clients=1, memtier can take >10s to drain the pipeline AFTER test-time ends, especially under sanitizer builds. `pytest.fail(...hung...)` fires on what is actually a clean (slow) shutdown.
- `tests/test_cli_fuzz.py:96-97` — `OK_RETURN_CODES = {0, 1, 2}`. memtier's parser uses `exit(EXIT_FAILURE)` (=1) for parse errors but some validation paths call `exit(EXIT_USAGE_FAILURE)` which is 64 on BSD-derived hosts (compat module returns 1 on Linux). On macOS / Free-BSD-runner future you get spurious fails. Pin to known set explicitly or use `< 64`.
- `tests/test_cli_fuzz.py:399-417` — `argv` strategy draws 1-8 unique flag indices; flags with `strategy is None` (booleans) take no value. But the fuzzer never combines mutually-exclusive booleans (`--ipv4` and `--ipv6`) or boolean+value flag with the same effect. Not catastrophic, just narrows coverage.
- `tests/test_cli_fuzz.py:75` — `MEMTIER = os.environ.get("MEMTIER_BINARY") or shutil.which("memtier_benchmark")`. If neither resolves to the in-tree binary (e.g. CI env unset), `which` finds an installed package binary — possibly a DIFFERENT VERSION — and the fuzzer reports false positives/negatives against it. Should fail-fast when env-var missing.
- `tests/test_cli_fuzz.py:439-446` — `max_examples=int(os.environ.get("MEMTIER_FUZZ_MAX_EXAMPLES", "200"))` with `deadline=None`. 200 examples × 10s timeout = up to 33 min nominal; under sanitizers 1-2hr easily. The wrapper in `run_tests.sh` doesn't set its own timeout, and the workflow inheriting `timeout-minutes` may not exist if MEMTIER_FUZZ=1 is set in a non-soak workflow.
- `tests/test_cli_fuzz.py:155-176` — `weird_str` blacklists `\x00` only via `blacklist_characters="\x00"` for the hypothesis-generated branch. The `st.sampled_from(...)` curated branch contains `"\x1b[31mANSI\x1b[0m"` (OK) but not NUL itself (good). However `"$(reboot)"` and `` "`reboot`" `` are passed unquoted via `subprocess.run([list])` — Python's subprocess doesn't shell-evaluate them, so they don't actually test command-substitution. Comment is misleading.
- `tests/run_tests.sh:117-122` — `MEMTIER_FUZZ=1` block runs unconditionally even when the previous RLTest E was non-zero. `((E |= $?))` ORs both, but a hung fuzz run dragging the job into its 60-min timeout could mask the original RLTest failure. Consider `[[ $E -eq 0 ]] && [[ $MEMTIER_FUZZ == 1 ]]`.

Verdict: Comment

fcostaoliveira (commented):

- **LOAD-BEARING:** `tests/test_cli_fuzz.py` lives at `tests/test_cli_fuzz.py` and matches RLTest's autodiscovery glob (`test_*.py`). When the standard `./tests/run_tests.sh` runs RLTest (without `MEMTIER_FUZZ`), RLTest will try to discover and invoke `test_cli_fuzz_does_not_crash` with its `env` fixture, but the function is wrapped by Hypothesis's `@given` which expects no fixture argument. RLTest's test loader will either error out at collection (RLTest may not understand pytest's `pytestmark = pytest.mark.skipif(...)`) or call the wrapper with mismatched args. At minimum, verify RLTest gracefully skips this file in a default `OSS_STANDALONE=1 ./tests/run_tests.sh` run; consider renaming to `cli_fuzz_test.py` or moving outside RLTest's discovery root.
- `tests/test_cli_fuzz.py:288-290` `_run_memtier` uses `subprocess.run(..., timeout=10)` — but argv-fuzz can generate `--test-time=1` PLUS `--rate-limiting=0` plus retries that cause memtier to spin-loop on reconnects (known bug, per PR description). The harness depends on the 10s timeout to catch hangs. When it triggers, `subprocess.TimeoutExpired` is raised and `pytest.fail` is called — fine. But the subprocess is NOT killed before `TimeoutExpired` propagates; Python 3.x kills it on context exit, but there's no explicit kill. Under high parallel pytest workers this can pile up zombie memtier processes.
- `tests/test_cli_fuzz.py:264-271` Hypothesis decorators: `@settings(...)` is applied to `@given(extra=argv())`. The shrinker can spend many seconds shrinking a failing argv; with `max_examples=200` and worst-case shrink times of 5-10s each, a single failure can blow past pytest's default timeout. Fine since the deadline is None — just noting CI runtime variability.
- `tests/test_cli_fuzz.py:OK_RETURN_CODES = {0, 1, 2}` — bash subprocess kill due to timeout returns `-9` (SIGKILL), but the timeout branch is handled separately via `TimeoutExpired`. memtier signal exits return `-N` (negative); none of `-N` are in the set. Good.
- `tests/test_cli_fuzz.py:285-291` `_run_memtier` passes argv with the binary as first element. If `MEMTIER` is None (binary not found via `shutil.which` and `MEMTIER_BINARY` unset), `cmd[0]` is None and `subprocess.run` raises `TypeError`. The `pytestmark` skipif handles `MEMTIER_FUZZ`, and the separate `@pytest.mark.skipif(MEMTIER is None, ...)` on the test function handles missing-binary. OK.
- `tests/test_cli_fuzz.py:165` `weird_str` strategy excludes `\x00` but includes all other control characters. `subprocess.run` on POSIX wraps in `execve`; control chars are passed through. Hex like `\x1b` (escape) in an argument is fine for execve, just may produce odd terminal output if memtier echoes the arg in an error message.
- `tests/run_tests.sh:117-122` `MEMTIER_FUZZ=1` block runs pytest in a subshell with `(cd $ROOT/tests && ...)`. The exit code propagates via `((E |= $?))` after the `()` exits — but `$?` inside the if-block refers to the subshell exit, which is the LAST command in the subshell (pytest). OK.
- `--hypothesis-seed=${MEMTIER_FUZZ_SEED:-0}` — Hypothesis seed 0 is fine; deterministic across runs. But fixed seed means a flaky failure won't be discovered by re-running; users need to vary the seed manually.

Verdict: Comment

PR #425, opened 2026-06-01
fix(cluster-slots): validate reply shape before indexing, log+skip malformed shards

### LINKED ISSUE #417: cluster-mode: SIGSEGV on malformed CLUSTER SLOTS reply from server (handle_cluster_slots blind index)

fcostaoliveira:
Fixes #417

`cluster_client::handle_cluster_slots()` walked the `CLUSTER SLOTS` reply with bare `mbulks_elements[N]` indexing and `as_bulk()` / `as_mbulk_size()` casts that `assert(0)` on type mismatch. A malformed reply from a hostile or buggy server crashed memtier with SIGSEGV — captured by the RESP-fuzzer fixture `tests/fixtures/resp_payloads/cluster_slots_malformed.bin` introduced in PR #419 (which is `env.skip()`'d with a `TODO(#417)` until this lands).

## Repro

```
*1\r\n*1\r\n+notanint\r\n
```

i.e. an outer 1-element array containing an inner 1-element array whose only element is a simple-string. Old code did `shard->mbulks_elements[1]` past the inner-array's only element → out-of-bounds read → SEGV.

## Fix

Validate before indexing:
- top-level `get_mbulk_value()` is non-null
- each shard element passes `is_mbulk_size()`
- shard has ≥ 3 children (start, end, [host, port, ...])
- elements 0 and 1 (slot bounds) pass `is_bulk()`
- element 2 (node tuple) passes `is_mbulk_size()` and has ≥ 2 `is_bulk()` children (host, port)

Any malformed shard logs a `benchmark_error_log` and is skipped — the existing bootstrap connection stays in service. If the whole reply is non-array, log and return without touching topology.

## Test plan

- [x] Builds clean (default + ASAN); `make format-check` clean.
- [ ] Once merged: drop the `env.skip()` + `TODO(#417)` from `tests/test_resp_response_fuzzer.py::test_cluster_slots_malformed`. That test now exercises the validation path. (Follow-up PR.)
- [ ] Full CI matrix on this PR — to be confirmed.

---

> [!NOTE]
> **Medium Risk**
> Changes cluster topology parsing and connection lifecycle on refresh; validation reduces crash/OOB risk but alters behavior when servers return bad or empty slot maps.
> 
> **Overview**
> **Hardens `cluster_client::handle_cluster_slots()`** so memtier no longer crashes or corrupts memory on a bad `CLUSTER SLOTS` RESP reply (fixes #417).
> 
> Before walking shards, the handler now requires a non-empty top-level array and validates each shard’s shape (`[start, end, [host, port, …]]`) with `is_*` checks instead of unchecked `as_bulk()` / `as_mbulk_size()` (which could `assert` or read past the end). **Slot bounds** are parsed with `strtol`, checked for `ERANGE`, and clamped to `[0, MAX_CLUSTER_HSLOT]` with `min <= max` so `m_slot_to_shard` cannot be written out of bounds. **Host/port** bulks must be non-empty; hosts with embedded NULs are rejected.
> 
> Malformed shards are **logged and skipped**; wholly bad replies (non-array, empty topology, or every shard invalid) are ignored so existing connections—including bootstrap—stay up instead of tearing down all shards via `close_sc`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 721fc26a43dac930dd18fb3f33760889da74bc89. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cluster_client.cpp
+++ b/cluster_client.cpp
@@ -370,6 +370,13 @@ void cluster_client::handle_cluster_slots(protocol_response *r)
         return;
     }
 
+    // Track whether any shard in the reply passed validation. If every shard
+    // is malformed and we fall through the loop with no `close_sc[j] = false`
+    // anywhere, the close-stale-connections pass below would tear down EVERY
+    // existing connection (including the bootstrap), which contradicts the
+    // documented "bootstrap stays in service" invariant. (Cursor bugbot.)
+    bool any_valid_shard = false;
+
     // run over response and create connections
     for (unsigned int i = 0; i < r->get_mbulk_value()->mbulks_elements.size(); i++) {
         mbulk_element *shard_el = r->get_mbulk_value()->mbulks_elements[i];
@@ -468,8 +475,18 @@ void cluster_client::handle_cluster_slots(protocol_response *r)
             m_slot_to_shard[j] = sc->get_id();
         }
 
+        any_valid_shard = true;
         free(addr);
         free(port);
+    }
+
+    // If every shard in the reply was malformed and skipped, treat the reply
+    // as unusable -- skip the close-stale pass below so we don't disconnect
+    // the bootstrap and any other currently-live connections (cursor bugbot).
+    if (!any_valid_shard) {
+        benchmark_error_log("warning: CLUSTER SLOTS: every shard in the reply was malformed; "
+                            "leaving existing connections in service\n");
+        return;
     }
 
     // check if some connections left with no slots, and need to be closed
--- a/cluster_client.cpp
+++ b/cluster_client.cpp
@@ -350,22 +350,51 @@ void cluster_client::handle_cluster_slots(protocol_response *r)
     unsigned long prev_connections_size = m_connections.size();
     std::vector<bool> close_sc(prev_connections_size, true);
 
+    // Validate the top-level reply shape before walking it. as_mbulk_size()
+    // and as_bulk() both call assert(0) on type mismatch, and bare
+    // mbulks_elements[N] indexing is UB past-end. A malformed CLUSTER SLOTS
+    // reply from a misbehaving / hostile server (#417: fixture
+    // `cluster_slots_malformed.bin` from #409) hit both. Drop malformed
+    // shards instead of crashing; if the entire reply is unusable, the
+    // existing bootstrap connection stays in service.
+    if (r->get_mbulk_value() == NULL) {
+        benchmark_error_log("CLUSTER SLOTS: server returned non-array; ignoring reply\n");
+        return;
+    }
+
     // run over response and create connections
     for (unsigned int i = 0; i < r->get_mbulk_value()->mbulks_elements.size(); i++) {
-        // create connection
-        mbulk_size_el *shard = r->get_mbulk_value()->mbulks_elements[i]->as_mbulk_size();
+        mbulk_element *shard_el = r->get_mbulk_value()->mbulks_elements[i];
+        if (shard_el == NULL || !shard_el->is_mbulk_size()) {
+            benchmark_error_log("CLUSTER SLOTS: shard %u not an array; skipping\n", i);
+            continue;
+        }
+        mbulk_size_el *shard = shard_el->as_mbulk_size();
+        if (shard->mbulks_elements.size() < 3 || !shard->mbulks_elements[0]->is_bulk() ||
+            !shard->mbulks_elements[1]->is_bulk() || !shard->mbulks_elements[2]->is_mbulk_size()) {
+            benchmark_error_log("CLUSTER SLOTS: shard %u malformed (need [start, end, [host, port, ...]]); skipping\n",
+                                i);
+            continue;
+        }
 
         int min_slot = strtol(shard->mbulks_elements[0]->as_bulk()->value + 1, NULL, 10);
         int max_slot = strtol(shard->mbulks_elements[1]->as_bulk()->value + 1, NULL, 10);
 
+        mbulk_size_el *node = shard->mbulks_elements[2]->as_mbulk_size();
+        if (node->mbulks_elements.size() < 2 || !node->mbulks_elements[0]->is_bulk() ||
+            !node->mbulks_elements[1]->is_bulk()) {
+            benchmark_error_log("CLUSTER SLOTS: shard %u node tuple malformed (need host, port); skipping\n", i);
+            continue;
+        }
+
         // hostname/ip
-        bulk_el *mbulk_addr_el = shard->mbulks_elements[2]->as_mbulk_size()->mbulks_elements[0]->as_bulk();
+        bulk_el *mbulk_addr_el = node->mbulks_elements[0]->as_bulk();
         char *addr = (char *) malloc(mbulk_addr_el->value_len + 1);
         memcpy(addr, mbulk_addr_el->value, mbulk_addr_el->value_len);
         addr[mbulk_addr_el->value_len] = '\0';
 
         // port
-        bulk_el *mbulk_port_el = shard->mbulks_elements[2]->as_mbulk_size()->mbulks_elements[1]->as_bulk();
+        bulk_el *mbulk_port_el = node->mbulks_elements[1]->as_bulk();
         char *port = (char *) malloc(mbulk_port_el->value_len + 1);
         memcpy(port, mbulk_port_el->value + 1, mbulk_port_el->value_len);
         port[mbulk_port_el->value_len] = '\0';
--- a/cluster_client.cpp
+++ b/cluster_client.cpp
@@ -358,7 +358,15 @@ void cluster_client::handle_cluster_slots(protocol_response *r)
     // shards instead of crashing; if the entire reply is unusable, the
     // existing bootstrap connection stays in service.
     if (r->get_mbulk_value() == NULL) {
-        benchmark_error_log("CLUSTER SLOTS: server returned non-array; ignoring reply\n");
+        benchmark_error_log("warning: CLUSTER SLOTS: server returned non-array; ignoring reply\n");
+        return;
+    }
+
+    // A *valid* zero-shard reply would silently retire every existing
+    // connection (the close_sc[] loop further down). That's worse than
+    // crashing -- the benchmark continues with no shards. Reject it.
+    if (r->get_mbulk_value()->mbulks_elements.size() == 0) {
+        benchmark_error_log("warning: CLUSTER SLOTS: server returned empty topology; ignoring reply\n");
         return;
     }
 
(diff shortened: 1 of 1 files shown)


fcostaoliveira:

verdict: O(1) per-shard validation, no hot-path impact ✓

• **Validation scope**: Runs once per CLUSTER SLOTS topology refresh (not per request or connection). The checks are type predicates (is_mbulk_size(), is_bulk()) and array-size comparisons, all O(1). Only walks the shard array once linearly in the existing loop. ✓

• **Defensive indexing**: Replaces unsafe bare `mbulks_elements[N]` with guarded access (`if (shard->mbulks_elements.size() < 3)`). The comparison is cheaper than the bounds-check exception that would have fired if out-of-bounds had been caught. ✓

• **Error path**: benchmark_error_log() on malformed shard is a single formatted write (happens rarely, off the request loop). No repeated stderr I/O in the hot path. ✓

• **No new allocations**: The shard loop now caches `node = shard->mbulks_elements[2]->as_mbulk_size()` instead of recomputing it twice. This is a **net reduction** in repeated pointer dereferences (trading stack space for fewer indirections). ✓

**Summary**: Pure hardening with zero algorithmic cost. The per-shard validation is compile-time conditional checking; negligible cost during bootstrap, zero cost once topology is loaded. No request-loop changes.

fcostaoliveira:

- **Conventional Commits:** ✓ `fix(cluster-slots): ...` compliant; "Fixes #417" correct (auto-closes)
- **Validation logic:** ✓ All checks documented (type checks, bounds, array depth); clear repro in body
- **Test plan:** ✓ Mentions dropping `env.skip()` in follow-up; fuzzer fixture #419 unblocks
- **Known-bad-server-behavior docs:** Missing. Follow-up should document RESP validation rules in DEVELOPMENT.md

**Verdict:** Commits clean; follow-up PR should document RESP shape validation in reference docs (graceful skip behavior on malformed CLUSTER SLOTS).

fcostaoliveira:

**Verdict:** The skip-malformed-shard strategy is the right default for a benchmarking tool but the operational visibility gap needs to be addressed before this lands.

- **Continue-on-malformed vs. fail-fast (load-bearing):** Skipping a malformed shard means the benchmark can silently proceed with an incomplete topology: some key slots will have no connection, causing key-not-found or connection-refused errors that look like server problems rather than bootstrap failures. For a benchmarking tool — where the operator is intentionally stressing a specific topology — silent partial topology is often worse than a clean startup failure. Consider failing fast when the fraction of skipped shards exceeds a threshold (e.g. >0 malformed shards could warrant a warning; all shards malformed should be fatal).

- **No accounting of skipped shards in results/JSON:** `benchmark_error_log` writes to stderr but the skip is invisible in `--json-out-file`. A benchmark result produced against a partial topology is misleading. At minimum the count of skipped shards should appear in the JSON output so post-processing tooling can detect it.

- **Slot coverage gap is undetected:** After the loop, if some slot ranges [0-16383] are uncovered because their shard was skipped, the existing routing logic will silently fail at request time. A post-loop check that logs a warning when any slot range is unmapped would surface the problem at bootstrap rather than mid-run.

- **`strtol` on slot bounds still unguarded:** `min_slot` and `max_slot` are parsed with `strtol(..., NULL, 10)` from `as_bulk()->value + 1` (skipping the RESP length prefix). If `value` is empty or non-numeric, `strtol` returns 0 silently. Adding a non-null `endptr` check here — consistent with how the rest of the codebase validates `strtol` results — would close the last unvalidated field.

- **`mbulk_addr_el->value` offset assumption:** `mbulk_port_el->value + 1` skips the RESP bulk-string length prefix byte. This is inherited from the original code, but the validation added here doesn't verify that `value_len >= 1` before the `+1` pointer arithmetic. A zero-length bulk string for the port field would produce a pointer past the allocation. A `value_len > 0` guard on the port element would be cheap and complete the validation chain.

fcostaoliveira (commented):

- LOAD-BEARING: hostile-but-shape-valid replies still slip past. `mbulk_addr_el->value` is `memcpy`'d byte-for-byte into `addr` with a trailing `\0`, but the rest of the code (`strcmp(addr, m_connections[j]->get_address())`, `getaddrinfo(addr, ...)`) treats it as a C-string. A node host of `"127.0.0.1\x00evil.example.com"` (value_len=29) will be deduped/compared only up to the embedded NUL — a malicious cluster master can alias a benign-looking host with a different real target, or skip dedup entirely for an otherwise-identical entry. Worth a `memchr(value, '\0', value_len)` reject in this PR's validation block.
- `port` is built via `memcpy(port, mbulk_port_el->value + 1, value_len)` — the `+ 1` skips one byte of the RESP integer prefix and writes `value_len` bytes into a `value_len + 1` buffer. That reads one byte past `value` (still inside the parser's allocation in practice, so not OOB on the heap, but the trailing `\0` lands on a non-zero byte and `strtoul`/`getaddrinfo` later sees garbage). If a future RESP version emits a port as a non-integer bulk, the `+ 1` skip is wrong. The new validation only checks `is_bulk()`, not "starts with `:`" — consider asserting `value_len > 0` before the `+ 1` deref.
- Validation correctness LGTM: top-level NULL check, per-shard `is_mbulk_size()`, ≥3 children, `is_bulk()` on slot bounds, `is_mbulk_size()` + ≥2 `is_bulk()` children on the node tuple all match `as_mbulk_size()` / `as_bulk()`'s assert-on-mismatch contract (protocol.h:71). The fuzzer fixture `cluster_slots_malformed.bin` is correctly addressed.
- `strtol(value + 1, NULL, 10)` for slot bounds silently accepts non-numeric content (returns 0) and writes into `int min_slot` / `int max_slot`. A reply with `start=-1` or `start > max_slot` would corrupt `m_slot_to_shard[]` indexing in the loop below (`for j = min_slot; j <= max_slot`). Bounds check (`0 <= min_slot <= max_slot < REDIS_CLUSTER_SLOTS`) belongs in this validator.
- Skip-and-continue semantics are correct (the existing bootstrap connection stays in service) and `benchmark_error_log` is `printf`-style with literal format strings, so no format-string risk from server-controlled data.
- DoS via repeated malformed replies: each malformed shard logs but doesn't rate-limit; a server that returns a 1M-shard array of malformed entries makes you log 1M lines. Low priority — only happens on the (rare) topology-refresh path.

Verdict: NEEDS WORK. The NUL-in-host issue is a real trust-boundary bug worth catching in this PR; slot-range and `value+1` hardening are nice-to-have. The shape validation itself is solid.

fcostaoliveira (commented):

- `cluster_client.cpp:354-405` empty-shard reply still wipes topology — if a misbehaving server returns a *well-formed* but empty `CLUSTER SLOTS` (`*0\r\n`), `get_mbulk_value()` is non-NULL, the new shape check passes, the loop runs zero iterations, and the post-loop block at lines 408-420 sees every existing shard with `close_sc[i]==true` and tears down the whole topology. Pre-existing behavior, but this PR is the natural place to also guard `mbulks_elements.size() > 0` and log+keep current topology on empty reply, otherwise #417's "stay in service on malformed reply" promise doesn't hold for the empty-reply variant.
- `cluster_client.cpp:368-370` port memcpy underflow — for a hostile `$0\r\n` (zero-length) port bulk, `mbulk_port_el->value == NULL` and `value_len == 0`. The code does `malloc(1)` then `memcpy(port, NULL + 1, 0)`. memcpy with n=0 from a NULL-derived pointer is UB by strict C even though no bytes move. `value+1` is a pre-existing skip-the-RESP-`:`-prefix hack that doesn't apply to bulk strings (only to inline integers); for an actual `$len` host/port bulk you're now silently dropping the first byte of the payload. Worth tightening in this PR since you're already validating shape here.
- `cluster_client.cpp:358-359` integer slot bounds — `strtol(value + 1, ...)` assumes `value[0]` is a RESP type byte (e.g. `:`). Slot bounds in `CLUSTER SLOTS` are returned as RESP integers (`:`), which match — but `is_bulk()` also passes for `$bulk` payloads. A hostile server replying with `$5\r\n12345\r\n` for the slot bound would survive `is_bulk()` and then strtol would skip the `'1'` and parse `"2345"`. No crash, but slot range silently corrupted. Consider validating `value[0] == ':'` for the slot-bound elements, or routing through a typed helper.
- `cluster_client.cpp:354` no upper bound on shard count — a hostile server replying with `*1000000` followed by valid shards would cause a million `create_shard_connection`/`malloc` calls and `m_slot_to_shard[16384]` writes (which is sized for 16384 slots, so out-of-range max_slot indexes UB the array). Pre-existing, but since this PR is hardening the function, a sanity cap (≤16384 shards) would close the rest of the attack surface from #417.
- `cluster_client.cpp:362-365` host malloc never freed on later validation failure — not actually triggered here because all the new validation runs before the `malloc(addr)`. Just verified, OK.

Verdict: Comment

fcostaoliveira (commented):

Verdict: REQUEST CHANGES (one real DoS-shaped gap on empty outer reply); rest are LGTM.

- **Empty outer mbulk_value (zero shards) wipes topology.** `get_mbulk_value()` non-null but `mbulks_elements.size() == 0` falls through the validation and skips the for-loop entirely. `close_sc` was initialized all-`true`, so EVERY existing shard connection gets disconnected and `m_slot_to_shard` is never updated. A hostile or buggy server returning `*0\r\n` to CLUSTER SLOTS now silently tears down the live cluster topology mid-run — arguably worse than the crash this PR fixes, because the benchmark keeps running on a now-empty cluster. Recommend treating `mbulks_elements.empty()` the same as a non-array: log + return without touching `close_sc` / `m_connections`.
- **Slot-bound bulk `value == NULL` or `value_len == 0` still UB.** Validation passes `is_bulk()` but doesn't verify the bulk has content. A `$-1` (null) reply for slot start/end produces `value = NULL`, then `strtol(NULL + 1, ...)` is undefined. A `$0` empty bulk produces `value` allocated but `value_len == 0`, so `value + 1` reads past the allocation. Add a `value != NULL && value_len > 0` guard before the `strtol(value + 1, ...)` calls.
- **Negative slot values land in `m_slot_to_shard[j]` unchecked.** `strtol` happily accepts `-1` or `99999` from a bogus server. The subsequent `for (int j = min_slot; j <= max_slot; j++) m_slot_to_shard[j] = ...` writes out-of-bounds for negative `min_slot` or `max_slot >= REDIS_CLUSTER_HSLOTS`. Worth bounds-checking `[0, 16383]` while the shape validation is in this PR's scope (same fuzz fixture would reach it).
- **`is_bulk()` is true for non-integer bulks too.** `strtol("hello" + 1, ...)` returns 0 and reports no error. A malformed `*3\r\n$5\r\nhello\r\n$5\r\nworld\r\n*2\r\n...\r\n` reply would assign min_slot=0/max_slot=0 to every shard, collapsing the topology onto one shard. Consider validating that `strtol` consumed the whole field (endptr check) before using the result.
- **Port `memcpy(... value + 1, value_len)` is still off-by-one for $-bulk replies.** Pre-existing, not introduced here, but the PR touches this exact line — easy to fix in the same patch. For a bulk `$5\r\n12345\r\n` the value has no leading prefix, so `value + 1` skips the "1" and copies one byte past the buffer. (Works only because most servers send the port as an integer reply where the +1 skips the `:`.) Worth a follow-up note.
- **Node sub-array with zero `mbulks_elements` is handled correctly** by `< 2` check — good. Likewise the top-level non-array early-return preserves the bootstrap connection.

fcostaoliveira (commented):

**Verdict: clean; one naming note.**

- **`benchmark_error_log` messages lack a severity prefix** (`cluster_client.cpp`): all four new messages ("CLUSTER SLOTS: ...") go straight into the body with no "warning:" or "error:" prefix, while every other `benchmark_error_log` call in the same file either uses "warning:" or is a factual diagnostic line. These are skip-not-abort situations, so "warning:" is the right prefix; omitting it makes them look like info-level noise vs the surrounding "warning: --transaction pin connection ..." lines.

- **Comment explains what the guard does, not why the old code crashed** (`cluster_client.cpp`, top of the new validation block): "as_mbulk_size() and as_bulk() both call assert(0) on type mismatch" is accurate, but the deeper *why* — that malformed input from an untrusted server is realistic, not just hypothetical, per the fuzzer fixture — is in the PR description but not at the call site. A single sentence referencing the attack surface ("a misbehaving or proxy-mangling server can return a shorter array") would help a future maintainer understand why the guard must stay.

- **Validation structure repeats the `is_mbulk_size()`/`is_bulk()` pattern twice** for shard and node: this is fine for two levels, but if a future protocol revision adds a third level the reader has no hint about the nesting depth. A brief structural comment ("// CLUSTER SLOTS reply: [[start, end, [host, port, ...]], ...]") above the loop would serve as the schema reference that explains both checks at once.

- **Commit message headline truncated by GitHub** (`fix(cluster-slots): validate reply shape before indexing, log+skip ma...`): the body text is thorough, but the 72-char subject line was cut mid-word. Reword to fit: e.g. "fix(cluster-slots): validate reply shape, skip malformed shards (#417)".

- **Nothing else to flag** — new C++ uses the project's existing `strtol`/`malloc`/`memcpy` idioms; no Python test in this PR.

Issue #426, opened 2026-06-01
CLI-fuzz follow-ups: 17 crash/hang classes (11 share connection-stage retry-loop root cause)


fcostaoliveira:
Follow-up to #410 / PR #424. While narrowing the Hypothesis strategies to keep the new CLI fuzzer (`tests/cli_fuzz.py`) green on master, the agent surfaced **17 pre-existing crash/hang classes** in memtier_benchmark's option parsing + connection-stage retry loops. The fuzzer's strategy was trimmed so the regression net is stable; each of the bugs below remains live in master and is filed here for triage.

## Most actionable pattern

**11 of the 17 hang classes share one root cause**: once memtier enters a reconnect / retry loop after a *connection-stage* failure (AUTH, HELLO, SELECT, CLUSTER SLOTS, protocol negotiation), it ignores `--test-time` and runs until SIGKILLed. A single bounded *"give up after N seconds of consecutive connect-stage failures"* policy closes most of them.

| # | Repro (minimal argv) | Symptom |
|---|---|---|
| 1 | `--authenticate ''` against a server with no auth configured | hangs (reconnect loop, ignores `--test-time`) |
| 2 | `--cluster-mode` against standalone Redis | hangs (`cluster slot failed.` loop, ignores `--test-time`) |
| 3 | `--protocol memcache_text` (or `memcache_binary`) against Redis | hangs (`response parsing failed.` loop) |
| 4 | `--run-count -1` | SIGABRT, `std::bad_alloc` (negative passed to vector resize) |
| 5 | `--run-count 2147483647` (INT_MAX) | SIGABRT, `std::bad_alloc` (allocator OOM, no bounds check) |
| 6 | `--pipeline -1` | hangs (negative size, underflow loop) |
| 7 | `--data-size -1` | SIGABRT, `protocol.cpp:309: assert value_len > 0 failed` |
| 8 | `--data-size-range 1-9999999999` | hangs after `-ERR Protocol error: invalid bulk length` (reconnect loop) |
| 9 | `--data-size-list 0:50` | SIGABRT, value_len=0 assert |
| 10 | `--data-size-list 8:0` | hangs (zero weight, sampler picks nothing) |
| 11 | `--select-db 100` against default 16-DB Redis | hangs (`database selection failed.` loop) |
| 12 | `--command __key__` (bare placeholder, no command name) | SIGABRT, `protocol.cpp:774: first arg is not command name?` |
| 13 | `--command ''` | hangs |
[...]

PR #427, opened 2026-06-01
fix(opts): validate --command / --command-ratio at parse time (Refs #426 phase 2c)

fcostaoliveira:
## Summary

Three CLI-fuzz reproducers from #426 share the same root cause: missing
parse-time validation lets bad `--command` / `--command-ratio` input
through into the worker loop where it either trips a runtime assert or
hangs forever ignoring `--test-time`. This PR moves the checks to the
argument-parsing path so the user gets a readable error and exit 2
instead of a crash or unbounded hang.

- `--command __key__` (and other bare placeholders) is rejected at parse
  time instead of SIGABRTing inside `protocol.cpp::format_arbitrary_command`
  at the `first arg is not command name?` assert.
- `--command ''` (or whitespace-only) is rejected; previously
  `split_command_to_args()` returned true with zero argv tokens and the
  worker loop spun forever.
- `--command-ratio 0` (also empty / whitespace / negative) is rejected;
  `set_ratio()` is tightened so `strtoul("")==0` and negative wrap-around
  no longer slip through.

Monitor-placeholder commands (handled via the existing
`MONITOR_PLACEHOLDER_PREFIX` branch in `o_command`) are unaffected. The
existing `test_arbitrary_command_misses` / `test_monitor_input` /
`test_scan` suites still pass.

## Items closed (from #426)

- 12. `--command __key__` SIGABRT at `protocol.cpp:774`
- 13. `--command ''` hangs
- 14. `--command-ratio 0` hangs (also empty / whitespace / negative)

## Test plan

- [x] Manual repro of items 12, 13, 14 against the unfixed binary
  (SIGABRT / hang) and the fixed binary (clean exit 2 with readable
  stderr).
- [x] New regression tests in `tests/test_cli_validation_command.py`
  (4 cases: items 12, 13, 14 plus a happy-path `--command-ratio 1`
  end-to-end run); each guarded by a subprocess timeout so a future
  regression that re-introduces the hang fails the test loudly.
- [x] `make` clean.
- [x] Adjacent suites still green:
  `test_arbitrary_command_misses` (12/12), `test_monitor_input` (11/11),
  `test_scan` (15/15).

Refs #426

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

---

> [!NOTE]
> **Low Risk**
> CLI-only validation and stricter ratio parsing; monitor-placeholder `--command` paths are unchanged, with new subprocess tests guarding regressions.
> 
> **Overview**
> Adds **parse-time validation** for `--command` and `--command-ratio` so bad CLI input fails with readable stderr instead of **SIGABRT** or an **infinite worker hang** (issue #426 items 12–14).
> 
> For **`--command`**, empty or whitespace-only strings are rejected after parsing (zero argv tokens). The **first token** must be a real command name: placeholders such as `__key__`, `__data__`, `__scan_cursor__`, and monitor markers are detected with the same **substring rules** as `protocol.cpp`, so values like `FOO__key__` are caught too.
> 
> For **`--command-ratio`**, `set_ratio()` now rejects null/empty/whitespace, signs, non-integers, **zero**, `ERANGE` overflow, and values above **`UINT_MAX`** before they can truncate or wrap and starve the ratio sampler.
> 
> New regression tests in `tests/test_cli_validation_command.py` cover the three rejection cases plus a happy-path run with `--command-ratio 1` (subprocess timeouts guard against reintroduced hangs).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 30b1898eb275cf5286841d21240053362842f503. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/config_types.cpp
+++ b/config_types.cpp
@@ -503,12 +503,31 @@ bool arbitrary_command::set_key_pattern(const char *pattern_str)
 
 bool arbitrary_command::set_ratio(const char *ratio_str)
 {
-    char *q = NULL;
-    ratio = strtoul(ratio_str, &q, 10);
-    if (!q || *q != '\0') {
+    // Reject empty / null / whitespace-only / negative input. strtoul()
+    // silently accepts the empty string (returns 0) and wraps negatives to
+    // a huge unsigned value, both of which lead to a worker loop that never
+    // picks this command and hangs forever (issue #426 item 14).
+    if (ratio_str == NULL || *ratio_str == '\0') {
+        return false;
+    }
+    const char *first = ratio_str;
+    while (*first && isspace((unsigned char) *first)) {
+        first++;
+    }
+    if (*first == '\0' || *first == '-' || *first == '+') {
         return false;
     }
 
+    char *q = NULL;
+    unsigned long parsed = strtoul(ratio_str, &q, 10);
+    if (!q || *q != '\0') {
+        return false;
+    }
+    if (parsed == 0) {
+        return false;
+    }
+
+    ratio = (unsigned int) parsed;
     return true;
 }
 
--- a/config_types.cpp
+++ b/config_types.cpp
@@ -519,11 +519,17 @@ bool arbitrary_command::set_ratio(const char *ratio_str)
     }
 
     char *q = NULL;
+    errno = 0;
     unsigned long parsed = strtoul(ratio_str, &q, 10);
     if (!q || *q != '\0') {
         return false;
     }
-    if (parsed == 0) {
+    // ERANGE catches the strtoul overflow case; the narrow-to-unsigned-int
+    // check catches values above UINT_MAX that strtoul accepted on a 64-bit
+    // unsigned long without ERANGE (e.g. "4294967296" on x86_64). Without
+    // either guard a too-large value silently truncates and a wrap-to-0
+    // value re-introduces the very hang we're guarding against.
+    if (errno == ERANGE || parsed == 0 || parsed > UINT_MAX) {
         return false;
     }
 
--- a/memtier_benchmark.cpp
+++ b/memtier_benchmark.cpp
@@ -742,15 +742,19 @@ static bool verify_cluster_option(struct benchmark_config *cfg)
 // the user gets a readable error instead of SIGABRT.
 static bool first_arg_is_placeholder(const std::string &token)
 {
-    if (token == KEY_PLACEHOLDER || token == DATA_PLACEHOLDER || token == SCAN_CURSOR_PLACEHOLDER ||
-        token == MONITOR_RANDOM_PLACEHOLDER) {
+    // The runtime assert in protocol.cpp ("first arg is not command name?")
+    // fires on substring containment via memmem(), not exact equality, so
+    // a literal like `FOO__key__` or `setkey__key__` still SIGABRTs without
+    // this guard. Mirror the same substring contract here at parse time so
+    // the user gets a readable error instead of an abort.
+    if (token.find(KEY_PLACEHOLDER) != std::string::npos || token.find(DATA_PLACEHOLDER) != std::string::npos ||
+        token.find(SCAN_CURSOR_PLACEHOLDER) != std::string::npos ||
+        token.find(MONITOR_RANDOM_PLACEHOLDER) != std::string::npos) {
         return true;
     }
-    // __monitor_lineN__ (any digits / @) - the placeholder prefix is the
-    // recognised marker. The whole-command monitor placeholder path is
-    // handled separately upstream; here we just guard against it leaking
-    // into the first-arg slot of a regular --command.
-    if (token.compare(0, strlen(MONITOR_PLACEHOLDER_PREFIX), MONITOR_PLACEHOLDER_PREFIX) == 0) {
+    // __monitor_lineN__ (any digits / @) -- the placeholder prefix is the
+    // recognised marker; same substring rule.
+    if (token.find(MONITOR_PLACEHOLDER_PREFIX) != std::string::npos) {
         return true;
     }
     return false;
--- a/memtier_benchmark.cpp
+++ b/memtier_benchmark.cpp
@@ -734,6 +734,28 @@ static bool verify_cluster_option(struct benchmark_config *cfg)
     return true;
 }
 
+// Returns true when `token` is one of the placeholder tokens that
+// arbitrary-command parsing recognises (__key__, __data__, __monitor_line*__,
+// __scan_cursor__). The first argv of --command must be a literal command
+// name; allowing a placeholder there is what triggers the runtime assert in
+// protocol.cpp ("first arg is not command name?"). We check at parse time so
+// the user gets a readable error instead of SIGABRT.
+static bool first_arg_is_placeholder(const std::string &token)
+{
+    if (token == KEY_PLACEHOLDER || token == DATA_PLACEHOLDER || token == SCAN_CURSOR_PLACEHOLDER ||
+        token == MONITOR_RANDOM_PLACEHOLDER) {
+        return true;
+    }
+    // __monitor_lineN__ (any digits / @) - the placeholder prefix is the
+    // recognised marker. The whole-command monitor placeholder path is
+    // handled separately upstream; here we just guard against it leaking
+    // into the first-arg slot of a regular --command.
+    if (token.compare(0, strlen(MONITOR_PLACEHOLDER_PREFIX), MONITOR_PLACEHOLDER_PREFIX) == 0) {
+        return true;
+    }
+    return false;
+}
+
 static bool verify_arbitrary_command_option(struct benchmark_config *cfg)
 {
     if (cfg->key_pattern) {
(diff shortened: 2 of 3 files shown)


fcostaoliveira:

verdict: parse-time only, zero steady-state impact ✓

• **Validation firing**: Runs once at argv parsing before any connections are established. All checks are string prefix matching and integer comparisons (first_arg_is_placeholder, split_command_to_args validation). Zero hot-path involvement. ✓

• **Error messages**: Clear parse-time exits (fprintf + return -1) prevent hang/SIGABRT. These run once and are user-facing, not instrumentation that fires repeatedly. ✓

• **No rate-limiting**: The ratio validation (set_ratio()) rejects empty/negative/zero inputs at parse time; no per-request sampling overhead introduced. Rejects pathological configs that would spin in worker loops. ✓

• **Net improvement**: By rejecting zero-ratio and placeholder-first-arg at parse time, we prevent infinite loops and asserts that would have eaten CPU and latency. This is a net performance gain for invalid configs (fast failure vs. hang). ✓

**Summary**: Pure parse-time safety, zero request-loop overhead. No steady-state allocations, syscalls, or lock contention. Prevents bugs that would have degraded performance (spinning workers, SIGABRT restarts).

fcostaoliveira:

- **Conventional Commits:** ✓ `fix(opts): ...` compliant; "Refs #426 phase 2c" correct (no auto-close)
- **Validation rules:** ✓ Documented (no empty `--command`, no placeholder first arg, `--command-ratio > 0`)
- **Test plan:** ✓ Regression tests with timeouts; adjacent suites green (12/12, 11/11, 15/15)
- **DEVELOPMENT.md:** Missing. Should document `--command-ratio` and `--command` validation rules (positive int, no placeholders)

**Verdict:** Commits clean; DEVELOPMENT.md needs CLI option ranges and placeholder rules.

fcostaoliveira:

**Verdict:** Correct fixes for the three #426 items; the bespoke per-flag validation pattern is the main design question for this series.

- **Shared `validate_int_option` helper absent (load-bearing for the series):** PRs 427, 428, and 429 each introduce their own inline integer-validation pattern: `optarg_is_negative()` in 428, `strtol`+`errno` in 429, whitespace-skip+sign-check in 427's `set_ratio()`. The logic is subtly different in each (e.g. 427 checks `*first == '+'` as invalid for ratio but 428 does not for pipeline). A single `static int parse_bounded_uint(const char *optarg, unsigned long *out, unsigned long min, unsigned long max, const char **err_msg)` helper in `memtier_benchmark.cpp` would unify all three patterns, eliminate the divergence, and make future option additions trivially consistent. The bespoke approach is fine for a one-shot fix but becomes a maintenance liability as the series grows.

- **`first_arg_is_placeholder` is a closed enumeration:** The function hard-codes `KEY_PLACEHOLDER`, `DATA_PLACEHOLDER`, `SCAN_CURSOR_PLACEHOLDER`, `MONITOR_RANDOM_PLACEHOLDER`, and the `MONITOR_PLACEHOLDER_PREFIX` prefix. If a new placeholder is added (e.g. for client-id or thread-id injection), this guard must be updated manually or it silently passes the new placeholder through to the runtime assert. A design that queries the placeholder detection from a single source of truth (e.g. a `is_placeholder_token(token)` function alongside where placeholders are defined) would be more maintainable.

- **`set_ratio` rejects `+` prefix but the error message says "positive integer":** Rejecting `+1` (which `strtoul` would parse as 1) while the error message says "must be a positive integer" is surprising UX. Either accept `+N` (consistent with most POSIX tools) or update the error to say "unsigned decimal integer with no sign prefix".

- **`cmd.command_args[0].data` type assumes string-like `.data` member:** The check `first_arg_is_placeholder(cmd.command_args[0].data)` passes a field named `.data` to a function taking `const std::string &`. This is fine if `command_arg::data` is `std::string`, but it would silently stop working if the type changes. A brief comment noting the assumed type, or using `.data.c_str()` explicitly, would make the coupling visible.

- **No validation that `--command-ratio` appears after `--command`:** `get_last_command()` is called to apply the ratio, which means `--command-ratio` before any `--command` will either crash or apply to a stale command. The existing behaviour predates this PR but the new, tighter `set_ratio` path makes the ordering constraint more visible — a parse-time check for `cfg->arbitrary_commands` being non-empty before calling `get_last_command()` would close this gap.

fcostaoliveira (commented):

- LOAD-BEARING: `set_ratio()` correctly hardens against the strtoul negative-wrap path — explicit reject of leading `-`/`+`, empty, whitespace-only, and `parsed == 0` after the trailing-junk check. The `unsigned long` -> `unsigned int` narrow at the end is unchecked: input `"4294967296"` (= 2^32) overflows on 64-bit `unsigned long` -> truncates to 0 in `unsigned int`, and the prior `parsed == 0` check fires only on the `unsigned long` value, not the narrowed one. Bigger inputs like `"8589934593"` (2^33 + 1) narrow to 1 silently. Tighten with `if (parsed > UINT_MAX) return false;` or check `errno == ERANGE`.
- `errno` is not checked after `strtoul` — input `"99999999999999999999"` sets `errno = ERANGE` and returns `ULONG_MAX`, which passes the `parsed != 0` and `*q == '\0'` checks and then narrows. Same fix: add `errno = 0;` before the call and reject on `ERANGE`.
- `first_arg_is_placeholder()` correctly handles the bare-placeholder case at the literal command-name slot; the `MONITOR_PLACEHOLDER_PREFIX` startswith check is defensive enough (a real Redis command would never start with `__`).
- `--command ''` and whitespace-only correctly land in the `command_args.empty()` branch (via `split_command_to_args`'s zero-token return). Good.
- `isspace((unsigned char) *first)` cast is correct (avoids UB on negative `char`); the surrounding code uses `isspace(*p)` unguarded in config_types.cpp:525,581,601 — pre-existing, not a regression here, but worth a separate hardening pass.
- No format-string / injection surface: error messages use literal format specifiers; `optarg`'s content is only printed via `%s`. No shell exposure.

Verdict: NEEDS MINOR FIX. The `unsigned long` -> `unsigned int` narrow + missing `errno = ERANGE` check is a real (low-severity) bypass — `--command-ratio 4294967296` would still be accepted as ratio=0 today.

fcostaoliveira (commented):

- `memtier_benchmark.cpp:746-757` `first_arg_is_placeholder` only catches *exact-match* tokens, but the assert at `protocol.cpp:774` fires whenever the first arg *contains* `__key__`, `__data__`, or equals `SCAN_CURSOR_PLACEHOLDER`. `--command "FOO__key__"`, `--command "SET__key__"`, `--command "x__data__y"` all still set `current_arg->type = key_type` / `data_type` for i==0 and hit the SIGABRT this PR was supposed to close. Either mirror `protocol.cpp`'s `data.find(PLACEHOLDER) != npos` semantics, or move the check to immediately *before* the assert in `format_arbitrary_command` and convert the assert to a graceful error.
- `config_types.cpp:506-525` `isspace((unsigned char) *first)` — `<ctype.h>` is not directly `#include`d in `config_types.cpp` (it comes in transitively today through other STL headers, but that's fragile). Add an explicit `#include <ctype.h>` so a future header reshuffle doesn't break the build silently.
- `memtier_benchmark.cpp:1463-1467` placeholder rejection message reads the placeholder via `cmd.command_args[0].data.c_str()` — `split_command_to_args()` may have stripped surrounding quotes/whitespace before the check, so the printed token can differ from the raw user input. Cosmetic, but the error message becomes harder to grep for a fuzz harness that quotes its inputs.
- `config_types.cpp:524` `unsigned long parsed = strtoul(...)` followed by `(unsigned int) parsed` — no `> UINT_MAX` check. `--command-ratio 4294967296` on 64-bit silently truncates to 0 and then the `parsed == 0` guard rejects, so the user-visible behavior is fine here, but values like `4294967297` truncate to 1 and pass. Low impact but inconsistent with the strictness of the other phase-2 PRs (#428, #429) which both added `> MAX` caps.
- Cursor-flagged leading `+` rejected and `strtoul` truncation bypass — acknowledged, not repeated.

Verdict: Request changes (the placeholder-substring bypass re-opens the SIGABRT this PR claims to fix)

fcostaoliveira (commented):

Verdict: LGTM with one paths-coverage note.

- **Invocation paths**: there is no config-file feature and `parse_uri()` does not handle `--command` or `--command-ratio`, so the long-option case is the only entry — validation correctly fires on every path. (Confirmed by reading `parse_uri` and `config_parse_args`.)
- **Cap boundary for ratio**: `set_ratio()` rejects 0, accepts every positive value up to `ULONG_MAX`, then truncates to `unsigned int`. So `--command-ratio 4294967296` (UINT_MAX+1) truncates to `1` silently — not catastrophic, but inconsistent with the new "positive integer" promise. Consider rejecting `parsed > UINT_MAX`.
- **First-arg placeholder check uses `==` for `KEY_PLACEHOLDER` / `DATA_PLACEHOLDER` / `SCAN_CURSOR_PLACEHOLDER` but `compare(0, strlen(PREFIX), PREFIX) == 0` for the monitor prefix.** That asymmetry means `--command __key__suffix` (a literal token that happens to start with `__key__` but isn't the placeholder) is *accepted*, while `--command __monitor_lineXXX__` (anything starting with the monitor prefix) is rejected. Probably intentional — the monitor prefix is a family — but worth a comment so this doesn't drift.
- **Multiple `--command` invocations**: each one goes through `first_arg_is_placeholder()` independently — good. But the `set_ratio` rejection branch references `cmd.command_name.c_str()` from `get_last_command()` — if the user passes `--command-ratio 0` *before* any `--command`, `get_last_command()` returns the implicit/default command (or whatever `arbitrary_commands_list` initialises with). Worth a quick check that no crash/UB occurs in that argv order; the existing code likely already handles it but the new error path didn't add a guard.
- **Whitespace handling in `set_ratio`**: rejects leading `+` and `-` after stripping spaces, but `"  5  "` (trailing whitespace) — strtoul stops at the space, `*q != '\0'` triggers rejection. Good. `"\t5"` — `isspace((unsigned char) '\t')` is true so the leading-sign check works correctly. Good.
- **`split_command_to_args()` returning `true` with `cmd.command_args.empty()`** is now caught at parse time — good. But the second branch (`first_arg_is_placeholder(cmd.command_args[0].data)`) reads `[0]` immediately after; if a future refactor lets `split_command_to_args()` return true with zero args without the empty-check above it, this becomes an OOB read. The two checks are sequenced correctly today, just brittle to reorder.

fcostaoliveira (commented):

**Verdict: clean; two minor notes.**

- **`first_arg_is_placeholder` comment explains what the function checks, not why the first-arg slot is special** (`memtier_benchmark.cpp`): the block comment describes what each placeholder token is, but the *why* — that `format_arbitrary_command()` in `protocol.cpp` unconditionally forwards `argv[0]` as the Redis command name without a placeholder expansion step, making a placeholder there structurally unformattable — is absent. One sentence on that invariant would make the guard self-documenting.

- **`set_ratio` whitespace skip uses `isspace()`; `optarg_is_negative` in PR #428 (same series) uses literal `' ' || '\t'`** (`config_types.cpp`): these are sister functions added in consecutive PRs. Using `isspace((unsigned char))` in one and manual space/tab in the other is a style inconsistency. Since they're both in the same fix series, aligning them (either direction) before merge would avoid future "which one is canonical?" confusion.

- **Error message phrasing differs from the PR-428 companion** (`memtier_benchmark.cpp`): the new `--command-ratio` rejection reads "must be a positive integer (got '...')"; the analogous `--pipeline` message in #428 reads "must be greater than zero." Both are correct but the extra `(got 'X')` parenthetical is not applied consistently. Either style is fine — pick one for the series.

- **Python test: `_run_memtier` docstring says "A timeout guards against regressions"** — this is a *why* comment and reads well. Consistent with the project style.

- **Commit message: single-commit, Conventional Commits-clean** (`fix(opts): validate --command / --command-ratio at parse time`); body enumerates all three items, which is good.

PR #428, opened 2026-06-01
fix(opts): validate --pipeline / --data-size / --data-size-list size (Refs #426 phase 2b)

fcostaoliveira:
## Summary

Phase 2b of #426: reject three size-related CLI values at argv-parse time so the binary errors out cleanly instead of hanging or aborting deep in the run loop.

- `--pipeline` must be > 0 and <= 1024 (saturates most NICs; tighter than UINT_MAX).
- `--data-size` must be > 0 and <= 512 MiB (matches Redis default `proto-max-bulk-len`).
- `--data-size-list` entry sizes must be > 0 (same 512 MiB cap per entry).
- Negative `optarg` (leading `-`) is now detected explicitly before `strtoul`, which previously cast `-1` to a huge unsigned value and sailed past every downstream check.

## Items closed (from #426)

- [x] item 6 — `--pipeline -1` no longer hangs (negative passed through unsigned, underflow loop).
- [x] item 7 — `--data-size -1` no longer SIGABRTs at `protocol.cpp:309 assert value_len > 0`.
- [x] item 9 — `--data-size-list 0:50` (zero **SIZE**) no longer SIGABRTs with `value_len=0`.

**Explicitly NOT in scope:** `--data-size-list 8:0` (zero **WEIGHT**, item 10) is left for **phase 3 (sampler safety)**. The zero-weight case is a sampler/loop issue, not an argv-range issue — fixing it cleanly belongs with the rest of the sampler-construction safety work.

## Caps chosen

| Flag | Lower bound | Upper bound | Rationale |
|---|---|---|---|
| `--pipeline` | > 0 | 1024 | Pipeline depth past ~1024 saturates the NIC/CPU; anything larger is almost certainly a typo or a negative-cast underflow. |
| `--data-size` | > 0 | 512 MiB | Matches Redis' default `proto-max-bulk-len = 512mb`; larger sizes the server would refuse anyway. |
| `--data-size-list` entry size | > 0 | 512 MiB | Same proto-max-bulk-len cap, applied per entry. |

## Test plan

Regression tests live in `tests/test_cli_validation_size.py`. They invoke the binary directly via `subprocess` — no Redis server needed — and check for non-zero exit + the expected error substring on stderr.

- [x] `--pipeline=-1` rejected
- [x] `--pipeline=0` rejected
- [x] `--pipeline=1025` rejected (cap)
- [x] `--pipeline=2` accepted (parses cleanly)
- [x] `--data-size=-1` rejected
- [x] `--data-size=0` rejected
- [x] `--data-size-list=0:50` rejected
- [x] `--data-size-list=8:50,16:50` accepted (parses cleanly)

Run locally:

```
TEST=test_cli_validation_size.py OSS_STANDALONE=1 ./tests/run_tests.sh
```

All 8 cases pass on this branch.

Refs #426

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

---

> [!NOTE]
> **Low Risk**
> Changes are limited to argument validation and new CLI tests; no benchmark runtime or network behavior changes for valid inputs.
> 
> **Overview**
> Phase 2b of #426 hardens **argv parsing** for size-related flags so bad values fail immediately with clear stderr messages instead of hanging or aborting during a run.
> 
> **`--pipeline`** and **`--data-size`** now reject leading-minus values via `optarg_is_negative()` (before `strtoul` can turn `-1` into a huge unsigned), enforce **> 0**, and cap at **1024** and **512 MiB** respectively. **`--data-size-list`** rejects entries with **size 0** and applies the same per-entry size cap.
> 
> New subprocess tests in `tests/test_cli_validation_size.py` cover the three original #426 repros plus cap and “valid value parses” cases, with no Redis server required.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 855c543b67bca80a58ae81884cd638c7079f9a3f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/memtier_benchmark.cpp
+++ b/memtier_benchmark.cpp
@@ -762,6 +762,29 @@ static bool verify_arbitrary_command_option(struct benchmark_config *cfg)
     return true;
 }
 
+// Validation caps for size-related CLI options (Refs #426 phase 2b).
+//   PIPELINE_MAX:  upper bound on --pipeline.  A 1024-deep pipeline already
+//                  saturates most NICs; values beyond that almost always
+//                  reflect a typo (e.g. negative input cast to unsigned).
+//   DATA_SIZE_MAX: upper bound on --data-size (and per-entry size in
+//                  --data-size-list).  Matches Redis' default proto-max-bulk-len
+//                  (512 MiB), past which the server rejects the bulk anyway.
+#define MEMTIER_PIPELINE_MAX 1024u
+#define MEMTIER_DATA_SIZE_MAX (512u * 1024u * 1024u)
+
+// True if optarg's first non-space character is '-'.  strtoul() silently
+// accepts a leading minus and underflows the result to a huge unsigned, which
+// then sails past the >0 check downstream and only manifests as a hang or
+// SIGABRT deep in the run loop.  Catching it at parse time keeps the error
+// message close to the offending flag.
+static bool optarg_is_negative(const char *s)
+{
+    if (s == NULL) return false;
+    while (*s == ' ' || *s == '\t')
+        s++;
+    return *s == '-';
+}
+
 static int config_parse_args(int argc, char *argv[], struct benchmark_config *cfg)
 {
     enum extended_options
@@ -1122,22 +1145,42 @@ static int config_parse_args(int argc, char *argv[], struct benchmark_config *cf
                 return -1;
             }
             break;
-        case o_pipeline:
+        case o_pipeline: {
             endptr = NULL;
-            cfg->pipeline = (unsigned int) strtoul(optarg, &endptr, 10);
-            if (!cfg->pipeline || !endptr || *endptr != '\0') {
+            if (optarg_is_negative(optarg)) {
                 fprintf(stderr, "error: pipeline must be greater than zero.\n");
                 return -1;
             }
+            unsigned long pipeline_val = strtoul(optarg, &endptr, 10);
+            if (!pipeline_val || !endptr || *endptr != '\0') {
+                fprintf(stderr, "error: pipeline must be greater than zero.\n");
+                return -1;
+            }
+            if (pipeline_val > MEMTIER_PIPELINE_MAX) {
+                fprintf(stderr, "error: pipeline must be <= %u.\n", MEMTIER_PIPELINE_MAX);
+                return -1;
+            }
+            cfg->pipeline = (unsigned int) pipeline_val;
             break;
-        case 'd':
+        }
+        case 'd': {
             endptr = NULL;
-            cfg->data_size = (unsigned int) strtoul(optarg, &endptr, 10);
-            if (!cfg->data_size || !endptr || *endptr != '\0') {
+            if (optarg_is_negative(optarg)) {
                 fprintf(stderr, "error: data-size must be greater than zero.\n");
                 return -1;
             }
+            unsigned long data_size_val = strtoul(optarg, &endptr, 10);
+            if (!data_size_val || !endptr || *endptr != '\0') {
+                fprintf(stderr, "error: data-size must be greater than zero.\n");
+                return -1;
+            }
+            if (data_size_val > MEMTIER_DATA_SIZE_MAX) {
+                fprintf(stderr, "error: data-size must be <= %u bytes (512 MiB).\n", MEMTIER_DATA_SIZE_MAX);
+                return -1;
+            }
+            cfg->data_size = (unsigned int) data_size_val;
             break;
+        }
         case 'R':
             cfg->random_data = true;
             break;
@@ -1161,6 +1204,27 @@ static int config_parse_args(int argc, char *argv[], struct benchmark_config *cf
             if (!cfg->data_size_list.is_defined()) {
                 fprintf(stderr, "error: data-size-list must be expressed as [size1:weight1],...[sizeN:weightN].\n");
                 return -1;
+            }
+            // Reject zero-SIZE entries (e.g. "0:50") -- they trip the
+            // value_len > 0 assert in protocol.cpp at first SET.  Zero-WEIGHT
+            // entries (e.g. "8:0") are a separate sampler-safety issue tracked
+            // for phase 3.
+            for (std::vector<config_weight_list::weight_item>::iterator it = cfg->data_size_list.item_list.begin();
+                 it != cfg->data_size_list.item_list.end(); ++it) {
+                if (it->size == 0) {
+                    fprintf(stderr,
+                            "error: data-size-list entries must have size > 0 "
+                            "(got 0 in '%s').\n",
+                            optarg);
+                    return -1;
+                }
+                if (it->size > MEMTIER_DATA_SIZE_MAX) {
+                    fprintf(stderr,
+                            "error: data-size-list entry size must be <= %u bytes "
+                            "(512 MiB), got %u.\n",
+                            MEMTIER_DATA_SIZE_MAX, it->size);
+                    return -1;
+                }
             }
             break;
         case o_expiry_range:
--- a/tests/test_cli_validation_size.py
+++ b/tests/test_cli_validation_size.py
@@ -1,0 +1,162 @@
+"""
+CLI validation regression tests for size-related flags (Refs #426 phase 2b).
+
+Covers the three repros from issue #426 that previously hung or crashed:
+
+  * --pipeline -1   (item 6)  -> negative passed through unsigned cast,
+                                 underflow drove an infinite retry loop.
+  * --data-size -1  (item 7)  -> SIGABRT, protocol.cpp:309 assert value_len > 0.
+  * --data-size-list 0:50 (item 9) -> same SIGABRT (zero-SIZE entry).
+
+All three must now be rejected at argv-parse time with a readable error and
+exit code 2 (via usage()), well before any socket is opened.  Companion checks
+ensure the upper caps (pipeline <= 1024, data-size <= 512 MiB) trip cleanly
+and that a known-good value still parses.
+
+These tests run the binary directly with subprocess -- no Redis server is
+needed because the parser bails out before connect.
+
+Run with:
+  TEST=test_cli_validation_size.py OSS_STANDALONE=1 ./tests/run_tests.sh
+"""
+import subprocess
+
+from include import MEMTIER_BINARY
+
+
+def _run_memtier(args):
+    """Invoke memtier_benchmark with *args* and return the CompletedProcess.
+
+    No server is contacted: every case in this file is expected to fail at
+    argument parsing time.
+    """
+    return subprocess.run(
+        [MEMTIER_BINARY] + list(args),
+        capture_output=True,
+        text=True,
+        timeout=10,
+    )
+
+
+def _assert_reject(env, args, needle, label):
+    """Run memtier with *args* and assert it exits non-zero with *needle* in stderr."""
+    result = _run_memtier(args)
+    env.assertNotEqual(
+        result.returncode, 0,
+        message="{}: expected non-zero exit, got 0 (stderr={!r})".format(
+            label, result.stderr),
+    )
+    env.assertTrue(
+        needle in result.stderr,
+        message="{}: expected {!r} in stderr, got {!r}".format(
+            label, needle, result.stderr),
+    )
+
+
+# ---------------------------------------------------------------------------
+# --pipeline (#426 item 6)
+# ---------------------------------------------------------------------------
+
+def test_pipeline_negative_rejected(env):
+    """--pipeline=-1 used to hang in a retry loop; must now fail at parse."""
+    _assert_reject(
+        env,
+        ["--pipeline=-1", "--test-time=1"],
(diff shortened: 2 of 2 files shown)


fcostaoliveira:

verdict: parse-time validation, zero request-loop impact ✓

• **Validation scope**: --pipeline, --data-size, --data-size-list validation runs once during argv parsing. All checks are integer comparisons and magnitude bounds (strtoul, range caps PIPELINE_MAX=1024, DATA_SIZE_MAX=512MiB). Zero hot-loop involvement. ✓

• **optarg_is_negative() helper**: Three branch-free pointer comparisons per flag (while *s isspace, then *s=='-'). Called once per invocation. Negligible cost that prevents silent unsigned-underflow hangs. ✓

• **Data-size-list iteration**: Iterates once at parse time over the entry list (not per-request). Each entry checks size bounds; O(n) where n = number of size buckets (typically <10). No request-loop overhead. ✓

• **Zero allocations added**: All validation uses stack temporaries and existing config structures. No new heap churn in the benchmark run. ✓

**Summary**: Pure parse-time hardening. The cap checks (pipeline<=1024, data-size<=512MiB) prevent configs that would either hang (pipeline too large causing backpressure loops) or crash (data-size causing OOM). Zero steady-state CPU, memory, or lock overhead.

fcostaoliveira:

- **Conventional Commits:** ✓ `fix(opts): ...` compliant; "Refs #426 phase 2b" correct (no auto-close)
- **Caps documented:** ✓ Table shows bounds (--pipeline 1–1024, --data-size 1–512 MiB, --data-size-list same per entry)
- **Negative handling:** ✓ Leading-minus detection before strtoul explained
- **DEVELOPMENT.md:** Missing. Should list all CLI cap ranges (pipeline, data-size, data-size-list 512 MiB limit)

**Verdict:** Commits clean; DEVELOPMENT.md needs comprehensive CLI bounds documentation.

fcostaoliveira:

**Verdict:** The fixes are correct and the caps are well-motivated, but the constant placement and naming deserve attention before the series is done.

- **`#define` constants instead of named header constants (load-bearing for the series):** `MEMTIER_PIPELINE_MAX` (1024) and `MEMTIER_DATA_SIZE_MAX` (512 MiB) are defined via `#define` inside `memtier_benchmark.cpp`, invisible to any external tooling, test harness, or future protocol adapter that might need to enforce the same limits. Moving them to `memtier_benchmark.h` as `static const unsigned` or `constexpr` values — or to a dedicated `memtier_limits.h` — would make them part of the documented API surface. The same applies to `MAX_RUN_COUNT` in PR #429: all three caps should live together so an operator reading the header can see what the tool will and won't accept.

- **`optarg_is_negative` duplicates logic that `strtol` already provides:** PR #429 uses `strtol` (correctly) to detect negatives via the sign of the return value. PR #428 introduces a separate `optarg_is_negative` helper that re-implements the sign check before calling `strtoul`. These two patterns should be unified: either use `strtol` everywhere (as in #429) or use `optarg_is_negative` + `strtoul` everywhere, but not both patterns in the same file.

- **Upper cap on `--data-size` uses `unsigned int` but `MEMTIER_DATA_SIZE_MAX` is `512u * 1024u * 1024u = 536870912`:** `cfg->data_size` is `unsigned int` (32-bit on all supported platforms) and 512 MiB fits, but the error message prints `MEMTIER_DATA_SIZE_MAX` with `%u` while the local variable `data_size_val` is `unsigned long`. On LP64 Linux this is fine; on LLP64 Windows (if ever ported) `%u` would mismatch `unsigned long`. Using `%lu` for the format and `(unsigned long)MEMTIER_DATA_SIZE_MAX` for the cast would be strictly correct.

- **`data-size-list` zero-SIZE check iterates the already-parsed list, zero-WEIGHT deferred to phase 3:** The iterator pattern here (C++03-style `iterator it = ...begin()`) is inconsistent with the range-for loops used elsewhere in the file. A range-for `for (const auto &item : cfg->data_size_list.item_list)` would match the surrounding style and be more readable. The deferred zero-weight case (item 10) is clearly marked — the comment referencing phase 3 is appropriate and appreciated.

- **No test for the 512 MiB cap itself:** `test_cli_validation_size.py` tests `--pipeline=1025` (one over the pipeline cap) but has no equivalent for `--data-size` above 512 MiB. Adding a `test_data_size_above_cap_rejected` case with e.g. `--data-size=536870913` would give the same coverage symmetry as the pipeline tests and guard against future accidental cap changes.

fcostaoliveira (commented):

- LOAD-BEARING: `optarg_is_negative()` correctly catches the `-1`-wrap path for `--pipeline` and `--data-size`, but `--data-size-list` still has the same bug at a lower layer: `config_weight_list::config_weight_list(const char *str)` at config_types.cpp:140 calls `strtoul(str, &p, 10)` on each `size:weight` segment with no `-` pre-check. `--data-size-list=-1:50` parses as size=`ULONG_MAX` -> narrows to `UINT_MAX`, then PASSES the new `it->size > MEMTIER_DATA_SIZE_MAX` check… wait, no — `UINT_MAX > 512 MiB`, so it actually trips the upper bound here. But `--data-size-list=-1:50,8:50` produces `UINT_MAX` for the first entry and is correctly caught. Worth pinning down with a regression test for `-1:50` explicitly.
- `strtoul` `errno = ERANGE` is not checked for `--pipeline` / `--data-size` — `"99999999999999999999"` returns `ULONG_MAX` and is then rejected by the `> 1024` / `> 512 MiB` cap, so the security outcome is correct, but the explicit `errno = 0` + `errno == ERANGE` check would make the intent clear and protect future maintainers who might raise the caps.
- `unsigned long` -> `unsigned int` narrow at `cfg->pipeline = (unsigned int) pipeline_val;` is safe because the cap is 1024; same for `data_size` (cap 512 MiB fits in unsigned int). Good.
- The 512 MiB cap matching `proto-max-bulk-len` is a sensible defense against DoS via giant `evbuffer_add` allocations downstream (each thread × client × pipeline holds a value-sized buffer; with the old unbounded cast, `data_size=-1` would have tried to allocate ~4 GiB per buffer × N clients).
- `optarg_is_negative` only strips space/tab, not vertical whitespace — fine for argv but inconsistent with the `isspace((unsigned char) *first)` loop in PR #427's `set_ratio`. Cosmetic.
- No format-string / credential risk in the new error paths.

Verdict: LGTM with one optional tightening (explicit `errno == ERANGE` reject for `strtoul`) and a request for an explicit regression test on `--data-size-list=-1:50`. The phase 2b scope is correctly delivered.

fcostaoliveira (commented):

- `memtier_benchmark.cpp:780-786` `optarg_is_negative` only skips `' '` and `'\t'`. `strtoul()` per C also accepts `\n\v\f\r` as leading whitespace, so `optarg = "\n-1"` (yes, contrived via env/script edge cases) would slip through this check and underflow into `strtoul`. Cheap fix: use `isspace((unsigned char) *s)` (and `#include <ctype.h>`).
- `memtier_benchmark.cpp:1167-1183` no check that the parsed value fits in `unsigned int`. `--pipeline 4294967297` on 64-bit truncates via `(unsigned int)` to 1 silently — caught accidentally by the `> MEMTIER_PIPELINE_MAX (1024)` cap, so user-visible behavior is fine, but worth a comment that the cap is what guards truncation. Same applies to `--data-size`.
- `memtier_benchmark.cpp:1208-1228` the new per-entry size validation only runs *after* `config_weight_list` has already parsed and stored every entry. `config_weight_list(optarg)` itself uses `strtoul` on each `size:weight` token; a hostile entry like `--data-size-list=99999999999:1,8:50` will silently truncate the first size to `unsigned int` and may end up *passing* the `> 512 MiB` cap depending on truncation. Recommend either (a) capturing original strings for re-validation, or (b) tightening `config_weight_list` itself to reject overflows.
- `memtier_benchmark.cpp:1208-1227` overlaps with PR #430 — both add a `for (... item_list)` loop in the same `case 'a':` arm (one checks `size==0`, the other `weight==0`). Merging order matters: whichever lands first will leave the other PR with a non-trivial conflict in the `case 'a':` block. Suggest pre-merging the loops in one of the PRs.
- `tests/test_cli_validation_size.py:90-99` the "valid passes parse" tests connect to port 1 and assert `"pipeline must be" not in stderr`. memtier's connect-failure path also prints to stderr; if a future error path ever includes the substring `pipeline must be` for a different reason the test would silently mis-pass. Tighten to an exit-code check or a positive "parsed --pipeline=2" assertion.
- The 512 MiB cap on `--data-size` is `unsigned`. `cfg->data_size` is `unsigned int`; on a platform where `unsigned int` is 16-bit (none we ship, but worth a `static_assert`) this overflows. Trivial belt-and-suspenders.

Verdict: Comment

fcostaoliveira (commented):

Verdict: LGTM with cap-boundary and per-entry-negative gaps.

- **Invocation paths**: `parse_uri()` does not handle `--pipeline` / `--data-size` / `--data-size-list`, and there's no config-file feature — long-option is the only entry, validation correctly catches every path.
- **Cap-boundary tests missing**: the test suite covers `1025` rejected and `2` accepted but not `1024` (cap) and `1023` (cap-1). Same for `--data-size`: `512*1024*1024` (cap, accepted) and `512*1024*1024+1` (cap+1, rejected) are not exercised. The `>` comparison is correct for the cap-as-inclusive-upper-bound intent, but a regression to `>=` would silently break the documented boundary. Add at least one cap-value-accepted test.
- **`--data-size-list` per-entry negatives bypass `optarg_is_negative`**: the helper is only called for `--pipeline` and `--data-size`. `--data-size-list=-1:50` flows through `config_weight_list` where `strtoul` wraps the `-1` to ULONG_MAX, then truncation to `unsigned int` yields `UINT_MAX`. The new `it->size > MEMTIER_DATA_SIZE_MAX` check catches it — good — but the error message ("entry size must be <= 512 MiB") misleads the user about what they actually typed. Consider applying `optarg_is_negative` to the whole `--data-size-list` optarg, or scanning for `-` in any entry before the cap check.
- **`strtoul` ERANGE not consulted**: `--pipeline=99999999999999999999` overflows ULONG_MAX → `strtoul` returns ULONG_MAX and sets errno=ERANGE; the code doesn't `errno = 0` before the call and doesn't check ERANGE. It's saved by the `> 1024` cap, but the error message ("must be <= 1024") is again misleading on truly out-of-range input. The neighboring `--run-count` PR (#429) does this correctly (`errno = 0; ... errno == ERANGE`).
- **`optarg_is_negative` whitespace skip is narrower than `isspace`**: only skips `' '` and `'\t'`. `--pipeline=\v-1` or `\n-1` wouldn't trip the negative check. Unlikely from a CLI in practice, but `isspace((unsigned char) *s)` matches the convention used in `set_ratio()` in PR #427.
- **Cast safety**: `pipeline_val` (unsigned long) is bounded to 1024 before the cast to `unsigned int` — safe. `data_size_val` similarly bounded to 512 MiB before the cast — safe. Per-entry `it->size` is already `unsigned int` from `config_weight_list`, so the cap check is on the post-truncation value — `it->size > MEMTIER_DATA_SIZE_MAX` correctly catches `UINT_MAX` from the negative-wrap case.

fcostaoliveira (commented):

**Verdict: clean; two notes on the new helper.**

- **`optarg_is_negative` skips only space and tab; `set_ratio` in the companion PR #427 uses `isspace()`** (`memtier_benchmark.cpp`): both helpers serve the same purpose (strip leading whitespace before inspecting the first character) but use different whitespace definitions. `isspace((unsigned char))` is the C idiom and covers all locale whitespace; the manual `' ' || '\t'` is narrower. Aligning the two before the series merges avoids a "which one is the intended contract?" question for later readers.

- **`optarg_is_negative` comment explains *what* it detects, not *why* it's needed alongside `strtoul`** (`memtier_benchmark.cpp`): the comment correctly names the underflow root cause (`strtoul` silently wraps), but the *why it must be a separate pre-check* — rather than just adding a post-check on the parsed value — is that `strtoul("−1", ...)` on some platforms returns `ULONG_MAX` and sets `errno=ERANGE`, while on others it wraps silently with no error. The pre-check is the platform-portable solution; saying so makes the guard obviously non-redundant.

- **Error message `"data-size must be <= %u bytes (512 MiB)."` embeds the literal unit** — the companion `data-size-list` message in PR #430 will say "weight greater than zero" (words, no symbol). Using `> 0` for size and words for weight in adjacent error messages is a minor inconsistency; words throughout would read more uniformly.

- **`#define MEMTIER_PIPELINE_MAX 1024u` / `MEMTIER_DATA_SIZE_MAX` placed after a function definition** (`memtier_benchmark.cpp`): project convention groups `#define` constants near the top of the file with other globals (see line 90 area). Moving these up avoids them being read as local-to-`optarg_is_negative`.

- **Commit message headline truncated** (`fix(opts): validate --pipeline / --data-size / --data-size-list size …`): subject line was cut by GitHub at 72 chars. Suggest: "fix(opts): validate --pipeline / --data-size / --data-size-list (#426)".

PR #430, opened 2026-06-01
fix(opts): validate sampler configs (--data-size-list, --key-stddev, --key-pattern=G:G) (Refs #426 phase 3)

fcostaoliveira:
## Summary

Phase 3 of #426: reject sampler configurations that can never produce a
value, at parse / construction time, instead of letting them hang the
worker loop or trip the Gaussian assert (SIGABRT) at runtime.

Closes the following items from #426:

- **Item 10** -- `--data-size-list 8:0`: a zero-weight bucket made the
  size sampler skip every entry, so it spun forever picking nothing.
  Now rejected with `error: data-size-list entries must have weight
  greater than zero`.
- **Item 15** -- `--key-pattern G:G --key-stddev inf` (or `nan`): the
  Gaussian rejection sampler can never satisfy `val < min || val > max + 1`
  on a non-finite stddev. NaN also slipped through the old `<= 0` guard
  because NaN compares false to everything. Now rejected with
  `error: key-stddev must be a finite number greater than zero`.
- **Item 16** -- `--key-pattern G:G` over a 1-key range: the Gaussian
  path asserts `median > min && median < max`, which is impossible when
  the range has fewer than 2 buckets (SIGABRT on master). Now rejected
  with `error: key-pattern=G requires a key range spanning at least 3
  keys`.

All three paths exit with code 2 (the parser-error convention used
elsewhere in `memtier_benchmark.cpp`).

## Test plan

Regression coverage in `tests/test_cli_validation_sampler.py`:

- [x] `--data-size-list=8:0` -> exit 2 with weight diagnostic
- [x] `--key-pattern G:G --key-stddev inf` -> exit 2 with finite diagnostic
- [x] `--key-pattern G:G --key-stddev nan` -> exit 2 with finite diagnostic
- [x] `--key-pattern G:G --key-minimum=1 --key-maximum=1` -> exit 2 with G-range diagnostic
- [x] `--key-pattern G:G --key-minimum=1 --key-maximum=1000 --key-stddev=100` (valid) -> exit 0

Manual repros from #426 confirmed fixed:

- `memtier_benchmark --data-size-list 8:0 --test-time=1` -> no hang, exit 2
- `memtier_benchmark --key-pattern G:G --key-stddev inf --test-time=1` -> no hang, exit 2
- `memtier_benchmark --key-pattern G:G --key-maximum 1 --test-time=1` -> no SIGABRT, exit 2

`make format-check` clean. `tests/test_mget_validation.py` still green
(no regression on the adjacent validation suite).

Refs #426

---

> [!NOTE]
> **Low Risk**
> Changes are input validation and parser-only tests; no runtime benchmark logic or security-sensitive paths beyond clearer early failures.
> 
> **Overview**
> Phase 3 of **#426** hardens CLI validation so impossible sampler settings fail at parse time with **exit code 2** instead of hanging workers or **SIGABRT** in the Gaussian path.
> 
> **`--data-size-list`** now rejects **zero-weight** entries (e.g. `8:0`), alongside existing checks for zero size and oversize values. **`--key-stddev`** must be a **finite** number **> 0** (`inf` / `nan` are rejected). After defaults apply, **`--key-pattern` with `G`** requires at least a **3-key span** (`key-maximum - key-minimum >= 2`), handles inverted min/max, and enforces a user **`--key-median`** strictly inside `(min, max)`.
> 
> New RLTest coverage lives in **`tests/test_cli_validation_sampler.py`**. CI sets **`PARALLELISM: "2"`** on the two OSS-cluster API matrix jobs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dd972702d766c4e8ec1319cdbba090b168033970. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -179,10 +179,10 @@ jobs:
             env: { OSS_STANDALONE: "1", OSS_CLUSTER: "0", TLS: "1", TLS_PROTOCOLS: "TLSv1.3", VERBOSE: "1" }
             tls: true
           - name: "OSS-CLUSTER API: TCP Plaintext"
-            env: { OSS_STANDALONE: "0", OSS_CLUSTER: "1", VERBOSE: "1" }
+            env: { OSS_STANDALONE: "0", OSS_CLUSTER: "1", VERBOSE: "1", PARALLELISM: "2" }
             tls: false
           - name: "OSS-CLUSTER API: TCP TLS"
-            env: { OSS_STANDALONE: "0", OSS_CLUSTER: "1", TLS: "1", VERBOSE: "1" }
+            env: { OSS_STANDALONE: "0", OSS_CLUSTER: "1", TLS: "1", VERBOSE: "1", PARALLELISM: "2" }
             tls: true
           - name: "OSS-CLUSTER API 49 shards: TCP TLS"
             env: { RLTEST_DEBUG: "1", LOG_LEVEL: "notice", OSS_STANDALONE: "0", OSS_CLUSTER: "1", TLS: "1", VERBOSE: "1", SHARDS: "49", TEST: "tests_oss_simple_flow:test_rate_limited_completion_no_hang" }
--- a/memtier_benchmark.cpp
+++ b/memtier_benchmark.cpp
@@ -3065,16 +3065,34 @@ int main(int argc, char *argv[])
 
     config_init_defaults(&cfg);
 
-    // Reject Gaussian (G) key-pattern over a degenerate 1-key range: the sampler
-    // requires `median > min && median < max`, which is impossible when the range
-    // has fewer than 2 buckets. Without this check, the Gaussian path trips an
-    // assert (SIGABRT). Issue #426.
+    // Reject Gaussian (G) key-pattern over a degenerate range. The sampler
+    // (gaussian_noise::gaussian_distribution_range) handles `min == max` by
+    // returning `min` directly, but the user almost certainly didn't mean a
+    // single-key Gaussian -- it's a no-op, indistinguishable from
+    // --key-pattern=R on a 1-key range. The actual SIGABRT path was a user
+    // also passing a `--key-median` outside (min, max) on a tiny range; the
+    // default `len/2.0 + min + 0.5` lands strictly inside even for the 2-key
+    // case (min=1, max=2 → median=1.5), so reject only `max <= min`.
+    // (Review-pass fix: the original `(max - min) < 2` cutoff was too strict
+    // and rejected the perfectly valid 2-key case.)
     if (cfg.key_pattern && (cfg.key_pattern[key_pattern_set] == 'G' || cfg.key_pattern[key_pattern_get] == 'G')) {
-        if (cfg.key_maximum < cfg.key_minimum || (cfg.key_maximum - cfg.key_minimum) < 2) {
+        if (cfg.key_maximum <= cfg.key_minimum) {
             fprintf(stderr,
-                    "error: key-pattern=G requires a key range spanning at least 3 keys "
-                    "(key-maximum - key-minimum >= 2); got key-minimum=%llu key-maximum=%llu.\n",
+                    "error: key-pattern=G requires key-maximum > key-minimum (a non-empty range "
+                    "for the Gaussian sampler); got key-minimum=%llu key-maximum=%llu.\n",
                     cfg.key_minimum, cfg.key_maximum);
+            exit(2);
+        }
+        // If the user supplied a non-zero median, it must land strictly inside
+        // (min, max) -- otherwise the sampler's `assert(median > min && median
+        // < max)` SIGABRTs. Default median=0 means "auto-compute" and is
+        // always valid because the auto-compute keeps it strictly inside.
+        if (cfg.key_median != 0.0 &&
+            (cfg.key_median <= (double) cfg.key_minimum || cfg.key_median >= (double) cfg.key_maximum)) {
+            fprintf(stderr,
+                    "error: --key-median=%g must be strictly inside (key-minimum=%llu, key-maximum=%llu) "
+                    "for the Gaussian sampler.\n",
+                    cfg.key_median, cfg.key_minimum, cfg.key_maximum);
             exit(2);
         }
     }
--- a/memtier_benchmark.cpp
+++ b/memtier_benchmark.cpp
@@ -1162,6 +1162,18 @@ static int config_parse_args(int argc, char *argv[], struct benchmark_config *cf
                 fprintf(stderr, "error: data-size-list must be expressed as [size1:weight1],...[sizeN:weightN].\n");
                 return -1;
             }
+            // Reject entries with zero weight: the size sampler would skip them and
+            // (if every entry has weight 0) loop forever picking nothing. Issue #426.
+            for (std::vector<config_weight_list::weight_item>::iterator it = cfg->data_size_list.item_list.begin();
+                 it != cfg->data_size_list.item_list.end(); ++it) {
+                if (it->weight == 0) {
+                    fprintf(stderr,
+                            "error: data-size-list entries must have weight greater than zero "
+                            "(got size=%u weight=0).\n",
+                            it->size);
+                    return -1;
+                }
+            }
             break;
         case o_expiry_range:
             cfg->expiry_range = config_range(optarg);
@@ -1210,8 +1222,10 @@ static int config_parse_args(int argc, char *argv[], struct benchmark_config *cf
         case o_key_stddev:
             endptr = NULL;
             cfg->key_stddev = strtod(optarg, &endptr);
-            if (cfg->key_stddev <= 0 || !endptr || *endptr != '\0') {
-                fprintf(stderr, "error: key-stddev must be greater than zero.\n");
+            // Reject non-finite stddev (NaN, +/-inf): the Gaussian sampler's
+            // rejection loop would never produce an in-range key. Issue #426.
+            if (!endptr || *endptr != '\0' || !std::isfinite(cfg->key_stddev) || cfg->key_stddev <= 0) {
+                fprintf(stderr, "error: key-stddev must be a finite number greater than zero.\n");
                 return -1;
             }
             break;
@@ -3050,6 +3064,20 @@ int main(int argc, char *argv[])
     }
 
     config_init_defaults(&cfg);
+
+    // Reject Gaussian (G) key-pattern over a degenerate 1-key range: the sampler
+    // requires `median > min && median < max`, which is impossible when the range
+    // has fewer than 2 buckets. Without this check, the Gaussian path trips an
+    // assert (SIGABRT). Issue #426.
+    if (cfg.key_pattern && (cfg.key_pattern[key_pattern_set] == 'G' || cfg.key_pattern[key_pattern_get] == 'G')) {
+        if (cfg.key_maximum < cfg.key_minimum || (cfg.key_maximum - cfg.key_minimum) < 2) {
+            fprintf(stderr,
+                    "error: key-pattern=G requires a key range spanning at least 3 keys "
+                    "(key-maximum - key-minimum >= 2); got key-minimum=%llu key-maximum=%llu.\n",
+                    cfg.key_minimum, cfg.key_maximum);
+            exit(2);
+        }
+    }
 
     // Open the failed-keys log if requested. Failure is reported but doesn't
     // abort the benchmark.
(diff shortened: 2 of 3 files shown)


fcostaoliveira:

- **Conventional Commits:** ✓ `fix(opts): ...` compliant; "Refs #426 phase 3" correct (no auto-close)
- **Validation rules:** ✓ Three cases documented (data-size-list weight > 0, key-stddev finite > 0, Gaussian >= 3 keys)
- **Test plan:** ✓ Manual repros confirmed fixed; regression tests + valid case; adjacent suite green
- **DEVELOPMENT.md:** Missing. Should document sampler validation (key-stddev finite > 0, Gaussian min 3 keys, data-size-list weight > 0)

**Verdict:** Commits clean; DEVELOPMENT.md needs sampler validation documentation.

fcostaoliveira:

verdict: parse-time sampler config validation, prevents runtime hangs ✓

• **data-size-list weight validation**: Iterates once at parse time over weight entries; rejects zero-weight buckets that would break the size sampler. The check is a simple equality test (it->weight == 0) per entry. Zero request-loop overhead. ✓

• **key-stddev finitude check**: Uses std::isfinite() at parse time to reject inf/nan inputs. The Gaussian rejection sampler would loop forever on non-finite stddev; early rejection prevents worst-case hang. Single parse-time call, zero per-request cost. ✓

• **key-pattern=G range validation**: Checks degenerate 1-key ranges (key_maximum - key_minimum < 2) at startup before the sampler runs. Prevents assert() in the Gaussian median calculation. One range comparison per startup. ✓

• **No sampler loop changes**: The rejection-sampler runs during the benchmark; by gating pathological configs at parse time, we prevent infinite loops/asserts without touching the hot-path sampler code. ✓

**Summary**: Pre-benchmark validation for distribution samplers. The checks (finitude, non-zero weights, range size) are single evaluations that prevent infinite loops or asserts during the benchmark run. Zero steady-state per-request overhead; actually *improves* latency by rejecting hang-prone configs early.

fcostaoliveira:

**Verdict:** The three fixes are correct and the test coverage is solid; the placement of the G-range check (in `main()` after `config_init_defaults`) is the main design concern.

- **G-range check in `main()` after `config_init_defaults` (load-bearing):** The `key-pattern=G` range check runs in `main()` after defaults are applied, which is correct for catching the interaction between `--key-minimum` and `--key-maximum`. However, the check uses `exit(2)` rather than returning through the normal `-1 → usage() → exit(2)` path that every other option error uses. This inconsistency means the error bypasses any cleanup code that might run between the check and `exit` in future. Using `return 2` (or routing through the same path as the parser errors) would align with the existing convention.

- **`key_maximum - key_minimum < 2` check uses unsigned subtraction:** `key_minimum` and `key_maximum` are `unsigned long long`. If `key_minimum > key_maximum` (an already-invalid configuration), the subtraction wraps to a huge positive value, bypassing the `< 2` check entirely. The PR guards against this with `cfg.key_maximum < cfg.key_minimum || (cfg.key_maximum - cfg.key_minimum) < 2`, which is correct — but the existing parser allows `key_minimum > key_maximum` through without error (it only checks `< 1`). A note or a follow-up issue to add `key_minimum < key_maximum` validation at parse time would close the gap.

- **`--key-stddev` `std::isfinite` fix is in the right place but NaN-input requires `<cmath>`:** `std::isfinite` is declared in `<cmath>`. The diff does not show a `#include <cmath>` addition. If the existing includes already pull it in transitively (likely through OpenSSL or libevent headers), this works today but is fragile. An explicit `#include <cmath>` near the other math includes would make the dependency explicit.

- **Zero-weight `data-size-list` check duplicated across phase 2b (PR #428) and phase 3 (PR #430):** PR #428 adds a zero-SIZE check in the `data-size-list` parser; PR #430 adds a zero-WEIGHT check in the same loop position. When both PRs land, the merged code will have two adjacent iterator loops over `cfg->data_size_list.item_list`. They should be merged into a single loop that checks both `size == 0` and `weight == 0`, ideally with a shared `validate_data_size_list_entries()` helper.

- **Sampler test `test_key_pattern_g_valid_range_accepted` uses `--test-time=1` without server:** The valid-case test invokes the binary against `_common_args(env)` (a live server port) with `--test-time=1`. If the server is not available or responds slowly, the test may hang for a full second. Using `--requests=1` instead of `--test-time=1` would make the happy-path test faster and more deterministic, consistent with how `test_cli_validation_command.py` structures its happy-path case.

fcostaoliveira (commented):

- LOAD-BEARING: the `--key-stddev` fix is correct — the new `!std::isfinite(cfg->key_stddev) || cfg->key_stddev <= 0` chain handles NaN (which compared false to everything and slipped through the old `<= 0` guard), `+inf`, and `-inf`. Note: `<cmath>` is already included at memtier_benchmark.cpp:75, so `std::isfinite` resolves cleanly. The same NaN-slips-through-comparison anti-pattern exists for other `strtod` users — `--miss-rate-threshold` at line 1415 already has the correct guard, but a quick `grep strtod` audit (e.g. `--key-stddev` was the only buggy one) confirms scope is closed.
- `--data-size-list` weight=0 reject correctly bounds the sampler-loop-forever DoS. Combined with PR #428's size=0 check, the size sampler's input domain is now fully constrained.
- `--key-pattern=G` range check (`max - min >= 2`) at memtier_benchmark.cpp:3075 correctly defends the Gaussian sampler assert. The subtraction is done on `unsigned long long` (`key_minimum` / `key_maximum`); the `cfg.key_maximum < cfg.key_minimum` check before the subtract correctly prevents unsigned underflow. Good.
- Defaults interaction: `config_init_defaults` sets `key_maximum = 10000000` if unset (memtier_benchmark.cpp:580), and `key_minimum` defaults to 0, so the common case stays unaffected. The validator runs after defaults are applied — verified at the call site.
- `exit(2)` here is consistent with the rest of the parser-error convention but slightly inconsistent with the `-1 -> usage() -> exit(2)` flow used elsewhere in argv parsing. Cosmetic.
- No format-string / credential exposure. The `%llu` matches `unsigned long long` cleanly. No new attacker-controlled input is logged.

Verdict: LGTM. Phase-3 sampler-safety scope is correctly delivered; the NaN/inf handling for `key_stddev` is the kind of subtle floating-point bug that's easy to miss and this catches it cleanly.

fcostaoliveira (commented):

- `memtier_benchmark.cpp:3068-3079` G-range check is overly strict: rejects `--key-minimum=1 --key-maximum=2` (range 1, max-min=1, fewer than 2) as well as `--key-minimum=1 --key-maximum=3` (range 2, max-min=2 — accepted). For min=1/max=2 the Gaussian median 1.5 satisfies `median>min && median<max`, so this would have worked. The current cutoff is more conservative than the actual sampler invariant. Either relax to `< 1` (i.e. reject only true degenerate ranges) and let the sampler invariant catch the edge cases, or keep the conservative bound and reword the error from "3 keys" to something the sampler actually demands.
- `memtier_benchmark.cpp:3073` `cfg.key_maximum < cfg.key_minimum` is checked first (good — avoids unsigned underflow in the subtraction) but the resulting error message conflates two cases ("max<min" and "range too small"). Two messages would be friendlier.
- `memtier_benchmark.cpp:3070` the check guards on `cfg.key_pattern[set] == 'G' || cfg.key_pattern[get] == 'G'`. If only the *get* pattern is G but `--ratio` is `1:0` (set-only), the run will never invoke the Gaussian sampler and a degenerate range would otherwise be fine. False positive in a real benchmark config. Probably acceptable trade-off for safety, but worth a note in the message ("required because ANY of set/get uses G").
- `memtier_benchmark.cpp:1162-1174` overlaps with PR #428 — both add a `for (... item_list)` loop in the same `case 'a':` arm. Whichever lands first will leave the other PR with a non-trivial conflict; consider pre-merging the loops into a single validation pass (`size==0`, `weight==0`, plus 512MiB cap) in one of the PRs.
- `memtier_benchmark.cpp:1225` `key_stddev` check ordering: `!endptr || *endptr != '\0' || !std::isfinite(...) || <= 0` — if `endptr == NULL` is taken, the later checks never run, so safe. `strtod("nan")` sets `*endptr='\0'`, then `!isfinite` rejects. `strtod("")` sets `endptr=optarg`, so `*endptr` is `'\0'`. That accepts empty string with stddev=0 — wait, `cfg->key_stddev <= 0` catches it. OK, defensible, but the empty-input case relies on the `<= 0` arm and not the `endptr == optarg` arm; an explicit `endptr == optarg` check would be more readable.
- `memtier_benchmark.cpp:3078` uses `exit(2)` directly while every other parse error in this file `return -1`s through `usage() -> exit(2)`. Same exit code so user-observable behavior matches, but the inconsistency means this path skips `usage()`-driven cleanup if any is ever added.

Verdict: Comment

fcostaoliveira (commented):

Verdict: LGTM with one scope gap on arbitrary-command per-command key patterns.

- **Invocation paths**: only the long-option case for `--key-stddev` and `--data-size-list`; the G-range check runs after `config_init_defaults()` so defaults (key_max=10_000_000) apply. `parse_uri` doesn't touch any of these. All paths covered.
- **Per-command G pattern is not checked.** `cfg.arbitrary_commands->at(i).key_pattern == 'G'` (e.g. `--command "GET __key__" --command-key-pattern G`) reaches the Gaussian sampler over the global `key_minimum/key_maximum` range too, and would re-trigger the same SIGABRT this PR closes. The new validation only inspects the top-level `cfg.key_pattern[set]/[get]`. Worth either widening the check or filing as a follow-up.
- **Cap boundary**: `(key_maximum - key_minimum) < 2` admits diff=2 (e.g. min=1, max=3 → 3 distinct keys), rejects diff=1 and diff=0. Edge: `key_minimum=0` (struct default, no `--key-minimum` given) + `key_maximum=2` → diff=2 → admitted, which is the smallest legal range. Edge: `key_minimum > key_maximum` is caught explicitly. Good — but tests cover only diff=0; consider diff=1 (cap-1) and diff=2 (cap) for completeness.
- **`std::isfinite` correctly rejects NaN, +inf, -inf** since NaN compares false to everything (the old `<= 0` check explicitly fails on NaN). `strtod("inf", ...)` returns HUGE_VAL with no errno, `strtod("nan", ...)` returns NaN — both rejected. Good.
- **`strtod` ERANGE not consulted**: `--key-stddev=1e9999` overflows to HUGE_VAL (which IS finite on glibc — it's INFINITY for typical doubles, so `isfinite` catches it). Underflow `1e-9999` returns 0 with errno=ERANGE → caught by the `<= 0` check. So effectively fine, but checking errno would be tidier and consistent with #429.
- **G-range check uses `exit(2)` directly** rather than `return -1` like the rest of `config_parse_args`. The PR comment says "matches usage() → exit(2)" — but the rest of `config_parse_args` returns -1 and the caller does `usage(); exit(2);`. Here the check is in `main()` (post-init_defaults), where direct `exit(2)` is more consistent with the other post-defaults validations (e.g. `--clients-start`, `--thread-conn-start-min-jitter-micros` use `exit(1)`). Mixing 1 and 2 in the same block is the smaller inconsistency — consider unifying.

fcostaoliveira (commented):

**Verdict: one structural inconsistency worth addressing.**

- **Gaussian key-range check is in `main()` after `config_init_defaults()` rather than in `config_parse_args()`** (`memtier_benchmark.cpp`): every other CLI validation returns -1 from `config_parse_args`, which then calls `usage()` and exits with code 2. The Gaussian check is a post-parse validation that calls `exit(2)` directly. The code comment says it must run after `config_init_defaults` (because defaults fill in the key range), which is the correct reason — but this should be called out explicitly, otherwise it looks like an oversight rather than a deliberate placement. A brief "// must follow config_init_defaults(): key_minimum/maximum defaults are applied there" would make it self-documenting.

- **Error message phrasing inconsistency with PR #428**: the new `data-size-list` weight-zero message says "weight greater than zero" (words); the size-zero message in PR #428 says "size > 0" (symbol). These two messages will appear adjacent in the same `case o_data_size_list:` block after both PRs land; aligning them to the same form avoids a jarring inconsistency.

- **`std::isfinite` check moves before `<= 0` check** (`memtier_benchmark.cpp`, `o_key_stddev` case): the reordering is correct because `NaN <= 0` is always false. The existing comment says "Reject non-finite stddev" which is *what*, not *why*. The *why* — that NaN's unordered comparison with `<= 0` yields false so the old guard didn't catch it — is the key insight worth preserving.

- **Python test: all five cases use `env.skipOnCluster()`** (`test_cli_validation_sampler.py`): the skip is harmless but unnecessary for tests that reject before any network activity. A comment noting that the skip is conservative (the parser fires before connect, so cluster/standalone doesn't matter) would clarify intent.

- **Commit message headline truncated** (`fix(opts): validate sampler configs (--data-size-list, --key-stddev, …`): suggest "fix(opts): validate sampler configs for zero-weight, non-finite stddev, G-range (#426)".

Issue #13, opened 2016-07-29
build error in Ubuntu 1510


simonxiaoss:
root@simon-d14-1:~/benchmark/memtier_benchmark-1.2.7# ./configure
................
checking for gai_strerror in -lsocket... no
./configure: line 5344: syntax error near unexpected token `LIBEVENT,'
./configure: line 5344: `PKG_CHECK_MODULES(LIBEVENT,'

root@simon-d14-1:~/benchmark/memtier_benchmark-1.2.7# cat /etc/*release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=15.10

yossigo:
@simonxiaoss Looks like autoconf doesn't have pkg-config support, make sure you have the `pkg-config` package installed.

simonxiaoss:
Thanks. Yes, I missed the pkg-config package.

PR #433, opened 2026-06-02
fix(reconnect,retry): cap exponential backoff at 60s

fcostaoliveira:
## Problem

`shard_connection.cpp` had two exponential-backoff multiplications with no upper bound:

1. **Reconnect backoff** (`m_current_backoff_delay *= reconnect_backoff_factor`) — triggered on every connection error when `--reconnect-on-error` is set.
2. **Retry backoff** (`m_current_retry_backoff_ms *= retry_backoff_factor`) — triggered on each retry drain schedule.

With a factor of 2.0 and `--max-reconnect-attempts=0` (unlimited), the delay doubles on every attempt. After ~30 doublings, the next scheduled reconnect or retry is approximately 34 years in the future. `event_add()` silently absorbs the absurd `timeval` without error, and the benchmark goes effectively dark — no reconnects, no retries, no visible failure — instead of surfacing the underlying connectivity problem.

## Fix

Adds two file-scope constants near the top of `shard_connection.cpp`:

```cpp
// Maximum backoff delay enforced after every exponential-backoff multiplication.
// Without a cap, a factor of 2.0 and --max-reconnect-attempts=0 (unlimited)
// causes the delay to double on every attempt: after ~30 doublings the next
// scheduled reconnect or retry fires in ~34 years. event_add() silently absorbs
// that value, making the benchmark go effectively dark instead of surfacing the
// underlying failure.  60 s is a practical upper bound: tight enough to remain
// responsive, large enough to avoid thundering-herd reconnect storms.
static const double MEMTIER_BACKOFF_CAP_SEC = 60.0;
static const double MEMTIER_BACKOFF_CAP_MS  = 60000.0;
```

After each `*= factor` multiplication:
- Reconnect delay is clamped to `MEMTIER_BACKOFF_CAP_SEC` (60 s)
- Retry delay is clamped to `MEMTIER_BACKOFF_CAP_MS` (60 000 ms)

No parser changes, no flag defaults changed, no semantic changes outside the runtime multiplication sites.

## Test

Adds `test_reconnect_backoff_cap_60s` to `tests/test_reconnections.py`. It targets a guaranteed-closed port with `--reconnect-backoff-factor=4.0 --max-reconnect-attempts=0` for 10 s, then:
- Asserts the process does not crash with SIGABRT.
- Parses every `"attempting reconnection … in X.XX seconds"` line from stderr and asserts `max(delays) <= 60.0 s`.

Refs: 2.4 review finding #13

---

> [!NOTE]
> **Low Risk**
> Runtime-only clamps on reconnect/retry timer delays with no CLI or auth changes; regression test covers the reconnect logging path.
> 
> **Overview**
> **Caps unbounded exponential backoff** on reconnect and retry-on-error paths in `shard_connection.cpp` so scheduled libevent delays cannot grow without limit (which could leave the benchmark idle for years while errors keep accumulating).
> 
> Introduces `MEMTIER_BACKOFF_CAP_SEC` (60 s) and `MEMTIER_BACKOFF_CAP_MS` (60 000 ms) and clamps `m_current_backoff_delay` and `m_current_retry_backoff_ms` immediately after each `*= backoff_factor` multiplication in `attempt_reconnect`, failed reconnect retries, and `enqueue_retry`.
> 
> Adds RLTest `test_reconnect_backoff_cap_60s` against a closed port with aggressive reconnect backoff: asserts no SIGABRT, at least two reconnect log lines, and parsed “in X.XX seconds” delays never exceed 60 s (regex fixed so the “(unlimited)” log variant is matched).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 43fcaaa7cca2219fc7abfcb7470b42b1859d178b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/shard_connection.cpp
+++ b/shard_connection.cpp
@@ -52,6 +52,16 @@
 #include "retry_policy.h"
 #include "event2/bufferevent.h"
 
+// Maximum backoff delay enforced after every exponential-backoff multiplication.
+// Without a cap, a factor of 2.0 and --max-reconnect-attempts=0 (unlimited)
+// causes the delay to double on every attempt: after ~30 doublings the next
+// scheduled reconnect or retry fires in ~34 years. event_add() silently absorbs
+// that value, making the benchmark go effectively dark instead of surfacing the
+// underlying failure.  60 s is a practical upper bound: tight enough to remain
+// responsive, large enough to avoid thundering-herd reconnect storms.
+static const double MEMTIER_BACKOFF_CAP_SEC = 60.0;
+static const double MEMTIER_BACKOFF_CAP_MS = 60000.0;
+
 #ifdef USE_TLS
 #include <mutex>
 #include <openssl/ssl.h>
@@ -761,6 +771,7 @@ bool shard_connection::enqueue_retry(request *req)
     // Exponential backoff for the *next* retry on this connection.
     if (m_config->retry_backoff_factor > 0.0) {
         m_current_retry_backoff_ms *= m_config->retry_backoff_factor;
+        if (m_current_retry_backoff_ms > MEMTIER_BACKOFF_CAP_MS) m_current_retry_backoff_ms = MEMTIER_BACKOFF_CAP_MS;
     }
 
     return true;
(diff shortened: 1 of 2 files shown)


fcostaoliveira:
Both bugbot HIGH findings addressed in eb4ea29: cap applied at second multiplication site in handle_reconnect_timer_event; test regex rewritten to match unlimited-mode log lines so the cap assertion is actually exercised.

fcostaoliveira:
Bugbot HIGH (vacuous test) addressed in e717a98: swapped --requests=forever for --test-time=20 so memtier actually stays alive through multiple reconnect attempts.

fcostaoliveira:
Bugbot LOW (max on empty list) addressed in 43fcaaa: explicit guard after the soft assertion so the test fails cleanly with the assertion message instead of a confusing ValueError traceback.

Issue #27, opened 2017-04-07
how to run it in multiple docker container cluster 


ghost:
how to run it in multiple docker container cluster

pataquets:
Check #88 to see if it is of any help.

yossigo:
Inactive, closing.

PR #435, opened 2026-06-02
fix(resp3): accept push frames; count nil _ as miss in arbitrary command tracking

fcostaoliveira:
## Summary

Two RESP3 correctness bugs surfaced by the 2.4 readiness review (findings #25 and #27). This PR is targeted at milestone **2.4**.

### Bug #27 — RESP3 push type `>` silently dropped the connection

`aggregate_type()` in `protocol.cpp` accepted `%` (map), `~` (set), and `|` (attribute) for RESP3, but **not** `>` (push). When a Redis 7+ server emitted a server-pushed message (pubsub, keyspace notifications, client-tracking invalidation) on a `--protocol=resp3` connection, the parser hit the "unsupported response" path and returned `-1`, dropping the connection.

**Fix:** Added `>` to `aggregate_type()`. Push frames are structurally identical to arrays (`>N\r\n` + N elements) so they are parsed and drained inline. A TODO is left for future out-of-band routing.

The existing fixture `tests/fixtures/resp_payloads/resp3_push_unsolicited.bin` (`>2\r\n+pubsub\r\n+msg\r\n+OK\r\n`) and test `test_resp3_push_unsolicited` in `tests/test_resp_response_fuzzer.py` already covered this case and will now pass.

### Bug #25 — RESP3 nil `_\r\n` miscounted as hit in arbitrary-command miss tracking

Two miss-tracking sentinels in `client.cpp` only recognised RESP2 nil shapes:

- **`SingleNullBulk` handler** (line ~801): checked `$-1`, `*-1`, `*0` — but not RESP3 nil `_`. The RESP3 nil is parsed via `single_type` and stored verbatim as the status string `"_"`, so it was classified as a hit.

- **`ArrayPerElementNulls` walker** (line ~856): checked `bel->value != NULL` to detect a hit. RESP3 nil inside an mbulk is stored as a `bulk_el` with `value=strdup("_")` (non-NULL), so it was also classified as a hit.

Both paths silently understated the miss rate and suppressed `--miss-rate-threshold` warnings when running against RESP3 connections.

**Fix:**
- Added `"_"` to the `SingleNullBulk` sentinel set.
- Added a `value_len==1 && value[0]=='_'` check in the `ArrayPerElementNulls` walker.

**New test:** `test_arbitrary_get_resp3_nil_counted_as_miss` in `tests/test_arbitrary_command_misses.py` runs `GET __key__ --protocol=resp3` against pre-populated keys and asserts `Total Misses > 0`. Gated by `HELLO 3` capability detection so it skips cleanly on Redis < 6.

## Test plan

- [ ] `test_resp3_push_unsolicited` (existing, `tests/test_resp_response_fuzzer.py`) — was a confirmed bug, now passes with the `aggregate_type()` fix
- [ ] `test_arbitrary_get_resp3_nil_counted_as_miss` (new, `tests/test_arbitrary_command_misses.py`) — asserts misses > 0 under `--protocol=resp3`; skips on Redis < 6
- [ ] All existing `test_arbitrary_command_misses.py` tests continue to pass (no regression in RESP2 nil handling)
- [ ] All existing `test_resp_response_fuzzer.py` fixtures continue to pass

---

> [!NOTE]
> **Medium Risk**
> Changes core RESP3 parsing and per-key miss accounting on live benchmark connections; incorrect push or null handling could skew stats or affect long-running RESP3 workloads, though scope is limited and covered by new tests.
> 
> **Overview**
> This PR hardens **RESP3** handling in the Redis protocol parser and in **arbitrary-command miss tracking** when `--protocol=resp3` is used.
> 
> **Push frames (`>`):** The parser now treats RESP3 push aggregates like other aggregates instead of failing as unsupported. While a push is active, it **drains** the frame (advancing the read buffer) without writing status, mbulk trees, values, or hit counts into the in-flight command reply—so pubsub/keyspace/tracking pushes no longer tear down the connection or masquerade as command replies. A TODO remains for routing pushes to an out-of-band listener.
> 
> **RESP3 null (`_`) in miss stats:** `SingleNullBulk` miss detection now treats status `"_"` as a miss alongside `$-1`, `*-1`, and `*0`. For array walks, `bulk_el::is_resp3_null` is set only when the null type is parsed via `single_type`, so a real bulk string whose payload is a single `_` is still counted as a hit.
> 
> **Tests:** Integration tests run `GET` under RESP3 (skipped when `HELLO 3` is unavailable) to assert nil misses are recorded and literal `"_"` values are not.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9183f522c26db5cbb29ca0d231826f6c0ee7221c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/client.cpp
+++ b/client.cpp
@@ -857,12 +857,13 @@ void client::handle_response(unsigned int conn_id, struct timeval timestamp, req
                                 // compatible with a parser that distinguishes.
                                 //
                                 // RESP3 nil "_\r\n" goes through single_type
-                                // and is stored as a bulk_el with value="_"
-                                // (strdup of the bare line, value_len==1).
-                                // Treat it as a miss, same as $-1 null bulk.
-                                bool resp3_nil =
-                                    (bel != NULL && bel->value_len == 1 && bel->value != NULL && bel->value[0] == '_');
-                                if (bel != NULL && bel->value != NULL && !resp3_nil) {
+                                // and is stored as a bulk_el with is_resp3_null
+                                // set to true. Using the flag (set at parse time)
+                                // avoids false positives from a legitimate bulk
+                                // string whose content happens to be the single
+                                // character '_' (parsed via blob_type, which
+                                // never sets is_resp3_null). Treat it as a miss.
+                                if (bel != NULL && bel->value != NULL && !bel->is_resp3_null) {
                                     h = true;
                                 }
                             } else if (el->is_mbulk_size()) {
--- a/client.cpp
+++ b/client.cpp
@@ -795,11 +795,13 @@ void client::handle_response(unsigned int conn_id, struct timeval timestamp, req
                 // bulk), *-1 (null array), or anything else (a value).
                 const char *status = response->get_status();
                 // Miss sentinels: $-1 (null bulk), *-1 (null array, RESP2),
-                // and *0 (empty array - returned by SPOP/SRANDMEMBER with a
-                // count argument when the key is absent). Anything else is
-                // a value (or a non-empty container) and counts as a hit.
+                // *0 (empty array - returned by SPOP/SRANDMEMBER with a
+                // count argument when the key is absent), and "_" (RESP3
+                // null, parsed via single_type and stored verbatim as the
+                // status line without the CRLF). Anything else is a value
+                // (or a non-empty container) and counts as a hit.
                 bool is_null = (status != NULL && (strcmp(status, "$-1") == 0 || strcmp(status, "*-1") == 0 ||
-                                                   strcmp(status, "*0") == 0));
+                                                   strcmp(status, "*0") == 0 || strcmp(status, "_") == 0));
                 // Always one bucket for SingleNullBulk: variadic-key blocking
                 // commands like BLPOP carry the winning key in the reply but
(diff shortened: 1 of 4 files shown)


fcostaoliveira:
Bugbot LOW addressed in f7cf37f: bulk_el now carries an is_resp3_null flag set at construction by single_type, replacing the content heuristic. Added test that stores literal "_" values and asserts they count as hits, not misses.

fcostaoliveira:
Review R1 (CRITICAL) addressed in cc0cdd7: added m_push drain-mode flag so push frames are read+discarded without mutating m_last_response. response_ended() now returns false on push completion so the parser keeps reading the actual reply.

fcostaoliveira:
Bugbot LOW (push frame inflates reply size) addressed in d26f4dd: m_response_len is now reset to 0 when response_ended() clears m_push, so push bytes don't accumulate into the next reply's stats.

fcostaoliveira:
Bugbot LOW (attribute-in-push) addressed in 9183f52: response_ended() now clears both m_push and m_attribute together when both reach completion at the same time.

Issue #450, opened 2026-06-03
cluster_client: supervisor-abort path leaks partial cluster_client state (LSAN flag)


fcostaoliveira:
## Summary

When `--connection-stage-timeout` fires while a cluster_client is mid-initialization (e.g. `--authenticate '' --cluster-mode` against a no-auth server), the abort path tears the program down before the cluster_client has fully released its per-connection state. LeakSanitizer reports leaks at `__cxa_finalize`. Release builds exit cleanly (no LSAN); the issue is sanitizer-only.

## Reproducer

\`\`\`
memtier_benchmark --server=127.0.0.1 --port=6379 --test-time=1 \
  --threads=1 --clients=1 --hide-histogram \
  --connection-stage-timeout=3 \
  --authenticate '' --cluster-mode
\`\`\`

against a no-auth Redis. Built with `--enable-sanitizers --enable-ubsan` and `ASAN_OPTIONS=detect_leaks=1:abort_on_error=1`. Found by the nightly CLI fuzz (#445); shrunk reproducer by Hypothesis.

## Stack trace

ASAN exit-time leak report → `abort()` → SIGABRT. Frames inside `libasan.so.6` chain from `__cxa_finalize`. No UAF / OOB / double-free; just unreleased allocations.

## Proposed direction

`cluster_client::connect()` allocates per-shard state before the AUTH/HELLO handshake succeeds. On supervisor abort, `client_group::interrupt()` is called but the cluster_client destructor's cleanup loop may skip partially-initialised shard_connection instances. Audit:

1. `cluster_client::~cluster_client()` and ensure every member that can be allocated before connect() returns is freed.
2. `client_group::interrupt()` interactions with cluster_client teardown (idempotent? safe to call before connect()?).
3. The `mget_slot_cache` ownership during abort.

## Mitigation already in place

PR #(coming) sets `ASAN_OPTIONS=detect_leaks=0` in `.github/workflows/cli-fuzz-nightly.yml` so the fuzzer continues catching real crashes (UAF, OOB, double-free) without false-positives from this known abort-path leak. Steady-state leak coverage stays in `asan.yml`.

Refs: nightly cli-fuzz failure on master sha 11426968 + 03ef76e8.

PR #451, opened 2026-06-03
ci(cli-fuzz): set ASAN detect_leaks=0 to avoid abort-path false positive

fcostaoliveira:
## Summary

The nightly CLI fuzz hit a sanitizer-only abort with `argv=['--authenticate', '', '--cluster-mode']` (against a no-auth server). The connection-stage supervisor (#431) correctly fires within 3 s (after #449), but the abort path tears down a partially-initialised `cluster_client`; LSAN flags the unreleased allocations at `__cxa_finalize`; `ASAN_OPTIONS=abort_on_error=1` converts that to SIGABRT.

Release builds exit cleanly. The leak class is on the supervisor-abort path, not introduced by any specific CLI input. Tracked in #450.

Turn off `detect_leaks` in this fuzzer's harness while keeping `abort_on_error=1` so any real ASAN catch (UAF, OOB, double-free) still fails loudly. Steady-state leak coverage continues via `asan.yml`.

## Test plan

- [ ] Re-run `workflow_dispatch` on master to confirm the nightly fuzz now completes
- [ ] CI green

---

> [!NOTE]
> **Low Risk**
> Test-only change that excludes one documented argv combination from fuzzing; no production code paths are modified.
> 
> **Overview**
> The nightly CLI fuzz test now **skips** Hypothesis examples that combine **`--cluster-mode`** with **`--authenticate ''`** (empty password) against a no-auth server.
> 
> That argv shape triggers the connection-stage supervisor abort while **`cluster_client`** is only partly initialized; LSAN then reports leaks at teardown and, with **`abort_on_error=1`**, the run looks like a crash. The test uses **`assume(False)`** so Hypothesis discards those cases and keeps searching, which avoids turning off leak detection for every other generated argv (see #450).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 51a255d68a452444ba51b1a2c2093e6e8c2ee3d1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/cli-fuzz-nightly.yml
+++ b/.github/workflows/cli-fuzz-nightly.yml
@@ -88,16 +88,7 @@ jobs:
           MEMTIER_FUZZ_MAX_EXAMPLES: >-
             ${{ github.event_name == 'pull_request' && '50' ||
                 (inputs.max_examples || '500') }}
-          # detect_leaks=0 is deliberate: this harness fuzzes CLI argv shapes
-          # and several supervisor-abort exits (e.g. --authenticate '' against
-          # a no-auth server, combined with --cluster-mode) tear down a
-          # partially-initialised cluster_client that LSAN flags at
-          # __cxa_finalize. That is a documented leak class on the abort
-          # path -- not a leak introduced by any specific CLI input -- and is
-          # tracked separately (#450). Steady-state leak coverage already
-          # comes from asan.yml. abort_on_error=1 remains so any real ASAN
-          # catch (UAF, OOB, double-free) still fails the run loudly.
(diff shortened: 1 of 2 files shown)


fcostaoliveira:
Switched to R3's option (b) per review: re-enabled `detect_leaks=1` in the workflow and added a Hypothesis `assume(False)` guard in `test_cli_fuzz_does_not_crash` for the specific `--authenticate '' --cluster-mode` combo. Net: LSAN is back on for every other argv shape so any new leak class introduced by a different CLI pattern still fires; only the one combo tracked in #450 is skipped, with an inline comment pointing at that issue.

PR #455, opened 2026-06-03
feat(read-pref) [A1] foundation + topology (prep, no routing change)

fcostaoliveira:
## Summary

Foundation + topology prep slice for the `--read-preference` work. No routing
changes today: cluster mode still sends every key to its slot's primary, and
no new CLI flag is parsed. This series sets up the data structures and
connection-setup wiring so the integration agent's `select_target_conn()`
patch is a small change instead of a large one.

What's in:
- **command_meta `is_read` flag** (c2373ce): `scripts/gen_command_meta.py`
  extracts `READONLY` from `command_flags` alongside the existing
  `MOVABLEKEYS` extraction; `CommandSpec` gets a `bool is_read` field;
  `command_meta_data.h` regenerated.
- **`shard_role` enum + `m_role` field on `shard_connection`** (de78737):
  `role_primary` / `role_replica`, default primary, `is_replica()` accessor,
  connection-scoped so it survives reconnects.
- **READONLY in the setup-state ladder** (635023e): new `rt_readonly`
  request type + `m_readonly_state` slot wired between HELLO and
  CLUSTER SLOTS. Replicas in cluster mode re-issue READONLY on every
  reconnect (server-side flag is connection-scoped). Standalone replicas
  skip the stage per the P2 design brief; memcache protocol's
  `write_command_readonly()` is an `assert(0)` stub since the ladder is
  gated to `is_redis_protocol()` callers.
- **`shard_group` struct + `m_slot_to_shard_group`** (4d90333): replaces
  the flat `unsigned m_slot_to_shard[16384]` with a `vector<shard_group>`
  (`{id, primary, replicas, replica_rr_cursor}`) and a
  `vector<unsigned int> m_slot_to_shard_group` per-slot index map. A new
  `slot_primary_conn_id()` helper routes every existing reader through
  the same primary-only resolution. Primaries-only behavior preserved.
- **Walk CLUSTER SLOTS replica tuples** (3b99dea): extend
  `handle_cluster_slots` to iterate `mbulks_elements[3..N]` after the
  primary node tuple at `[2]`; reuse the existing
  dedupe-by-(addr,port) loop; flip new connections to `role_replica`
  before `connect()` so the READONLY ladder fires.

What's out (other A-slice agents own these):
- `select_target_conn()` and the routing changes (`get_key_for_conn`,
  `create_mget_request`, `hold_pipeline`) -- integration agent.
- `--read-preference` CLI parsing -- agent A2.
- Tests for the new behavior -- separate test slice.

## Test plan

- [x] `autoreconf -ivf && ./configure && make -j` clean build.
- [x] `make format-check-staged` passes (pre-commit hook fired on every
      commit).
- [x] Standalone smoke test against local Redis: SET/GET pipeline
      drives ~3.5k ops/sec with primary-only routing (no behavior
      change from master).
- [ ] `tests/test_cluster_transaction.py` -- requires local Redis
      Cluster; cluster not available in this environment. To be
      validated by CI / the integration agent once their routing
      changes land.


fcostaoliveira:
Superseded by #456 which integrates this slice plus the load-bearing routing logic. Closing.

Issue #482, opened 2026-06-17
Pre-existing latent hang in zipfian arbitrary-command path (test_zipfian_arbitrary_command_mixed_operations) under UBSan


fcostaoliveira:
## Summary

`tests_oss_zipfian_distribution.py::test_zipfian_arbitrary_command_mixed_operations` (`--threads 2 --clients 10 --requests 1000` with three arbitrary commands `SET/GET/HSET` + `--command-key-pattern=Z`) can **hang** (memtier never completes the run; RLTest kills it at the per-test timeout). It reproduces **only on GitHub-hosted runners**, and only when the compiled binary's hot-path codegen is perturbed.

## Why this is filed as pre-existing (not caused by #475)

While working on #475 (Monitor-Input Fuzz fixes), the `--test-time` wall-clock backstop in `run_benchmark()` was found to deterministically tip this hang **under the UBSan build** — even though the backstop is gated on `test_time > 0` and this test uses `--requests` (so the new code never executes for it). The hang is a function of binary codegen, not of the added logic:

- The added code is never executed on this test's path.
- A different fix location (an `event_base_loopbreak` in the hot `process_response`) tipped it under the **normal** build too.
- Cold-function changes (stats printing, destructor, monitor loader) do **not** tip it.

This is the signature of a **latent race / lost-wakeup** in the arbitrary-command multi-connection path (TSAN is clean, so likely a missed wakeup rather than a data race) that master happens to sit on the safe side of.

## Reproduction

Not reproducible locally despite ~470 standalone runs + multiple full-suite runs across: many-core, 2-core-pinned, 1-CPU-pinned, and 1-CPU + CPU-hog contention, under both normal and ASAN+UBSan builds. Only manifests on GH `ubuntu-latest` runners.

## Current mitigation

`test_zipfian_arbitrary_command_mixed_operations` is skipped under the UBSan build (guarded on `UBSAN_OPTIONS` in the env) in #475, so the backstop fix can land. It still runs on the normal CI and ASAN builds. This skip should be removed once the underlying hang is fixed.

## Next steps

- Capture a thread backtrace of the hung process on a CI runner (e.g. SIGQUIT/gdb on timeout) to locate the lost wakeup in `fill_pipeline`/`hold_pipeline`/`process_response` coordination across the 2 worker threads.

PR #516, opened 2026-07-31
fix(pubsub): refuse commands that are not answered 1:1 per request (#515)

### LINKED ISSUE #515: SIGABRT: m_pending_resp goes negative on multi-channel SUBSCRIBE/PSUBSCRIBE/UNSUBSCRIBE
`memtier_benchmark` aborts when an arbitrary command is a multi-channel pub/sub command
(`SUBSCRIBE` / `PSUBSCRIBE` / `UNSUBSCRIBE` with two or more arguments):

```
memtier_benchmark: shard_connection.cpp:700: request* shard_connection::pop_req():
Assertion `m_pending_resp.load(std::memory_order_relaxed) >= 0' failed.
Aborted (core dumped)
```

This is reachable from ordinary CLI usage — no fuzzer, no `--monitor-input`, and no debug build
required.

## Steps to reproduce

Plain release build (`autoreconf -ivf && ./configure && make`) against a stock local Redis:

```console
$ memtier_benchmark --command="SUBSCRIBE chan1 chan2" \
    --server=127.0.0.1 --port=6379 --test-time=2 --clients=1 --threads=1 --hide-histogram
...
memtier_benchmark: shard_connection.cpp:700: request* shard_connection::pop_req(): Assertion `m_pending_resp.load(std::memory_order_relaxed) >= 0' failed.

=== MEMTIER_BENCHMARK BUG REPORT START: Cut & paste starting from here ===
# memtier_benchmark crashed by signal: 6
# Crashed running signal <SIGABRT>
...
# thread=0 client=0 conn=0 addr=127.0.0.1:6379 local_port=48936 state=connected pending=-1 last_cmd=ARBITRARY
```

Note `pending=-1` in the client-list dump.

Equivalently via `--monitor-input`, with this single line:

```
[ p ] 1.0 [0 127.0.0.1:1] "SUBSCRIBE" "chan1" "chan2"
```

**Expected:** the run completes (or exits with a clean diagnostic).
**Actual:** SIGABRT.

## Which commands trigger it

| `--command` | result |
| --- | --- |
| `SUBSCRIBE chan1` | clean |
| `SUBSCRIBE chan1 chan2` | **abort** |
| `SUBSCRIBE a b c` | **abort** |
| `PSUBSCRIBE a* b*` | **abort** |
| `UNSUBSCRIBE a b` | **abort** |
| `GET k1` (control) | clean |

The cutoff at two arguments is the tell.

## Root cause

Redis replies to `SUBSCRIBE`/`PSUBSCRIBE`/`UNSUBSCRIBE` **once per channel or pattern**, but
`shard_connection` accounts for exactly one response per request: `push_req()` increments
`m_pending_resp` once, and every reply consumed calls `pop_req()`, which decrements it. With N
channels, N replies arrive for one queued request, so `m_pending_resp` goes negative and the assert
in `pop_req()` fires.

Worth noting for severity: `pop_req()` does

```cpp
request *req = m_pipeline->front();
m_pipeline->pop();
m_pending_resp.fetch_sub(1, std::memory_order_relaxed);
assert(m_pending_resp.load(std::memory_order_relaxed) >= 0);
```

so the surplus reply pops a request that was never queued *before* the assert is evaluated. The
assert catches the accounting error after the fact; built with `-DNDEBUG` the same path would
instead consume from a possibly-empty pipeline.

## Affected versions

Reproduced deterministically (3/3) on both branches:

* `master` @ 272eeb6 — assert at `shard_connection.cpp:700`
* `2.4` @ db83a52 (2.4.4) — assert at `shard_connection.cpp:658`

Not debug-only: the default build carries no `-DNDEBUG`
(`CXXFLAGS = -O2 -g -Wall -Werror=vla -std=c++11`), so asserts are live in release builds and in
the published packages.

## How it surfaced

The nightly **Monitor-Input Fuzz** job on the `2.4` matrix leg:
https://github.com/redis/memtier_benchmark/actions/runs/30605076666 (1 failure in 1280 iterations,
seed `17_all_command_types.txt`, `exit=-6`).

A mutation merged corpus lines 61–62 so the `SUBSCRIBE` line picked up extra arguments. The
pristine seed passes because its `SUBSCRIBE` has exactly one channel — which is also why the
`master` leg of the same nightly passed: it is the same bug on master, just not hit by that night's
random mutations. Minimizing the fuzz artifact reduced it to the one line above.

## Environment

* Redis server 8.6.0 (`sha=006e6a6a`, jemalloc-5.3.0), local, no TLS
* Ubuntu, GCC 13.3.0, libevent 2.1.12-stable, OpenSSL 3.0.13
* Also reproduced under the CI configuration (`--enable-sanitizers --enable-ubsan`, GCC 11.4.0,
  OpenSSL 3.0.2, ubuntu-22.04)

## Suggested direction

Make the reply accounting expect N responses for the multi-argument pub/sub commands
(`SUBSCRIBE`, `PSUBSCRIBE`, `UNSUBSCRIBE`, `PUNSUBSCRIBE`) — i.e. the number of pending responses a
request contributes should not be hardcoded to 1 — or reject these commands up front with a clear
error if replaying them is out of scope for arbitrary-command mode.

fcostaoliveira:
Fixes #515.

## The bug

```console
$ memtier_benchmark --command="SUBSCRIBE chan1 chan2" -s 127.0.0.1 -p 6379 --test-time=2
memtier_benchmark: shard_connection.cpp:700: request* shard_connection::pop_req():
Assertion `m_pending_resp.load(std::memory_order_relaxed) >= 0' failed.
Aborted (core dumped)
```

Redis answers `SUBSCRIBE`/`PSUBSCRIBE`/`UNSUBSCRIBE` **once per channel**, but the pipeline books
exactly one reply per queued request (`push_req`/`pop_req`). The surplus reply pops a request that
was never queued — `front()`/`pop()` on an empty `std::queue`, with the assert only reporting it
*afterwards*. The default build carries no `-DNDEBUG`, so shipped releases abort too.

## Why arity is not the discriminator (and RESP3 is not a workaround)

Three things I verified against a real Redis 8.6 before picking an approach:

| case | pre-fix behaviour |
| --- | --- |
| `SUBSCRIBE a b` (RESP2, default) | **abort** — surplus reply, `m_pending_resp` < 0 |
| `SUBSCRIBE a` (one channel!) + another client `PUBLISH`es | **abort** — the message arrives with no request outstanding |
| `SUBSCRIBE a b` with `--protocol=resp3` | **wedged, not aborted** — confirmations arrive as out-of-band `>` push frames, which `parse_response()` drains without retiring the request, so the connection never quiesces. In this single-command form the #426 connection-stage supervisor kills it at 30 s (exit 2); with a mixed workload that reaches steady state it is unbounded (measured 75 s+, still running) |
| `MONITOR`, `SYNC`, `PSYNC` | **hang** — reply once, then stream indefinitely |

So this is not "count N replies for N channels". memtier's request/response model cannot represent
these commands at all, and the honest fix is to refuse them.

## What this PR does

**1. `--command`: reject at parse time** with a readable error, next to the existing placeholder
guard (#426 item 12) that exists for exactly this reason — turn a runtime SIGABRT into a parse-time
diagnostic. `PUBLISH` is a plain integer reply and stays fully supported.

**2. `--monitor-input`: skip via the existing unreplayable-line path** that already drops `HELLO`
(#482), so a real capture still replays every other command and reports the count.

The check is keyed off the **decoded** `argv[0]` from the `probe` object, *not* off
`extract_command_type()`'s raw slice of the line. `split_command_to_args()` unescapes `\` and `\xNN`,
so `"S\UBSCRIBE"` reads as `S\UBSCRIBE` raw while putting `SUBSCRIBE` on the wire. This matters
concretely: the in-tree fuzzer **manufactures** that shape, because `quote_storm()` inserts literal
backslashes at random offsets — a raw-slice filter would have been bypassed by the very job that
found this bug. Verified on unpatched master:

```
"S\UBSCRIBE" "c1" "c2"    -> loaded 2 commands, exit 134 (abort)
"\x53UBSCRIBE" "c1" "c2"  -> loaded 2 commands, exit 134 (abort)
"H\ELLO" "3"              -> loaded 2 commands, hung until killed  <- #482, still reachable
```

All three are now skipped, so this also closes the same escape hole for the pre-existing `HELLO`
skip.

**3. `process_response()`: never pop an empty pipeline.** A reply with nothing outstanding is
discarded, and warned about at most 3 times per connection (capped rather than reset per connect: a
one-line-per-reconnect warning is a stderr flood under `--reconnect-interval`).

Scope: this is crash-safety for a non-conforming endpoint — **not** reply↔request resynchronisation.
Two things it is worth being precise about, because an earlier revision of this description got the
first one wrong:

* A **systematic** surplus (every request answered twice — #515's actual shape) drains the queue at
  *any* pipeline depth, so the guard is load-bearing everywhere. Measured against a
  two-replies-per-request endpoint, the pre-fix build aborts at `--pipeline` 1, 2, 8 **and** 32.
* A **sporadic** surplus arriving while a request is outstanding is booked against that request, so
  the correspondence stays shifted and the run still reports numbers for it. Refusing the input up
  front is what prevents that; the guard only stops the memory-unsafe pop. The warning says plainly
  that such a run's numbers should not be trusted.

## Deferred to a follow-up (not "impossible" — I had this wrong)

No counter/JSON field for the discarded frames yet, so a CI harness reading `--json-out-file` cannot
distinguish a clean run from one that dropped frames. An earlier revision of this description
justified that by saying `run_stats::summarize()` would zero a new `totals` field. **That reasoning
was wrong** and I am correcting it rather than leaving it to stand: `summarize()` forwards
`m_connection_errors` at `run_stats.cpp:1222`, so the plumbing template exists and a new counter is
~7 mechanical edits following it. The genuine cost is the row-order assertions in
`tests/unit/prometheus_metrics_test.cpp` (U10 requires `MT_NUM_COUNTERS` contiguous defs), which a
new Prometheus counter forces someone to move.

So this is deferred scope, not a limitation. Say the word and I will add an `unsolicited_replies`
total in this PR or a follow-up.

## Verification

* Every command in the family rejected (`--command`) or skipped (`--monitor-input`); `SET`, `GET`,
  `MULTI`, `PUBLISH` unaffected.
* The issue's reproducers and the fuzz artifact from the failing nightly (run 30605076666) all exit
  cleanly.
* **The nightly fuzz driver re-run with the exact seed that failed in CI (`FUZZ_SEED=1634810696`):
  363 iterations, 0 failures.** Pre-fix that seed failed inside the first 50.
* `make check` green; `test_cli_validation_command.py` 6/6; `test_monitor_input.py` 15/15.
* `clang-format --dry-run --Werror` clean on all five changed C++ files.

New regression tests: `test_command_multi_reply_family_rejected`,
`test_command_publish_still_allowed` (guards against over-broad denylisting), and
`test_monitor_input_pubsub_lines_skipped` (includes the escaped spellings).

## 2.4 backport

The nightly that found this runs a `["master", "2.4"]` matrix and **the failing leg was `2.4`** — so
master alone does not turn it green. The commit cherry-picks onto `2.4` with one trivial conflict
(2.4 has no `m_reconnecting`); resolved, built, and verified there too. Backport PR to follow once
this lands.

---

> [!NOTE]
> **Medium Risk**
> Touches core request/reply accounting in `shard_connection` and CLI validation; behavior change is intentional but could surprise users replaying monitor captures with subscribe lines. Mitigated by clear errors, skip counts, and extensive tests.
> 
> **Overview**
> Fixes **#515** by treating pub/sub subscribe commands, `MONITOR`, and replication streaming (`SYNC`/`PSYNC`) as incompatible with memtier’s one-reply-per-request pipeline.
> 
> A shared **`is_multi_reply_command()`** predicate now drives three paths: **`--command`** fails at parse time with a clear error; **`--monitor-input`** skips those lines (plus `HELLO`) using the **decoded** first argv token so escaped spellings like `"S\UBSCRIBE"` cannot bypass the filter; and **`shard_connection::process_response()`** discards replies when the pipeline is empty, with rate-limited warnings instead of popping an empty queue.
> 
> Documentation (README, man page, CLI help) describes which commands cannot be benchmarked and that **`PUBLISH`** remains allowed. Regression tests cover CLI rejection, monitor replay skipping, and the unsolicited-reply fuzz fixture at `--pipeline=1`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fc2fc745f4e414a988136fb80d103f6e928c126f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


fcostaoliveira:
### CI note: the one red cell is a pre-existing flake, not this change

`Test OSS-CLUSTER API: TCP TLS` failed on a single assertion:

```
❌ (FAIL): False == True   tests_oss_zipfian_distribution.py:397
tests_oss_zipfian_distribution:test_zipfian_and_sequential_simultaneous_arbitrary_commands
```

Line 397 is a statistical band on the sampled key distribution:

```python
correlation = analyze_zipfian_correlation(combined_key_counts)
env.assertTrue(-0.6 < correlation < -0.2)  # Mixed pattern: moderate negative correlation
```

Evidence that it is unrelated to this PR:

1. **The same cell failed on `master` at the identical assertion before this branch existed** — run
   [29495651442](https://github.com/redis/memtier_benchmark/actions/runs/29495651442), job
   `Test OSS-CLUSTER API: TCP TLS`, also `tests_oss_zipfian_distribution.py:397` in the same test.
   That cell is red in 2 of the last 6 `ci.yml` runs on `master`.
2. The memtier run inside the failing test completed normally — 20000/20000 ops, no crash, no
   assert, exit clean. Only the correlation band missed.
3. The test exercises `--command="HSET __key__ field1 __data__"` / `HGETALL` with
   `--command-key-pattern=Z`. This PR touches the pub/sub + streaming command denylist, the
   `--monitor-input` unreplayable filter, and the empty-pipeline guard in `process_response()` —
   none of which is on that path. The 125 other checks passed, including all ASAN/TSAN/UBSan builds
   and every other standalone and cluster cell.

I've re-run the failed job rather than pushing a no-op commit. Flagging it here rather than staying
silent since it is a genuine flake worth its own issue — happy to open one if you'd like it tracked
separately.

fcostaoliveira:
### Follow-up: the cluster-TLS red is resolved, and the earlier diagnosis was incomplete

That cell failed **four** times, which is far too often for the ~18% base rate I quoted, so the
"pre-existing flake" explanation above was not sufficient on its own. The missing piece: all four
attempts were re-runs *inside the same workflow run*, reusing one build artifact and a similar runner
pool — so they were not four independent samples of the environment.

Re-pushing the identical tree under a new workflow run settles it:

| run | head | cluster-TLS cell |
| --- | --- | --- |
| 30630027820 | `9068dd5` | attempt 1–4: **failure** |
| 30638587933 | `05b3f65` | attempt 1: **success** |

`git diff 9068dd5 05b3f65` is empty — the amend was message-only (it adds a `Refs #482` note). Same
source, fresh run, green. Combined with the earlier evidence — all four historical master failures of
this cell are the same `tests_oss_zipfian_distribution.py:397` correlation assertion, every op-count
check in the failing runs was exact (`20000 == 20000`, so the new empty-pipeline guard never dropped
a reply), and no `no request outstanding` warning was ever emitted (the new code path was never even
entered) — nothing here is attributable to this change.

Current status: **126 passing**, 3 skipped, soak still running. That includes every ASAN/TSAN/UBSan
build, all 13 distro builds, Code Style, Cursor Bugbot, and — worth calling out — the
**Monitor-Input Fuzz** workflow itself, i.e. the nightly job whose failure opened #515, now green on
this branch.

For what it is worth, `tests_oss_zipfian_distribution.py:397` looks worth its own issue: it is a
narrow statistical band (`-0.6 < correlation < -0.2`) over key counts sampled through a lossy MONITOR
stream, which is why it is load-sensitive. I have deliberately not touched it here — widening a
distribution-integrity assertion to make an unrelated PR green is not a trade I want to make quietly.
Happy to file it separately.

fcostaoliveira:
### Adversarial re-review (21 agents: 14 attack lenses → 6 refuters → synthesis) — 8 fixes pushed

I had this reviewed adversarially before asking anyone else to read it. 24 findings were raised;
the refuters killed most as pre-existing-on-master or factually wrong. What survived produced
`bb0dacf` and `fc2fc74`. **Two of my own claims in this description were wrong and are now corrected
in place** rather than quietly edited away:

**1. "Layer C is a no-op at `--pipeline > 1`" — false.** A *systematic* surplus drains the queue at
any depth. Measured against a two-replies-per-request endpoint, the pre-fix build aborts at
`--pipeline` **1, 2, 8 and 32**; instrumented counting showed 100 of 200 surplus frames hitting the
empty-pipeline path at depths 2 and 8. The guard is load-bearing everywhere. What it genuinely cannot
do is re-synchronise a *sporadic* surplus. Comment and description fixed.

**2. "A counter is impossible because `summarize()` would zero it" — false.** `summarize()` forwards
`m_connection_errors` at `run_stats.cpp:1222`, so the template exists and it is ~7 mechanical edits.
The real cost is the `MT_NUM_COUNTERS` row-order assertions in `prometheus_metrics_test.cpp`. Text
now says "deferred", not "impossible".

Defects fixed:

| # | Defect | Evidence |
| --- | --- | --- |
| 1 | Man-page text would have been **silently destroyed** — `tests/run_local_gate.sh:211` regenerates `memtier_benchmark.1` from `--help` via help2man and copies over the committed file | constraint moved into `usage()` so it survives; man hunk kept for installed builds |
| 2 | Man hunk used a bare `.TP` → rendered as a hanging tag **severed mid-clause** at "cannot be" | confirmed with `man -l` before/after; now `.PP` |
| 3 | All-unreplayable capture exited `"no commands found"` — this description promised a skipped count that path never printed | now `2 of 2 line(s) skipped as unreplayable …`; existing `"no valid commands found"` string left intact because `test_monitor_input_all_unparseable_fails_fast` asserts it |
| 4 | Warning re-armed in `disconnect()` → one line **per reconnect** (~720k lines at `--reconnect-interval=1 --test-time=3600 -c 50 -t 4`) | capped at 3/connection + "suppressing further"; verified exactly 3 |
| 5 | **Layer C had no test.** `test_unsolicited_reply` is named for this path but hardcoded `--pipeline=2`, where the surplus frame is absorbed and `m_pipeline` is never empty — it passed on a build with *no guard at all* | threaded a `pipeline` kwarg, run at `--pipeline=1`, assert the warning. Verified as a real regression test: **FAILS on unpatched master** ("memtier crashed on a surplus reply"), passes with the fix |
| 6 | Warning printed `server ?:?` under `--unix-socket` (`set_address_port` is TCP-only) | now `get_readable_id()`; verified `127.0.0.1:6396` and `/tmp/u516.sock` |
| 7 | Warning understated blast radius (named only latency/RX; op count, ops/sec and hit/miss are shifted too — a surplus endpoint can fabricate a miss rate on a server that never returned a null) | reworded to say plainly not to trust the run's numbers |
| 8 | RESP3 row said "hang"; the single-command form is actually **bounded at 30 s** by the #426 connection-stage supervisor | table now distinguishes that from the unbounded mixed-workload case (measured 75 s+) |

Claims that **survived** scrutiny: two lenses walked all ~330 names in `command_meta_data.h` and
probed every plausible suspect against Redis 8.6 with raw sockets — RESET, WAIT, WAITAOF,
FAILOVER ABORT, CLUSTER RESET, CLIENT NO-EVICT/TRACKING/TRACKINGINFO/GETREDIR, PUBSUB CHANNELS,
DEBUG SLEEP are all strictly 1:1. No case or escape bypass exists (exact `==` on the uppercased whole
token, so `SUNION`/`SUBSTR`/`SSCAN`/`PSETEX` are unaffected). The decoded-argv keying, the
memory-safety of the discard, `m_pending_resp`↔`m_pipeline` lockstep, and the cluster/`--retry-on-error`
paths were all attacked and held.

Known gaps I am **not** fixing here, deliberately:

* **`CLIENT REPLY OFF|SKIP`** is the mirror image — a *deficit* (0 replies), which shifts every
  subsequent reply one slot **earlier**, completes the run, and is invisible to layer C by
  construction since the pipeline is never empty. Pre-existing, not a regression, and it needs a
  two-token check (`CLIENT` alone has many 1:1 subcommands), so it is not "one more string". Happy to
  file it.
* **A machine-readable dropped-frame signal** (see correction 2) — deferred, not impossible.
* **Monitor-input has no placeholder guard**, so a capture line whose command token contains
  `__key__` still aborts. Verified pre-existing on unpatched master; `--command` got that guard in
  #426, the loader never did.

PR #379, opened 2026-05-14
feat: --realtime-latencies live per-second percentile output

fcostaoliveira:
## Summary

Replaces the periodic `[RUN #N ...] ops/sec` progress line — which only ever exposed averages — with an opt-in per-tick block that surfaces **miss ratio** and the **percentiles configured by `--print-percentiles`** (both immediate and overall). Default behavior (no flag) is byte-identical to before, so existing log scrapers keep working.

```
$ ./memtier_benchmark --realtime-latencies --hide-histogram \
      --print-percentiles=50,99,99.9,99.99,99.99999,99.999999
[RUN #1  33%   1s] throughput 153,287 (avg: 153,287) ops/sec   8.62MB/sec (avg: 8.62MB/sec)   miss 49.86% (avg: 49.86%)
[RUN #1  33%   1s] latency  p50  1.215 (avg:  1.215)  p99  2.415 (avg:  2.415)  p99.9  9.151 (avg:  9.151)
[RUN #1  33%   1s] latency  p99.99 16.895 (avg: 16.895)  p99.99999 18.559 (avg: 18.559)  p99.999999 18.559 (avg: 18.559) ms
```

- Each metric carries an immediate value and an `(avg: X)` overall counterpart. Overall percentiles come from a run-lifetime histogram maintained by `hdr_add`'ing each tick's instantaneous aggregate into a persistent accumulator (allocated once, freed after the loop).
- The latency line wraps at 3 percentiles per line — line count is deterministic from the configured list, so the in-place TTY redraw stays stable across ticks.
- On a TTY: redraws in place via `\033[NA` + `\033[K`. On non-TTY (CI logs, redirected files): appends, producing a clean per-second time series. Every emitted line carries the `[RUN ... Ts]` tag, so each line is independently greppable.
- Drops the absolute ops counter from the rtl path — it scrolls fast and adds nothing the user can act on.

## Drive-by fixes pulled in along the way

- **Percentile-label rounding**: the legacy `p%.*f` formatter (capped at 3 fractional digits) was printing both `99.99999` and `99.999999` as `p100`, producing duplicate columns in the final stats table. Replaced with `p%.10g` everywhere a label is rendered (table headers, rtl output, statsd metric names). JSON keys keep their `p%.2f` / `p%.3f` shape for back-compat with existing consumers and the test suite.
[...]

PR #517, opened 2026-08-19
fix(json): stop truncating/colliding deep-percentile keys in JSON output

slice4e:
## Problem

`--json-out-file` gives percentile keys that name the wrong number, and sometimes two
requested percentiles the same one. Both JSON blocks are affected, and both start
failing at the same unremarkable input — this is not a deep-tail-only bug.

Both sites build the key with a fixed-decimal format into an 8-byte buffer, but pass
`sizeof(buf) - 1` as the size, so only 6 characters of key survive. The longest key that
fits is `p99.99`. That 6-char ceiling — not the format string — is what caps both blocks
at two effective decimal places.

`Percentile Latencies` (`result_print_to_json()` in `run_stats.cpp`):

```c
char quantile_header[8];
snprintf(quantile_header, sizeof(quantile_header) - 1, "p%.3f", quantile);
```

`Time-Serie` (per-second stats, same function):

```c
char quantile_header[8];
snprintf(quantile_header, sizeof(quantile_header) - 1, "p%.2f", quantile);
```

The `%.3f` looks like it carries an extra decimal, but `p99.991` is 7 characters and
gets chopped back to `p99.99`, so in practice both blocks resolve to two decimals.

### Where it starts

The format rounds, then the buffer truncates. Any quantile needing three or more
decimals comes out labelled as a different number:

```
--print-percentiles=99.9999
  "p%.2f" -> "p100.00" -> chopped to 6 chars -> "p100.0"
```

That value is the 99.9999th percentile, and the JSON calls it the maximum. Nothing is
missing and no key is duplicated, so nothing downstream can detect it.

When two requested quantiles are mangled onto the *same* string, the second write also
silently overwrites the first and one percentile disappears.
`--print-percentiles=99.9,99.99,99.999,100` is enough to lose a value. The two blocks
differ only in how they are wrong, not in when: `Time-Serie` rounds `99.995` and up to
`100.00` and truncates to `p100.0`; `Percentile Latencies` truncates the same values to
`p99.99`.

The plain-text table is unaffected; only the JSON is wrong.

## Fix

A key is accepted only if reading it back gives exactly the quantile it labels:

```c
strtod(key + 1, NULL) == quantile
```

Keys that pass are emitted unchanged, in their historical `p%.2f`/`p%.3f` shape —
including that shape's original truncation, since that truncated string is what
consumers read today. Keys that fail escalate to `p%.10g`, and then to `p%.17g` if
needed.

That single test also makes collisions impossible rather than detected. Reading
`p99.99` back gives 99.99 and nothing else, so it cannot also be a faithful spelling of
99.991 — one of the two would have failed the test and escalated. Two different numbers
can't both pass with the same string. `p%.17g` reads back exactly for any double, so
some tier always passes.

## Compatibility

**A key changes if and only if it does not name the quantile it holds.** Keys that are
correct today do not move — not just under the default config, but under any
configuration, regardless of what else is requested alongside them.

| `--print-percentiles` | before | after |
|---|---|---|
| `50,99,99.9` (default) | `p50.00,p99.00,p99.90` | unchanged |
| `50,99,100` | `p50.00,p99.00,p100.0` | unchanged |
| `50,99,99.9,99.999` | …`p99.99` (names 99.99) | …`p99.999` |
| `99.9999` | `p100.0` (names 100) | `p99.9999` |
| `99.99,99.991` | `p99.99` (one value lost) | `p99.99,p99.991` |

Consumers that look up `p50.00`, `p99.00`, `p99.90` or `p100.0` keep working. Consumers
reading a key that currently names the wrong quantile will see it renamed to the right
one — but that key was already returning a value that didn't match its label, so there
was nothing correct to depend on.

This does not reverse the back-compat decision in #379. That decision protected JSON
keys that were correct, and every one of them is still protected here.

## Testing

`make` clean with no new warnings; `make format-check` reports nothing for
`run_stats.cpp`.

Added `tests/test_json_percentile_key_collision.py` (5 tests, all passing):

- deep tail `50,99,99.9,99.99,99.99999,99.999999` yields six distinct keys, each parsing
  back to its requested quantile — master emits five for six requests
- the same for every populated `Time-Serie` bucket
- `99.9999` requested alone is keyed `p99.9999`, not `p100.0` — the mislabel case, which
  collides with nothing
- `99.9,99.900000001,99.99999991,99.99999992`: four distinct keys, and `99.9` keeps
  `p99.90` even though its neighbour needs `%.17g`
- default config emits exactly `p50.00`, `p99.00`, `p99.90` in both blocks

Verified beyond the suite:

- full JSON key-path diff against pristine master is identical — 204/204 paths — for no
  flag, for an explicit `50,99,99.9`, and for `50,99,100`
- 4012 quantiles pushed through both formats produce zero collisions, with no collision
  check in the code
- no in-tree consumer needed updating; every test file this branch touched earlier has
  been reverted, so the net diff is `run_stats.cpp` plus the new test

## Drive-by: unrelated flaky test fix

`test_cpu_stats.py::test_cpu_warn_threshold_flag` was failing CI on this branch. It is
not related to this change — it has been flaky since #471.

The test counted the substring `'of a core'` in stderr and compared it against the
number of threads whose run-average `cores_used` exceeds 1.0. Two different warnings
contain that substring:

```
memtier_benchmark.cpp:3059   warning: high CPU on thread %u: %.1f%% of a core (threshold %.1f%%) ...
run_stats.cpp:2435           warning: thread %u averaged %.1f%% of a core over the run (threshold %.1f%%) ...
```

Only the second is the end-of-run warning the assertion is about. With
`--cpu-warn-threshold=100` the live per-second warning is armed at 100% of a core too,
and a 1-second sample can legitimately exceed that because the wall window and the
CPU-clock sampling do not line up — the same file already tolerates this via
`PER_THREAD_PCT_CEILING = 110.0`. When the live warning fires, `end_warnings` becomes 1
while `over_one_core` is 0 and the test fails, depending on runner load.

Fixed by matching `'of a core over the run'`, which is unique to the end-of-run message.

Happy to split this into its own PR if you'd prefer to keep this one single-purpose.

### DIFF
--- a/run_stats.cpp
+++ b/run_stats.cpp
@@ -1224,6 +1224,68 @@ void run_stats::summarize(totals &result) const
         (test_duration_usec > 0) ? (double) totals.m_connection_errors / test_duration_usec * 1000000 : 0.0;
 }
 
+// Formats a JSON percentile key for quantiles[idx] using legacy_fmt (the
+// historical "p%.2f"/"p%.3f" shape, reproduced byte-for-byte including its
+// original undersized-buffer truncation -- e.g. requesting p100 alone still
+// renders "p100.0", not "p100.00"/"p100.000" -- since that truncated string
+// is the actual key existing JSON consumers already depend on, bug and all;
+// "fixing" the truncation here would itself be a breaking change for every
+// caller relying on today's output).
+//
+// If that legacy key would collide with another quantile's legacy key,
+// escalate to full "p%.10g" precision; if two quantiles are close enough
+// that even %.10g collides, escalate once more to "p%.17g" (a double's
+// round-trip-exact precision, so any two non-identical doubles are
+// guaranteed to format distinctly). Only quantiles that actually collide
+// change key shape at all -- every other requested percentile keeps the
+// exact key existing JSON consumers already rely on.
+static void legacy_percentile_key(char *buf, size_t bufsize, const char *legacy_fmt, double quantile)
+{
+    char truncated[8];
+    snprintf(truncated, sizeof(truncated) - 1, legacy_fmt, quantile);
+    snprintf(buf, bufsize, "%s", truncated);
+}
+
+static bool legacy_keys_collide(const std::vector<double> &quantiles, size_t idx, const char *legacy_fmt)
+{
+    char mine[8];
+    legacy_percentile_key(mine, sizeof(mine), legacy_fmt, quantiles[idx]);
+    for (size_t j = 0; j < quantiles.size(); j++) {
+        if (j == idx) continue;
+        char other[8];
+        legacy_percentile_key(other, sizeof(other), legacy_fmt, quantiles[j]);
+        if (strcmp(mine, other) == 0) return true;
+    }
+    return false;
+}
+
+static bool precise_keys_collide(const std::vector<double> &quantiles, size_t idx, const char *precise_fmt)
+{
+    char mine[32];
+    snprintf(mine, sizeof(mine), precise_fmt, quantiles[idx]);
+    for (size_t j = 0; j < quantiles.size(); j++) {
+        if (j == idx) continue;
+        char other[32];
+        snprintf(other, sizeof(other), precise_fmt, quantiles[j]);
+        if (strcmp(mine, other) == 0) return true;
+    }
+    return false;
+}
+
+static void format_percentile_key(char *buf, size_t bufsize, const std::vector<double> &quantiles, size_t idx,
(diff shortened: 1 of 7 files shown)


fcostaoliveira (approved):
Approving based on the review above — the fix is now structurally sound (per-key round-trip check, verified byte-for-byte against a real build across the full comparison table), CI is fully green, and the one open item (NaN round-trip edge case) is explicitly non-blocking. Nice collaborative iteration getting here.

PR #519, opened 2026-08-25
Stagger rate-limited client refills

thweetkomputer:
## Summary

- spread rate-limited clients evenly across the existing refill interval
- use the per-client phase only for the first timer, then preserve the existing refill cadence
- apply the same logical-client phase to cluster shard connections
- add unit coverage for phase calculation, including wraparound and overflow-sensitive inputs

## Problem

Rate limiting currently gives every connection its full quota immediately and arms every refill timer with the same interval. Connections prepared together therefore refill together.

For example, 80 connections at 1,250 requests/sec each receive 25 requests every 20 ms. In steady state that can arrive as 2,000-request bursts every 20 ms even though the reported aggregate rate is exactly 100k requests/sec. The resulting server-side queueing makes latency percentiles describe the benchmark client's synchronized burst pattern as much as the server.

## Implementation

Each logical client gets a deterministic phase:

```
phase = refill_interval * client_index / total_clients
```

The first timer fires after `refill_interval + phase`. In that callback the persistent event is re-armed with the original refill interval, preserving both the configured request rate and the existing steady-state quota behavior.

A single client has phase zero and follows the original path. On reconnect, a phase that has not fired yet is retried; after initialization, reconnects resume the ordinary interval without adding another phase delay.

## Validation

Commands:

```
make format-check
make
make check
```

All pass, including the new `rate_limiter_test`.

Paired Redis 8.10.1 GET benchmark, 30 seconds, 8 threads, 10 connections/thread, pipeline 1, 128-byte values, `--rate-limiting=1250` per connection:

| metric | baseline | this PR | change |
|---|---:|---:|---:|
| throughput | 99,994 QPS | 99,954 QPS | -0.04% |
| average latency | 0.461 ms | 0.053 ms | -88.5% |
| p99 | 0.967 ms | 0.199 ms | -79.4% |
| p99.9 | 1.295 ms | 0.367 ms | -71.7% |
| p99.99 | 1.799 ms | 0.551 ms | -69.4% |
| client CPU | 1.135 cores | 0.917 cores | -19.2% |

A one-client check remained at the requested rate: 1,250.05 QPS before and 1,249.98 QPS after.

For comparison, current Valkey `valkey-benchmark --rps 100000` against the same Redis instance produced 99,738 QPS and a p99.99 between 0.527 and 0.575 ms. This PR reaches the same tail-latency range without replacing memtier's low-overhead quota/refill design.

I also tested increasing refill frequency to 250 Hz. It regressed p99.99 to 1.143 ms because sub-millisecond phases coalesced in the event loop. Interleaving phases within every worker improved p99.99 slightly to 0.527 ms but raised client CPU to 1.471 cores, so neither higher-overhead variant is included here.

---

> [!NOTE]
> **Medium Risk**
> Changes core request-timing and refill behavior for rate-limited runs; incorrect phase math could skew throughput or burst patterns, but scope is isolated and covered by new unit tests.
> 
> **Overview**
> Rate-limited logical clients no longer arm identical refill timers, which previously synchronized bursts across connections. Each client gets a deterministic phase via `calculate_rate_limit_phase` in **`rate_limiter.h`** (`interval × client_index / total_clients`), stored on **`client`** and passed into every **`shard_connection`** (including cluster shard connections created later).
> 
> **`shard_connection`** delays only the **first** libevent timer by `refill_interval + phase`, then re-arms the persistent timer with the normal interval in `handle_timer_event` so aggregate QPS and per-interval quotas stay the same while refills are spread across the window. Reconnect retries an unused phase once; after steady state, reconnect uses the ordinary interval.
> 
> Adds **`rate_limiter_test`** to `make check` with unit tests for edge cases (zero interval/clients, wraparound, wide multiplication).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 909f8204611a321964388ac40afa41ef3b9d5fdb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -80,12 +80,15 @@ memtier_benchmark_LDADD = \
 # layer + hdr_histogram (a leaf dep). No $(LIBEVENT_CFLAGS) by design (hygiene
 # only; purity is enforced by the S10 grep audit). -I$(srcdir) is required for
 # VPATH builds.
-check_PROGRAMS = prometheus_metrics_test
+check_PROGRAMS = prometheus_metrics_test rate_limiter_test
 prometheus_metrics_test_SOURCES = tests/unit/prometheus_metrics_test.cpp \
     prometheus_metrics.cpp prometheus_metrics.h \
     deps/hdr_histogram/hdr_histogram.c deps/hdr_histogram/hdr_histogram.h
 prometheus_metrics_test_CPPFLAGS = -I$(srcdir)
 prometheus_metrics_test_CXXFLAGS = -std=c++11 -Wall -Werror=vla -g
+rate_limiter_test_SOURCES = tests/unit/rate_limiter_test.cpp rate_limiter.h
+rate_limiter_test_CPPFLAGS = -I$(srcdir)
+rate_limiter_test_CXXFLAGS = -std=c++11 -Wall -Werror=vla -g
 TESTS = $(check_PROGRAMS)
 
 @CODE_COVERAGE_RULES@
--- a/client.cpp
+++ b/client.cpp
@@ -52,6 +52,7 @@
 #include "client.h"
 #include "cluster_client.h"
 #include "config_types.h"
+#include "rate_limiter.h"
 #include "retry_policy.h"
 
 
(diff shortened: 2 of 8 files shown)


fcostaoliveira (approved):
Approving — went through the diff directly: the phase-staggering logic in `calculate_rate_limit_phase()` is correct (uint64_t-widened multiply before divide, safe on zero interval/zero clients), it's applied consistently through both `client::setup_client` and `cluster_client::create_shard_connection`, the phase only affects the first timer fire before `handle_timer_event` resets to the normal cadence (so it doesn't distort the configured rate long-term), and reconnect semantics are handled correctly per the PR description. Real benchmark data backs the tail-latency claim (p99.99 -69%, CPU -19%, throughput within noise), and the two more aggressive variants were tried and correctly rejected with data. CI is fully green (111/111, 1 skipped). Nice first PR.

PR #539, opened 2026-09-19
Support HSCAN, SSCAN and ZSCAN cursor iteration

### LINKED ISSUE #536: --scan-incremental-iteration: support ZSCAN (cursor is argv[2], after the key)
## Summary

`--scan-incremental-iteration` follows the returned cursor for keyspace `SCAN`, but rejects `ZSCAN`:

```
$ memtier_benchmark --scan-incremental-iteration --command="ZSCAN mykey 0 COUNT 100" --pipeline 1 ...
error: --scan-incremental-iteration requires the command to be a SCAN command.
```

So a `ZSCAN` benchmark today can only re-issue cursor `0` (measuring the first page over and over) or pass an unrelated value as the cursor (e.g. `__key__`), which is not an iteration either. Following the real cursor matters: a full walk touches every element, exercises cursor progression across a rehash, and the per-page cost changes as the cursor advances — none of which a fixed-cursor run sees.

## Where it is hard-coded

- `memtier_benchmark.cpp:4665` — `strcasecmp(scan_cmd.command_type.c_str(), "SCAN") != 0` rejects anything but `SCAN`.
- `memtier_benchmark.cpp:4679` — the continuation command is built as `"SCAN " __scan_cursor__` followed by `args[2..]`, i.e. the cursor is assumed to be `argv[1]`. For `ZSCAN key cursor [MATCH pattern] [COUNT count]` the cursor is `argv[2]` and the key has to be carried into the continuation.
- The generic parts already work unchanged: cursor extraction reads element 0 of the reply (`client.cpp:981`), and the continuation builder is placeholder-driven (`client.cpp:566`, `__scan_cursor__` → `scan_cursor_type`), so the `[cursor, [member, score, ...]]` reply shape needs no parser change.

## Proposed behaviour

- Accept `ZSCAN` in `--scan-incremental-iteration` (companion issues cover the other key-scoped variants).
- Cursor position: `argv[1]` for `SCAN`, `argv[2]` for `ZSCAN`; `argv[1]` (the key, literal or `__key__`) is preserved in the continuation command. With `__key__`, the key chosen for the initial `ZSCAN key 0` must be reused for that connection's continuations until the cursor returns `0` (per-connection state, as the cursor already is).
- Stats split as today: `ZSCAN 0` vs `ZSCAN <cursor>`.
- Keeping the current restrictions (`--pipeline 1`, exactly one `--command`, no cluster mode) is fine for a first version.

## Repro

```
redis-server --port 7781 --save "" &
for i in $(seq 0 1999); do redis-cli -p 7781 ZADD mykey $i m:$i >/dev/null; done   # >128 members → skiplist-encoded
memtier_benchmark -p 7781 --scan-incremental-iteration --command="ZSCAN mykey 0 COUNT 100" -t 1 -c 1 --pipeline 1 --test-time 3
# error: --scan-incremental-iteration requires the command to be a SCAN command.
```

Keyspace `SCAN` works in the same build (`--command="SCAN 0 COUNT 100"` → separate `Scan 0s` / `Scan <cursor>s` rows, cursor followed to `0` and restarted). Tested on current master (sha `113c3755`).

The public `ZSCAN` benchmark definitions in redis/redis-benchmarks-specification all use a fixed cursor for this reason; once this lands they can move to a real iteration.

Companion issues: HSCAN #537, SSCAN #538.

### LINKED ISSUE #537: --scan-incremental-iteration: support HSCAN (cursor is argv[2], after the key)
## Summary

`--scan-incremental-iteration` follows the returned cursor for keyspace `SCAN`, but rejects `HSCAN`:

```
$ memtier_benchmark --scan-incremental-iteration --command="HSCAN mykey 0 COUNT 100" --pipeline 1 ...
error: --scan-incremental-iteration requires the command to be a SCAN command.
```

So a `HSCAN` benchmark today can only re-issue cursor `0` (measuring the first page over and over) or pass an unrelated value as the cursor (e.g. `__key__`), which is not an iteration either. Following the real cursor matters: a full walk touches every element, exercises cursor progression across a rehash, and the per-page cost changes as the cursor advances — none of which a fixed-cursor run sees.

## Where it is hard-coded

- `memtier_benchmark.cpp:4665` — `strcasecmp(scan_cmd.command_type.c_str(), "SCAN") != 0` rejects anything but `SCAN`.
- `memtier_benchmark.cpp:4679` — the continuation command is built as `"SCAN " __scan_cursor__` followed by `args[2..]`, i.e. the cursor is assumed to be `argv[1]`. For `HSCAN key cursor [MATCH pattern] [COUNT count] [NOVALUES]` the cursor is `argv[2]` and the key has to be carried into the continuation.
- The generic parts already work unchanged: cursor extraction reads element 0 of the reply (`client.cpp:981`), and the continuation builder is placeholder-driven (`client.cpp:566`, `__scan_cursor__` → `scan_cursor_type`), so the `[cursor, [field, value, ...]] (or [cursor, [field, ...]] with NOVALUES)` reply shape needs no parser change.

## Proposed behaviour

- Accept `HSCAN` in `--scan-incremental-iteration` (companion issues cover the other key-scoped variants).
- Cursor position: `argv[1]` for `SCAN`, `argv[2]` for `HSCAN`; `argv[1]` (the key, literal or `__key__`) is preserved in the continuation command. With `__key__`, the key chosen for the initial `HSCAN key 0` must be reused for that connection's continuations until the cursor returns `0` (per-connection state, as the cursor already is).
- Stats split as today: `HSCAN 0` vs `HSCAN <cursor>`.
- Keeping the current restrictions (`--pipeline 1`, exactly one `--command`, no cluster mode) is fine for a first version.
- `NOVALUES` only changes the inner array; the cursor is still element 0, so it needs no special handling.

## Repro

```
redis-server --port 7781 --save "" &
for i in $(seq 0 1999); do redis-cli -p 7781 HSET mykey f:$i v >/dev/null; done   # >128 fields → hashtable-encoded
memtier_benchmark -p 7781 --scan-incremental-iteration --command="HSCAN mykey 0 COUNT 100" -t 1 -c 1 --pipeline 1 --test-time 3
# error: --scan-incremental-iteration requires the command to be a SCAN command.
```

Keyspace `SCAN` works in the same build (`--command="SCAN 0 COUNT 100"` → separate `Scan 0s` / `Scan <cursor>s` rows, cursor followed to `0` and restarted). Tested on current master (sha `113c3755`).

The public `HSCAN` benchmark definitions in redis/redis-benchmarks-specification all use a fixed cursor for this reason; once this lands they can move to a real iteration.

Companion issues: ZSCAN #536, SSCAN #538.

### LINKED ISSUE #538: --scan-incremental-iteration: support SSCAN (cursor is argv[2], after the key)
## Summary

`--scan-incremental-iteration` follows the returned cursor for keyspace `SCAN`, but rejects `SSCAN`:

```
$ memtier_benchmark --scan-incremental-iteration --command="SSCAN mykey 0 COUNT 100" --pipeline 1 ...
error: --scan-incremental-iteration requires the command to be a SCAN command.
```

So a `SSCAN` benchmark today can only re-issue cursor `0` (measuring the first page over and over) or pass an unrelated value as the cursor (e.g. `__key__`), which is not an iteration either. Following the real cursor matters: a full walk touches every element, exercises cursor progression across a rehash, and the per-page cost changes as the cursor advances — none of which a fixed-cursor run sees.

## Where it is hard-coded

- `memtier_benchmark.cpp:4665` — `strcasecmp(scan_cmd.command_type.c_str(), "SCAN") != 0` rejects anything but `SCAN`.
- `memtier_benchmark.cpp:4679` — the continuation command is built as `"SCAN " __scan_cursor__` followed by `args[2..]`, i.e. the cursor is assumed to be `argv[1]`. For `SSCAN key cursor [MATCH pattern] [COUNT count]` the cursor is `argv[2]` and the key has to be carried into the continuation.
- The generic parts already work unchanged: cursor extraction reads element 0 of the reply (`client.cpp:981`), and the continuation builder is placeholder-driven (`client.cpp:566`, `__scan_cursor__` → `scan_cursor_type`), so the `[cursor, [member, ...]]` reply shape needs no parser change.

## Proposed behaviour

- Accept `SSCAN` in `--scan-incremental-iteration` (companion issues cover the other key-scoped variants).
- Cursor position: `argv[1]` for `SCAN`, `argv[2]` for `SSCAN`; `argv[1]` (the key, literal or `__key__`) is preserved in the continuation command. With `__key__`, the key chosen for the initial `SSCAN key 0` must be reused for that connection's continuations until the cursor returns `0` (per-connection state, as the cursor already is).
- Stats split as today: `SSCAN 0` vs `SSCAN <cursor>`.
- Keeping the current restrictions (`--pipeline 1`, exactly one `--command`, no cluster mode) is fine for a first version.

## Repro

```
redis-server --port 7781 --save "" &
for i in $(seq 0 1999); do redis-cli -p 7781 SADD mykey m:$i >/dev/null; done   # >128 members → hashtable-encoded
memtier_benchmark -p 7781 --scan-incremental-iteration --command="SSCAN mykey 0 COUNT 100" -t 1 -c 1 --pipeline 1 --test-time 3
# error: --scan-incremental-iteration requires the command to be a SCAN command.
```

Keyspace `SCAN` works in the same build (`--command="SCAN 0 COUNT 100"` → separate `Scan 0s` / `Scan <cursor>s` rows, cursor followed to `0` and restarted). Tested on current master (sha `113c3755`).

The public `SSCAN` benchmark definitions in redis/redis-benchmarks-specification all use a fixed cursor for this reason; once this lands they can move to a real iteration.

Companion issues: ZSCAN #536, HSCAN #537.

fcostaoliveira:
`--scan-incremental-iteration` now follows cursors for HSCAN, SSCAN, and ZSCAN, using the key before the cursor and retaining each client's generated arguments throughout a walk. Generated keys advance when the walk completes or reaches its continuation cap, including with transactions and planned reconnects. The existing one-command, pipeline-1, standalone restrictions remain.

The independent CI/oracle repairs landed in #540. This PR is rebased onto master `2552312` and contains only the nine feature-related files. The rebase preserved the complete previously validated `eb2164a` tree. Subsequent changes only correct a Python test comment and clarify C++ formatting/cache invariants; executable Python and C++ are unchanged.

Continuation commands copy parsed arguments and replace only the cursor, preserving quoted, empty, and binary keys/MATCH patterns. This fixes quoted MATCH continuations for existing SCAN workloads. Keeping generated MATCH arguments fixed also fixes inconsistent query semantics: a cursor walk must retain its query filter. Previously those arguments were regenerated on continuations. We considered an opt-out and preserving regeneration only for keyspace SCAN, but chose to make the existing opt-in cursor-walk mode retain one query throughout the walk. The old changing-filter load profile is not retained; a fixed request budget can therefore cover fewer distinct generated patterns. README explicitly documents that behavior change, and `test_scan_generated_match_stays_fixed_per_walk` fails on the old master and passes here.

All four commands use the same stable initial-statistics bucket, `<COMMAND> 0`, even for a supplied nonzero cursor; the supplied cursor is still sent on the wire. Existing SCAN consumers retain compatibility; the three new commands adopt that convention for cross-command consistency. README and the man page explain this convention. Copied metadata remains valid because cursor replacement does not change key positions; continuation accounting still uses the separate stats-only entry. Cache allocation covers both argument forms in release builds, with assertions for diagnostics.

Release-note wording: HSCAN, SSCAN and ZSCAN now support `--scan-incremental-iteration`; existing SCAN walks preserve quoted arguments and keep generated MATCH arguments fixed until the walk ends. This changes generated-pattern coverage per request budget for existing users of that opt-in flag.

Closes #536. Closes #537. Closes #538.

Validation:

- 15 new integration tests cover all three commands, RESP2/RESP3, literal/binary/empty/generated arguments, continuation caps, multiple clients/threads, errors, reconnects, transactions, nonzero cursors, NOVALUES, existing SCAN behavior, and bounded MONITOR capture. They verify Redis-received arguments and structured JSON counts.
- All 15 new tests plus 15 existing SCAN tests pass in ten local build/transport cells: ASan+UBSan plaintext/TLS 1.2/TLS 1.3, TSan plaintext/TLS 1.3, no TLS with/without Prometheus, real OpenSSL 1.0.2u TLS 1.2, and optimized assertions-disabled plaintext/TLS 1.3: **300/300**. No sanitizer diagnostics. Unit tests and formatting pass.
- GCC coverage: **38/38 changed executable C++ lines (100%)**. This is added-line coverage, not full-repository coverage.
- Independent review passed 18 additional real-Redis adversarial cases. Timeout fault tests verify missing-marker and continuous-traffic deadlines, cleanup, and unchanged shared connection settings.
- Final full optimized local suites: plaintext **286 passed / 55 expected skips**, TLS **286 passed / 55 expected skips**, cluster plaintext and TLS each **162 passed / 179 expected skips**. All four 341-case suites completed with zero failures: **896 passed / 468 expected skips** in total. An earlier concurrent TLS run hit an existing Prometheus live-scrape versus final-average timing assertion; the isolated test and the complete serial TLS rerun passed unchanged.

The final source reads the immutable scan flag once before the argument loop. Nine alternating baseline/final pairs per arbitrary GET and SET workload (pipeline 1, 10 measured seconds per run, fixed client/server CPU affinity, otherwise quiet host) gave median throughput changes of **-0.84% / -0.24%**. Paired medians were **-1.16% / -0.17%**, with descriptive 95% bootstrap intervals **[-2.58%, +2.75%] / [-1.04%, +2.87%]**. Run spreads were 4.68–7.37%; both intervals include zero. These results retain the negative central estimates, establish no demonstrated regression at this resolution, and do not establish zero overhead or rule out a small regression. Shared-host scheduling and CPU-frequency variation limit causal attribution.

All applicable CI and both review bots passed on final head `eddf238`: **130 successful checks**, no failures or pending checks, and one intentional package-publishing skip. This includes CLI/MONITOR fuzz, differential testing, and all seven soak scenarios including the 30-minute memory soak. All three independent agents approve the final commit.


fcostaoliveira:
Addressed the first review in dc2480a:

- Both stale validation messages now list SCAN, SSCAN, HSCAN and ZSCAN, with assertions for the complete accepted-command list.
- The initial request sends directly from the cached generated key, removing duplicate affix assembly. The cache is sized once during client setup.
- Command names were already uppercased by `arbitrary_command`'s constructor before this code runs. Added stdout assertions to the existing lowercase-command tests, alongside JSON assertions, to lock that compatibility down.
- Documented the existing initial-statistics label `0` for caller-supplied nonzero starting cursors, preserving the SCAN interface.
- Kept explicit transaction-key invalidation: the cursor path owns command selection, so resetting its key pin is sufficient without advancing generic command rotation. The comment now describes exactly that operation; transaction cycle-progression tests pass.

The two mode checks remain in the generated-argument branches of the shared arbitrary writer; only iteration mode copies argument bytes. This avoids a second arbitrary-command implementation. No performance improvement is claimed for those checks.

All 28 new/existing SCAN tests pass after these changes. The full optimized standalone suite passed 339/339 before this small refactor and is being repeated on the new head. The earlier slow instrumented timing failures do not reproduce in the optimized full run.

5a823f7 separately repairs the three unrelated CI build failures while retaining all matrix cells: Bullseye's missing security packages use a signed immutable snapshot; the macOS OpenSSL 1.0.2 job builds libevent against that same SSL installation. Bullseye Docker build/unit tests and a locally built OpenSSL 1.0.2/libevent stack both pass; macOS remains a remote CI gate.

fcostaoliveira:
Addressing the second review:

- Confirmed the copied command metadata is intentional and remains accurate: replacing a cursor does not change the command specification or key positions. Scan replies are not miss-tracked, and continuation accounting uses the separate index-1 stats-only entry with a null specification.
- Added assertions at all three cache-indexing sites. Validation guarantees one command, and the continuation preserves its argument count; future violations now fail explicitly in assertion-enabled builds. All eight local build/transport cells still pass 28/28 tests, including ASan/UBSan, TSan, disabled TLS/Prometheus, and real OpenSSL 1.0.2.
- Full optimized local suites completed: standalone plaintext 284 passed/55 skipped, standalone TLS 284 passed/55 skipped, cluster plaintext 162 passed/177 skipped; zero failures. The description now reports these results and distinguishes expected skips. Coverage is 36/36 changed executable C++ lines.
- The CI repair remains a separate commit and is explicitly documented in the description. It is needed to exercise the existing complete matrix: Bullseye package downloads and the macOS legacy SSL build failed before reaching the feature tests. The immutable security snapshot retains signature verification and disables expiry checking only for that archived source; the libevent source is SHA-256 pinned and linked against the same OpenSSL 1.0.2 installation. All three repaired jobs have now passed remotely. Keeping the independently reviewable prerequisite commit here avoids leaving this PR dependent on a second unmerged repair PR; no matrix cell was removed or weakened.

The requested non-scan performance comparison initially suggested a SET pipeline-1 slowdown in three short, busy-host pairs. Nine longer alternating baseline/current pairs after local suites finished did not reproduce it: median throughput +0.73%, median CPU/op -0.76%, with mixed signs between pairs and varying CPU frequency. No persistent regression was detected; this is not a claim that the branches are literally free. I kept the shared argument writer rather than duplicating it or introducing template specialization without measured benefit.

A pre-existing cluster TLS statistical-test failure is still being investigated before final CI approval.

fcostaoliveira:
Addressed the latest review:

- Cache allocation now uses the maximum of the initial and continuation argument counts, so both loops are structurally bounded in release builds too. Current commands have equal counts, so allocation is unchanged. The debug assertions remain.
- Help and the man page now describe the supplied initial command, command-specific statistics labels, and a “cursor walk,” including the continuation-limit description.
- The iteration setting is already immutable for a client's lifetime. Copying it into another per-client boolean would retain the same conditional branches and duplicate state. Removing those branches requires separate command-writing paths; given the paired performance results, the shared writer remains the smaller supported implementation.
- Retained the existing Bullseye coverage and the independently reviewable CI repair commit. Dropping a platform would reduce the requested validation matrix. The archived source retains signature checks; only its expired metadata lifetime check is disabled. A separate commit also permits independent reversion.

The cluster TLS failure has been diagnosed and fixed in its own test-only commit: the old pooled Zipfian correlation bound failed on unmodified master 3/8 times and on the feature binary 2/8 times under constrained scheduling, despite identical HSET samples. The replacement verifies per-command key patterns and exact MONITOR/Redis/JSON counts. All 32 transport/topology cases and six constrained baseline/current reruns pass. Swapped-pattern and frozen-key mutations fail as intended.

Independent implementation, build, and test reviews approve the result. Final-head CI remains the gate before merge, along with the repository's required maintainer review.

fcostaoliveira:
Addressed the latest review:

- README now explicitly calls out the existing SCAN behavior change: generated MATCH arguments remain fixed for a cursor walk and are regenerated for the next walk. A dedicated regression checks both generated keys and data; it fails on unmodified master and passes here. The man page now also documents the compatible initial-statistics label for nonzero starting cursors.
- MONITOR capture now uses its own connection pool, preserving RLTest's TLS/auth/DB settings while imposing socket timeouts and an absolute drain deadline. It always closes its owned connections. New fault tests cover a missing marker, continuous unrelated frames, and a failed benchmark, then verify the original pool is unchanged and still usable.
- The maximum argument count is equal to either input today. It implements the preceding review's explicit request for structural release-build bounds on both loops; retaining it adds no current allocation or hot-path work.
- The Zipfian repair remains an isolated test-only commit because the existing oracle caused the requested full CI gate to fail. Baseline reproduction, per-command checks, exact accounting, and rejected mutations establish that this is a correction rather than relaxed thresholds or skipped coverage. Independent reviewers approved both its scope and validity.
- The Bullseye and macOS legacy jobs have passed on repeated revisions with the prerequisite repair. We have not observed a rate-limit or duration problem warranting additional cache policy in this change, and retaining those jobs preserves the requested platform coverage.

No C++ changes were needed in this round. The expanded scan tests are being run across the same ten local variants, followed by the complete remote matrix and bot checks.

fcostaoliveira:
The independent validation repairs are now split into #540. Both PRs target master because the existing workflow branch filters require that target; #540 is the prerequisite and should merge first. Its changes can then be removed from this diff by rebasing if necessary. The feature rebase preserves the entire validated Git tree exactly.

On the remaining design questions:

- Yes, regenerating MATCH during one SCAN walk is treated as a bug: it changes the query being iterated. The README explicitly documents the correction, and the generated-key/data regression fails on master and passes here.
- The `0` label identifies the initial-request bucket for the whole SCAN family. Keeping the same stable bucket convention for all four commands is deliberate; making collection labels depend on the supplied cursor would give the same flag different output rules by command. The supplied nonzero cursor is still sent unchanged and tested. Both README and man page explain this.
- The two MONITOR consumers have different requirements. Zipfian tests count textual keys; the scan oracle must preserve exact RESP argument boundaries, empty values, and binary escaping. Redis-py's joined command string loses that information. A shared implementation would require a wider parser/threading refactor, so the new bounded raw-argument capture remains local to these tests.
- `max()` is redundant for today's copied command, but implements the preceding review's requested release-build bounds for both loops. It has no current allocation or hot-path cost. Assertions provide diagnostics; the allocation size, not assertions, supplies the release bounds. Both optimized assertions-disabled transport suites pass.
- Retiring Bullseye is a separate platform-support decision. This work preserves the current matrix; both repaired Bullseye jobs and the legacy macOS job have passed repeatedly. Future archive migrations can be maintained in the prerequisite/CI context without changing iterator behavior.

Latest local validation is complete: 300/300 focused tests across ten variants; full plaintext and TLS each 286 passed/55 expected skips; full cluster 162 passed/179 expected skips; zero failures. Added executable C++ coverage is 37/37. All three agents approve the code and tests. Full remote checks are running on both reviewable PRs.

fcostaoliveira:
The feature is rebased onto the reviewed prerequisite update (now #540/e95fb1a). Compared with the previous feature head, only the prerequisite CI/test guard files changed; all scan C++ and integration tests are byte-identical.

On the latest questions:

- Yes: after #540 merges, the intended feature diff is exactly the nine feature files. Both currently target master to retain the workflow branch-filter coverage.
- Keeping regeneration for plain SCAN was considered. It would preserve the previous wire sequence, but would also preserve changing a query's MATCH filter halfway through one cursor walk and give this flag different semantics for keyspace and collection iteration. We chose the explicit bug fix, documented it, and supplied the old-master-failing regression. This is a conscious behavior correction, not a claim of unchanged wire behavior.
- The collection labels are new, so existing-consumer compatibility alone does not require `0`. The reason for choosing it is one uniform initial/continuation bucket convention for the same option across all four commands, with stable output keys independent of the supplied starting cursor. Documentation and nonzero-cursor tests establish exactly what the label means.
- Caching the flag in another member retains the per-argument branch, duplicates immutable configuration state, and at most saves a pointer load. The measured comparisons give no demonstrated reason to introduce it. The completed GET follow-up also did not reproduce the initial large slowdown: aggregate throughput +0.85%, paired median -0.63%, amid 4.76–8.08% run spreads. We do not claim sub-percent or zero overhead.
- Catching MONITOR `TimeoutError` and continuing is unsafe with redis-py: `read_response()` disconnects on error and clears parser/socket state. Reconnection would lose the continuous capture stream. Here the benchmark has finished and the ECHO reply has arrived before the drain, so frames are already queued; a read timeout deliberately fails capture rather than risk accepting incomplete evidence. The separate absolute deadline bounds continuous unrelated traffic. All ten local sanitizer/build/transport cells pass this behavior. If a real runner demonstrates the one-second socket bound is too tight, it can be raised without retrying a broken capture.

The optional CLI fuzz, MONITOR fuzz, differential, and soak gates are now enabled in addition to the full platform matrix. All three agents approve the implementation and tests; maintainer review remains required.

fcostaoliveira:
Settled the two requested points with another independent source/evidence review:

1. Bounds do not depend on assertions. Release validation rejects anything except one real command; the incremental branch in `create_request()` then calls only `create_arbitrary_request(0)` or the separate continuation and returns before generic command selection. The continuation copies the initial argv and replaces one argument's data; formatting does not change the argument count. `max(initial, continuation)` therefore bounds every actual cache access under NDEBUG. Allowing multiple real commands would also require redesigning the single cursor state and hardcoded continuation statistics index, so scanning hypothetical future commands would not make that future mode safe. We retain the current contract and the passing assertions-disabled tests rather than add speculative multi-command sizing.

2. Confirmed: the pinning regression fails specifically for regeneration, independently of quoting. It sends unquoted `SCAN 0 MATCH __key__ COUNT 1` (and separately `__data__`) with a two-continuation cap. Against pinned master `5694a3d`, the first failure is:

```
b'generated-match:2' != b'generated-match:1'
Generated MATCH changed during a SCAN walk
```

The next page similarly sends `:3` instead of `:1`; random data changes between pages too. The same test passes on this patch in plaintext and TLS and in all ten focused variants. `test_scan_quoted_match_continuations` separately covers parsed-argument preservation.

The remaining choices were weighed explicitly in the [previous reply](https://github.com/redis/memtier_benchmark/pull/539#issuecomment-5746215964): uniform initial-bucket labels, retaining one shared emitter, and failing capture on a socket timeout because redis-py disconnects. A member bool retains the branch; selecting separate emitters introduces another dispatch/code path, with no demonstrated benefit in the measured results. The short socket timeout bounds a stalled post-benchmark drain; the absolute deadline separately bounds continuous traffic.

The helper fault tests prove that missing markers, continuous frames, and failed benchmarks cannot hang or mutate the shared pool. That evidence is worth retaining after adding this capture oracle. The two SCAN regressions share its exact raw-argument capture helper and remain in this file to avoid a cross-module helper refactor; every sanitizer matrix explicitly runs both scan files.

All three agents approve the current implementation. The post-prerequisite rebase still needs the usual fresh review/gates; no bypass is proposed.

fcostaoliveira:
Addressed the final hot-loop request in eb2164a: the immutable scan flag is read once into a local `const bool` before the argument loop, and both key/data branches use it. README also explicitly explains that generated MATCH keys advance once per walk, so the same request budget can visit fewer patterns.

Fresh local validation: 300/300 focused tests across all ten variants, 38/38 added executable C++ lines covered, 2/2 unit tests and formatting passed. All four full local topology/transport suites are rerunning on the matching optimized V5 binary. The performance measurements in the description are now explicitly identified as pre-hoist measurements.

The latest resize and MONITOR-timeout suggestions were evaluated in the preceding replies: release-mode one-command validation plus fixed scan dispatch bound actual accesses; retrying a redis-py timeout would continue after disconnection and lose capture continuity. No change to those deliberate choices. The prerequisite rebase and maintainer approval remain required before merge.

fcostaoliveira:
The latest review raises the same compatibility decision already weighed in the [earlier reply](https://github.com/redis/memtier_benchmark/pull/539#issuecomment-5746215964): we considered preserving regeneration for keyspace SCAN, and chose the documented bug fix because changing MATCH mid-walk changes the query being iterated. Gating only the collection key would preserve that inconsistency. The explicit README note and isolated old-master-failing regression are deliberate; final maintainer review should evaluate that behavior choice. We do not claim it is wire-compatible with the old generated-MATCH behavior.

The prerequisite split/rebase requirement is unchanged: #540 must land first, then this feature receives a fresh post-rebase review and gates. The final intended feature diff is the nine feature files. The shared-builder tradeoff has also been evaluated: the final code hoists the immutable configuration read; a dedicated builder would duplicate argument-generation/transaction/error handling. We have not claimed zero overhead, and the measured comparison is being refreshed for the final hoist after local full suites free the host.

The new `responses.stop()` concern is not reproducible: `unittest.mock._patch.stop()` catches removal of a never-started patch from its active list and returns `None`. A direct fault probe confirmed that stop-before-start and repeated stop are safe, and a RuntimeError raised before start remains the exception after `finally: responses.stop()`. No change is needed. The helper fault tests intentionally ship because a capture oracle that can hang or corrupt shared pool state would weaken the product tests.

No further code change is justified by this review round. Final-head focused validation is 300/300 with 38/38 added executable C++ lines covered; the full local and remote suites are still running.

fcostaoliveira:
Final validation is complete on `eb2164a515a80a47c44c8b72f6f4edaa6e63bbc8`.

- **130 successful CI checks**, no failures or pending checks, one intentional package-publishing skip. This includes both review bots, CLI/MONITOR fuzz, differential testing, and all seven soak scenarios.
- **300/300 focused local tests** across ten build/transport variants; no sanitizer diagnostics. **38/38 added executable C++ lines covered (100%)**, plus unit and formatting checks. This is changed-line coverage, not a claim of complete repository/branch coverage.
- Four final full local suites: **896 passed, 468 expected skips, zero failures** across standalone/cluster and plaintext/TLS. An earlier concurrent TLS run hit the existing Prometheus live-scrape timing assertion; its isolated rerun and the full serial TLS rerun passed without source or threshold changes.
- All three independent implementation, test, and build/performance agents approve the final outcome. All bot findings have replies, and no review thread remains unresolved.

Final-source arbitrary-command performance versus master `5694a3d`, pipeline 1, nine alternating pairs per workload, ten measured seconds per run:

| Workload | Median throughput change | Paired median | Descriptive 95% bootstrap interval |
|---|---:|---:|---:|
| GET | -0.84% | -1.16% | [-2.58%, +2.75%] |
| SET | -0.24% | -0.17% | [-1.04%, +2.87%] |

Both uncertainty intervals include zero, with observed run spreads of 4.68–7.37%. The negative central estimates are retained: no demonstrated regression at this resolution, but neither zero overhead nor exclusion of a small regression is established. CPU affinity was fixed and other local agent workloads stopped; shared-host clock/scheduling variation still limits attribution. Raw pairs and provenance are retained locally.

Both PR worktrees are clean and all owned Redis processes have been stopped. #540 is also green (119 successful checks, six intentional conditional skips). The remaining gate is the repository's required maintainer approval: merge #540 first, then refresh this PR against master and require the applicable review/checks before merging.

fcostaoliveira:
Addressed the concrete documentation points from the latest review: the description now names `test_scan_generated_match_stays_fixed_per_walk`, and 322d850 corrects the affix-test comment. In scan mode that test exercises retention of short/long affixed arguments; it does not enter the ordinary stack/heap affix branch. Its executable Python AST is unchanged. The separate ordinary arbitrary-command affix tests covered that branch in the reported coverage run.

For the design questions:

- The initial `<COMMAND> 0` label deliberately has the same meaning for all four iterators: initial-request statistics, separately from continuation statistics. The wire cursor is unchanged. We considered labeling new commands by the supplied initial cursor, but chose one consistent bucket convention for consumers; README and man page state it explicitly.
- A separate scan argument builder was considered. It would duplicate key generation, transaction pinning, affix assembly, and error handling from the ordinary builder. We retained shared generation with the immutable flag read outside the loop. The reported performance intervals do not resolve small overhead; the PR explicitly retains the negative central estimates and makes no zero-overhead claim. This is a maintainability/performance tradeoff, not a claim that the extra branches are free.
- The two argument counts are equal today because the continuation copies the parsed initial command. `max` defensively sizes for both forms that actually index the cache in release builds, and per-access assertions provide debug diagnostics. Replacing that with only an equality assertion would not preserve that release bound. The existing setup comment identifies both consumers.
- The bounded-capture harness self-test intentionally remains: a hanging or corrupting observation helper would invalidate the product regressions. It performs a short timeout probe, not another memtier load test, and guards the suite in each transport environment where the helper is used.

#540 is now merged and removed from this diff. All three independent agents verified the exact rebase and reaffirmed approval. Fresh CI is running on the final comment correction; no runtime source or test threshold changed.

fcostaoliveira:
The final review's requested invariant comments are now explicit in eddf238: cursor assignment notes that the later separate formatting pass classifies `scan_cursor_type`, and cache setup records the matching copied argument indices while retaining a release-build bound for both consumers. These are comments only; executable code is unchanged.

The label question is answered in the [previous reply](https://github.com/redis/memtier_benchmark/pull/539#issuecomment-5748343903): cross-command consistency is intentional, and `<COMMAND> 0` identifies the initial-request statistics bucket, not the cursor sent. Nonzero wire cursors are tested.

I added release-note wording to the PR description for the generated-MATCH behavior change. This repository has no checked-in release-notes/changelog file; the README documents the change for users, and the PR wording is ready to carry into the next release notes.

For MONITOR timing, the drain begins only after the benchmark exits and the ECHO marker is acknowledged by Redis, so the observed command stream is already produced. The one-second socket timeout bounds a missing read; the separate absolute deadline also bounds continuous unwanted traffic. Both local TSan plaintext/TLS cells passed, as did the earlier complete final-source CI TSan cells. Fresh cells are running again. We retain the harness fault test because its deadline/cleanup behavior is necessary for trustworthy integration results; dependency-induced failures should be visible when redis-py changes.

No runtime source or test threshold was changed in this review round. The previous hot-path question is now explicitly accepted in the bot's review, and neither bot has an unresolved inline finding.

fcostaoliveira:
Both remaining design questions are now explicit in the PR description as well as the earlier replies:

- We considered an opt-out and preserving regeneration only for keyspace SCAN. We chose to keep one query throughout the existing opt-in cursor-walk mode. Changing MATCH between cursor pages is an incoherent iteration query; preserving it as a second cursor-walk behavior would perpetuate that problem. This intentionally does not retain the old changing-filter load profile, and we explicitly acknowledge that a fixed request budget can cover fewer generated patterns after upgrading. README, the named old-code-failing regression, and release-note wording document that choice.
- Existing SCAN labels preserve compatibility. The three new commands use the same initial-request versus continuation-request bucket convention for cross-command consistency; compatibility with pre-existing HSCAN/SSCAN/ZSCAN consumers is not the rationale. Supplied nonzero cursors remain unchanged on the wire.

The cache-bound nit was answered in [the prior design reply](https://github.com/redis/memtier_benchmark/pull/539#issuecomment-5748343903), and eddf238's setup comment now states that the copied commands have matching indices. We retain the small defensive release-build bound and diagnostic assertions. No further runtime change is warranted.

All three independent agents approve eddf238; final-head CI is still running, with no failures currently. Merge will wait for its completion.

PR #542, opened 2026-09-21
Isolate MONITOR input fuzzing from server command effects

fcostaoliveira:
The scheduled MONITOR-input fuzz run timed out on both master and 2.4 because it executed mutated commands against Redis: one became `WAIT 8 0`, and another replayed `HELLO 3` while memtier expected RESP2.

Run this parser fuzzer against a private streaming RESP sink that validates complete requests and replies `+OK` without executing commands. Replay both original nightly inputs unchanged before mutations, requiring actual requests. Crashes, sanitizer reports, timeouts, malformed frames and fixture failures retain a reproducer and fail the run. Per-input diagnostic draining prevents earlier errors from obscuring a later failure, including after diagnostic capping.

Each invocation parses the full input and finishes 1–100 sequential requests. Small files retain a 100-request budget; large files scale down to one to avoid amplifying replay cost or canceling a partially written frame. The outer timeout remains enforced.

The workflow checks out its own harness and corpus revision separately, using the same fixed fixture and seeds against master and 2.4 binaries. Failure artifacts have distinct matrix names. This does not add runtime support for arbitrary HELLO switching or bound blocking Redis commands. Real command/reply semantics remain covered by integration tests and the independent response fuzzer.

Validation:
- 42 harness/oracle/replay tests pass on master-source and 2.4 ASAN+UBSan binaries, including exact reproducers, binary/fragmented/pipelined requests, bounded memory/shutdown, malformed frames, resets and injected fixture failures.
- Each sanitizer binary completes 100 mutations across all 20 seeds with zero failures. The workflow's final mutation cap applies to synthetic seeds too.
- A valid 17 MiB command completes its scaled request budget on both sanitizer binaries in about 1.3 seconds locally. The committed large-command regression verifies a full request before disconnect.
- All 27 changed executable driver lines and all 222 executable sink lines covered; sink branch coverage 66/68 (97%). This is new-harness coverage, not full-product coverage.
- Python 3.10 syntax checks, workflow YAML parsing and whitespace checks pass.

Original failure: https://github.com/redis/memtier_benchmark/actions/runs/35558796567


fcostaoliveira:
Addressed in 046b720.

- Added a monotonic sink error counter and per-input baseline, including a regression proving a previous error does not fail a later input. The counter continues increasing after diagnostic capping. Every subprocess completion drains its connection before the next baseline, including crashes.
- Replaced timed replay with 100 sequential requests, so normal benchmark termination waits for complete responses and cannot intentionally truncate an in-flight large frame. The external timeout still fails. The committed large-command test requires exactly 100 requests; an additional valid 17 MiB command completed 100 times under ASAN+UBSan on both master and 2.4 in about 22 seconds locally.
- Updated the contract: the valid preflight and saved regressions intentionally require real requests. The saved inputs contain valid commands alongside their mutations; accepting a connection-refused exit would erase the isolation regression. General mutations still allow parser rejection.
- Documented that the workflow uses the same workflow-revision corpus on both branches. This is intentional. Existing maintenance-only seeds would need to be included in that shared corpus rather than silently selected by binary checkout.
- Shutdown now raises its exact message even when error storage is capped; this path is tested. Added the binary reproducer length and NUL offsets to its README.

The short deadline test deliberately holds a worker: reaching its timeout is expected, with a 1-second outer allowance for the 0.1-second close budget. The normal shutdown test unblocks idle reads immediately rather than consuming its 2-second bound. The allocation check uses an 8 MiB payload and measures a sink that streams in 64 KiB chunks. They exercise concrete failure paths; I kept these tests and will treat any CI failure as a result to investigate.

Local final evidence: 41 tests pass on both sanitizer binaries; each also completed 100 mutations across all 20 seeds with zero failures. New sink line coverage is 220/220, changed driver lines 24/24, and combined sink branch coverage 67/68.

fcostaoliveira:
Addressed the request budget and diagnostic attribution in 304c1a6.

Replay now scales from 100 requests for small inputs to one for large inputs. The 100-request small-file budget deliberately covers the current 92-line all-command seed sequentially; large inputs are parsed fully without multiplying their serialized payload 100 times. Both sanitizer binaries finish a valid 17 MiB command in about 1.3 seconds locally with the new budget. The workflow mutation cap already applies to synthetic seeds through mutate(), but scaling also protects valid uncapped local inputs.

The sink now atomically snapshots and clears diagnostics after every producer, including timeout/crash paths. There is no initial clearing that could hide a between-input fixture failure. The new input-40 regression proves a later distinct error remains visible after an earlier capped batch is drained.

Reply semantics remain explicitly tested against Redis by tests/test_arbitrary_command_misses.py, including ArrayPerElementNulls/MGET and EmptyCollection/SMEMBERS/SPOP cases. Those continue in the existing plain/TLS and cluster integration jobs. The independent protocol fuzzer exercises reply parsing; I do not claim it reproduces every client-side statistics classification. Keeping unrestricted mutated commands against real Redis would reintroduce the original WAIT/HELLO failures.

Confirmed the artifacts were mutations of the public in-tree corpus, not production traffic. AUTH contains the literal `(redacted)` placeholder. The README now records that provenance, and the comment/docstring wraps are fixed.

Final local validation: 42 tests pass on each sanitizer binary, both 100-mutation sweeps pass, and all 27 changed driver lines plus 222 sink lines are covered.

fcostaoliveira:
The current description reflects the scaled budget: small files replay up to100 requests and large files replay one. The obsolete exactly100-large-replies sentence has been removed; the committed large-input test deliberately expects one completed request.

The old time-limit rationale involved legitimate Redis error replies and server behavior. The sink never executes AUTH or other commands and always acknowledges a complete valid frame, so that path no longer applies. A malformed or truncated RESP request from the formatter is a target failure, not a benign Redis command error. If that defect also leaves memtier waiting, the outer timeout remains a failure and now prints the drained framing diagnostics. Reintroducing timed cancellation would itself produce incomplete frames and weaken this check.

Cross-branch request pinning is intentional: the saved corpus mutations still contain ordinary valid commands on both supported binaries, verified locally under sanitizers. If a future parser rejects all such commands, the regression fixture should be reviewed explicitly rather than treating a no-execution run as successful. The real-Redis integration jobs continue testing command/reply semantics.

Keeping --monitor-pattern=S explicit makes the fixed replay order independent of default/documentation changes. No production default is changed in this harness PR.

PR #543, opened 2026-09-21
Keep MONITOR reproducibility tests on the configured Redis endpoint

fcostaoliveira:
The MONITOR random reproducibility test configures the selected Redis endpoint on a temporary dictionary, discards it, and builds its benchmark from a fresh copy. On nondefault test ports memtier consequently connects to port 6379. TLS runs can fail against the wrong service, while plaintext runs can falsely pass by comparing two empty counter dictionaries from the intended Redis.

Pass the same configuration through the endpoint helper and RunConfig, require its port to equal the fixture port before launch (except Unix sockets), and require each run to record its full configured command budget before comparing counts.

Validation: on a dedicated password-protected nondefault port, a port guard rejects the original test's attempt to use 6379 before any connection. The fixed test passes on plaintext and TLS; both runs record identical nonempty counts `[21, 25, 22, 13, 19]`, totaling 100. Independent review approved the routing and stronger oracle. No production code changes.


fcostaoliveira:
Addressed in f201790. The test now requires an explicit redis_process_port in the shared configuration before constructing RunConfig (except Unix sockets), so discarding the helper's mutation fails even on CI's default 6379. Removed the shallow copies and redundant nonempty assertion; expected command count comes from the configured request budget.

Verified the guard by injecting the original discarded-mutation behavior with fixture metadata ports 6379 and 48379: both fail before Benchmark.from_json is called. The corrected test also passes against a password-protected nondefault Redis endpoint over plaintext and TLS, recording all 100 commands in both runs.

fcostaoliveira:
The guard now checks equality with the fixture's actual port and still stops before launching on mismatch. RLTest assertions accumulate failures rather than aborting immediately, so the early return is intentional: it prevents a failing guard from contacting an unrelated service. Both missing and wrong explicit ports were injected at fixture metadata 6379 and 48379; all four cases fail before benchmark construction.

I kept the repair local rather than removing RunConfig's existing default for all callers, which would change the contract used by unrelated CLI/configuration tests. The suite search found this discarded-copy bug only here. The shared configuration matches the surrounding fixtures; the PR description now describes that final implementation.

PR #544, opened 2026-09-21
Fix ownership of URI credentials and hostnames

fcostaoliveira:
A credential-free `--uri` combined with `--authenticate`, or an empty-host URI retaining a CLI/default hostname, can finish a benchmark and then abort: cleanup frees a borrowed argv pointer or string literal. Keep URI credentials and hostnames in `std::string` objects owned by `main`; CLI arguments and default values retain their existing borrowed lifetimes. URI authentication precedence is unchanged. Precedence warnings for authentication, host, port, database, and TLS are emitted next to the corresponding URI assignments, so omitted fields no longer produce misleading warnings. Explicit CLI localhost and port6379 are recognized before defaults are applied; existing database-zero sentinel behavior is preserved.

Add regressions for option order, URI precedence, repeated options, colon-containing passwords, and empty URI hosts retaining either a CLI hostname or the default. Successful cases verify actual JSON command counts. Wrong named-user credentials must fail explicitly, so the precedence tests discriminate even when the default Redis user has `nopass`. The temporary ACL user is removed afterward. Three URI parse-error cases also remain covered.

Validation:
- The committed authentication regression fails on the original source in four CLI/URI cases; the added empty-host cases fail before the host fix in both optimized and ASAN builds over plaintext/TLS.
- Fixed optimized and ASAN+UBSan+LSan builds pass over plaintext/TLS: authentication, host, warning-presence/absence, database, and TLS scenarios, plus a wrong named-user password negative control and four parse-error/validation cases. Focused three-master cluster plaintext/TLS runs also pass, with ACL users created and removed on each master.
- All changed executable C++ lines are covered: 19/19 (100%), including both outcomes of all five warning conditions.
- Existing unit checks pass in normal and sanitizer builds; formatting checks pass.
- Independent review approved both ownership and test changes. Local fixtures were loopback-only, password-protected, and stopped after testing.

Updated onto master `76b2b8f` (including the merged socket fix #541); CI and bot review are running on the final followup.

Release-note text: Fixed crashes when URI configuration retains CLI credentials or a CLI/default hostname. Corrected URI precedence warnings for omitted fields; explicit `--server=localhost` and `--port=6379` now warn when the URI supplies those fields. Existing empty-userinfo, database-zero, and `redis://` plus CLI TLS semantics are unchanged.


fcostaoliveira:
Both substantive requests are addressed in the latest followup, on top of master830ddbe.

URI hostnames now have their own main-lifetime std::string, just like URI credentials. The final URI cleanup block is removed. Empty hosts preserve either the CLI server or default hostname without attempting to free borrowed storage. Both cases reproduced the cleanup crash before the host fix in normal and ASAN builds over plaintext/TLS; both now complete successfully and verify ten SETs in JSON.

The wrong credential is now a wrong password for the explicitly created named user, not the default user. An explicit negative control requires authentication failure without timing out. Thus the precedence cases discriminate even when the default account is nopass.

The committed authentication test was run against the original production source and fails in four CLI/URI combinations; this is not only a separate manual probe. Fixed normal and ASAN+UBSan+LSan builds pass over both transports. Final coverage is 8/8 changed executable C++ lines; unit and formatting checks pass. Independent review approved the complete change.

The unsupported-scheme row also exercises the leading CLI-auth argument on the early parser error path; keeping that small overlap helps pin this ownership regression without changing existing URI validation semantics.

fcostaoliveira:
Addressed the warning mismatch in 8afce43. Auth and host precedence warnings now live inside the parser branches that actually supply those URI fields, retaining the existing warning wording and localhost exception. The test checks stderr for supplied and omitted fields, no CLI host, and an explicit localhost host. The preceding dfd7f00 binary fails nine new warning assertions; the corrected version passes.

Final normal and ASAN+UBSan+LSan runs pass over plaintext/TLS: twelve successful cases, the wrong named-user password negative control, and three parse errors per configuration. Full-PR changed executable-line coverage is 13/13; all six warning-condition branch outcomes are covered. Unit and format checks pass; independent review approved the followup.

The installed RLTest Env does provide skipOnVersionSmaller and skipOnUnixSocket (env.py lines631 and646), and these exact tests executed successfully in all four local configurations. No API change is needed. Ownership is documented at both parser assignments next to the lifetime-bearing strings; the benchmark config continues borrowing pointers from CLI/default/main-owned storage.

fcostaoliveira:
Completed the warning consistency fix in4f7b3b9. All five warnings now appear beside the corresponding actual URI assignment. Explicit CLI localhost and port6379 correctly count as supplied values because defaults have not run yet. URI fields that are absent no longer claim to override retained CLI fields. Connection and URI parsing precedence are unchanged; the existing database-zero sentinel is retained without adding option-state flags. Added a declaration-site comment documenting string lifetime.

Tests compare the complete expected warning set across omitted/present fields, explicit localhost/defaultport, empty URI fields, database selection, and rediss with/without CLI TLS. The prior version fails these assertions; fixed normal and ASAN+UBSan+LSan builds pass plaintext/TLS. Focused three-master cluster plaintext/TLS tests also pass: the ACL account is installed on every master, and database cases respect cluster DB0 constraints. The explicit6379 check exits through cluster/database validation before any connection attempt.

Coverage:19/19 changed executable lines and both outcomes of each of the five warning conditions. Unit and formatting checks pass. Independent review approved the final delta. RLTest skip helpers were verified in the installed API and these tests executed in the listed configurations. Local fixtures are stopped.

fcostaoliveira:
The redis:// + --tls combination is deliberate and covered: every ordinary redis:// case runs in the TLS test configurations after addTLSArgs appends --tls. Those cases assert the complete warning set excludes tls, complete ten SETs against the TLS fixture, and pass normal/sanitizer runs. redis:// continues to leave an explicit CLI TLS setting intact, so a precedence warning would be false. The two rediss cases additionally exercise the scheme itself enabling TLS with and without a duplicate CLI flag.

An explicit empty userinfo (@host) and an omitted userinfo (host) retain their existing distinct parser meanings. Adding auth_end != ptr would change authentication precedence and could make supplied empty credentials unexpectedly fall back to other credentials. This PR repairs lifetime and misleading diagnostics; it intentionally preserves that authentication behavior. The ownership change safely stores either string value.

Database zero remains the existing sentinel shared by unset and explicit0. Distinguishing those for an extra warning requires new option-state tracking, while host and port already have unambiguous NULL/0 unset values. The description explicitly records this limit; no database selection semantics change.

Added a release-note line for the corrected warnings, including the newly emitted messages for explicit localhost and6379. The declaration-site comment documents that URI storage stays alive for cfg and its workers and does not change after parsing. Moving owning strings onto benchmark_config would also change all CLI/default assignments and config copies, so the main-owned storage keeps this repair focused. The test's two small setup exceptions remain explicit by case name; if the TLS helper stops appending --tls, that test should fail rather than silently stop covering the intended variant.

fcostaoliveira:
The remaining database-zero asymmetry is intentional and documented in the PR description: select_db uses0 for both unset and explicitly selected DB0, unlike the distinct unset states already available for host and port. Adding option-presence tracking only to emit another warning would broaden this ownership fix. The test records current behavior; a future deliberate diagnostic change should update that expectation together with the new flag.

redis:// plus explicit --tls remains a supported combination in the surrounding URI tests. Its TLS behavior is unchanged and is covered by the new matrix as explained above. The URI parser leaves that CLI setting intact, so the incorrect precedence message is removed without introducing a new warning for a supported combination.

I considered ownership on benchmark_config, but that would change CLI/default assignment and config-copy contracts. Two main-owned strings make the lifetime fix local, with a comment recording that they remain alive and unchanged until all cfg/worker uses finish. The latest master merge did not change this URI diff; all three combined unit tests and focused plaintext/TLS cases pass afterward.

The JSON serialization observation is an existing output behavior outside this diff and is recorded separately for followup. These tests use a synthetic temporary ACL account, remove it from each master, and keep their JSON outputs in temporary directories. No real user credentials were added to the tests. I agree any broader JSON credential-redaction change should be reviewed separately rather than mixed into this repair.

Issue #535, opened 2026-09-16
libevent 2.1.12 4 KiB read cap depresses pipelined 1 KiB GET results


gs252525:
## Summary

When `memtier_benchmark` 2.4.3 is built against libevent 2.1.12-stable, libevent's `EVBUFFER_MAX_READ` limit makes the client drain socket data in at most 4 KiB chunks. For 1 KiB RESP GET replies with a pipeline of 16, this causes response parsing and pipeline replenishment to proceed in batches of roughly 3-4 replies, even when more replies are already available.

This is a client-side measurement artifact, not a Redis result. I am reporting it because `configure.ac` currently accepts libevent `>= 2.0.10`, and the effect was material in this workload.

## Environment

- memtier_benchmark 2.4.3, x86_64 musl static build
- libevent 2.1.12-stable
- Redis 8.8.0, loopback on the same 6-vCPU Linux VM
- Redis pinned to CPU 0; memtier pinned to CPUs 1-5
- 100,000 prefilled keys, 1,024-byte values, GET-only, 100% hits
- 4 threads, 40 connections per thread, uniform random keys
- 5 s warm-up and 30 s measurement windows

The two memtier binaries differed only in the libevent 2.1.12 `buffer.c` definition below:

```c
EVBUFFER_MAX_READ 4096
```

The comparison binary changed that value to 16384 and was otherwise rebuilt from the same source and toolchain. Both binaries report `libevent=2.1.12-stable` in `--version`.

## Pipeline 16 result

I alternated the variants as `A B B A A B`, where A is the default 4 KiB cap and B is the 16 KiB local comparison build.

| metric | A: 4 KiB | B: 16 KiB | change |
|---|---:|---:|---:|
| throughput, mean of 3 | 303,603 ops/s | 328,938 ops/s | +8.35% |
| p99 latency | 14.527 ms | 13.652 ms | -6.02% |

A separate `strace` diagnostic used one thread, one client, pipeline 16, and 10,000 GETs. It was not included in the throughput figures.

| variant | `readv()` | `ioctl()` | dominant `readv()` returns |
|---|---:|---:|---|
| A: 4 KiB | 2,522 | 2,524 | 4,096 bytes: 2,521 times |
| B: 16 KiB | 647 | 649 | 16,384 bytes: 323 times; 15,639 bytes: 322 times |

## Pipeline 1 control

[...]

PR #545, opened 2026-09-21
Support opt-in libevent 2.2.2-alpha and guard TLS error access

fcostaoliveira:
A TLS-enabled memtier queried the OpenSSL error queue even for plaintext connections. Libevent 2.2.2-alpha requires an SSL bufferevent there; a refused plaintext connection caused an invalid access found only after instrumenting libevent itself. Guard the accessor with the same TLS-context condition used by the constructor, and add bounded plaintext-refusal and TLS-handshake-failure regressions.

Add a checksum-pinned helper for isolated 2.1.13-stable (default) or opt-in 2.2.2-alpha builds, with documented incremental builds. Existing system-library builds and minimum versions stay unchanged. `--version` already reports the runtime libevent; CI asserts that field and verifies every linked libevent component comes from the selected prefix.

The six compatibility jobs run full standalone and cluster suites: stable/alpha × plaintext/TLS/TLS-disabled. The alpha plaintext job instruments **both memtier and libevent** with ASan/UBSan, including the new TLS-handshake-failure test in its TLS-enabled binary. Jobs run for build/connection-related changes, weekly, and on demand. Existing macOS custom builds reuse the helper with an explicit stable pin and CMake 4 compatibility.

Validation:
- Upstream libevent alpha CTest: 84/84 passed (Release build).
- All memtier unit targets passed after the guard in stable, alpha, all four alpha TLS/Prometheus compile combinations, and memtier ASan+UBSan builds.
- Before/after regression: the new plaintext-refusal test fails on the preserved master-source binary with instrumented alpha (`AddressSanitizer: heap-buffer-overflow`, `bufferevent_ssl.c:1108` → `bufferevent_openssl.c:543` → `shard_connection.cpp:1475`) and passes after the guard. Stable safely tolerates the original call, so failure is correctly specific to the alpha/instrumented cell.
- After the guard, both memtier and libevent instrumented: focused plaintext and TLS suites each **40 passed, one skipped, no sanitizer findings**, including the new error-path regressions.
- Earlier integration validation on the same base source: alpha standalone plaintext/TLS each 119 passed, 15 skipped; cluster plaintext/TLS each 68 passed, 66 skipped; TLS/Prometheus-disabled standalone 94 passed, 13 skipped and cluster 50 passed, 57 skipped. Final-head CI reruns the full suites with the guard.
- Helper builds, checksum/error-path probes, loader checks, incremental builds, CMake 4.0.3, shell syntax and actionlint passed. Independent reviews covered the implementation and benchmark evidence.
- Existing macOS paths passed on 709dab0: [OpenSSL 1.1](https://github.com/redis/memtier_benchmark/actions/runs/35620699852/job/106402729098) and [OpenSSL 3.0](https://github.com/redis/memtier_benchmark/actions/runs/35620699852/job/106402729099). Libevent explicitly uses an absolute [Apple INSTALL_NAME_DIR](https://github.com/libevent/libevent/blob/release-2.1.13-stable/cmake/AddEventLibrary.cmake#L147); no additional loader workaround is required. CI reruns on the final head before merging.

A separate fixed-source, 24-run whole-release comparison completed with zero connection errors or GET misses. Independent trace review confirmed P16 positive socket reads fell from 2,526 (max 4 KiB) to 913 (max 16 KiB), with identical byte totals. Shared-host load and large run-to-run variation prevent a reliable quantitative performance claim. These measurements used the original fixed source, before the error-path guard; they are not presented as final-head performance measurements.

Known alpha limitation: its SSL error accessor currently returns no queued reasons for genuine TLS bufferevents. Detailed TLS diagnostics may differ from stable; failure/reconnect behavior is tested, and the limitation is documented. Alpha remains optional. No claim of 100% line/branch coverage.

Related to #535; this adds a tested evaluation path, not a stable dependency upgrade.

---

> [!NOTE]
> **Medium Risk**
> Changes connection error handling on the reconnect path for all TLS-enabled builds; behavior is narrowed (safer) but still touches hot networking code, with broader CI coverage mitigating regressions.
> 
> **Overview**
> Fixes a **libevent 2.2.2-alpha** crash/UB when a TLS-capable binary hits a **plaintext** connection error: `bufferevent_get_openssl_error` is now only called when `m_config->openssl_ctx` is set, matching how SSL bufferevents are created.
> 
> Adds an **opt-in, checksum-pinned** `scripts/build-libevent.sh` (default **2.1.13-stable**, optional **2.2.2-alpha**) and wires macOS CI libevent builds through it (with **cmake** on brew). A new **Libevent compatibility** workflow runs a **2.1 / 2.2 × plaintext / TLS / no-TLS** matrix with linkage checks, unit tests, and full integration suites; the alpha plaintext job can **ASan/UBSan both memtier and libevent**.
> 
> **`tests/test_connection_error_types.py`** covers refused plaintext (no TLS error spam, no sanitizer) and TLS handshake failure/reconnect. **DEVELOPMENT.md** documents local alpha evaluation; **.gitignore** ignores private libevent prefix build dirs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b3c67f1dfbb137f90563fb2b504f38b0d9ce1d4a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


fcostaoliveira:
Addressed in b3985ac:

- The existing macOS wrapper now uses the shared helper, so the 2.1.13 pin/checksum has one definition. The separate OpenSSL 1.0.2 job retains its existing 2.1.12/autotools build; migrating that legacy combination is outside this change. Both affected macOS jobs now install CMake explicitly. A CMake 4.0.3 build also caught and verified the required policy-minimum compatibility setting for stable libevent.
- Compatibility CI now runs on build-related path changes, weekly, and via workflow_dispatch. It does not run on unrelated source/documentation PRs. Existing distribution-library CI remains unchanged.
- Removed the enumerated TEST list and string substitutions. Each compatibility job now runs the complete standalone and cluster suites. The no-TLS configuration keeps the default Prometheus support enabled; the four TLS/Prometheus compile combinations were separately checked locally.
- The helper prefers sha256sum on Linux and falls back to shasum for macOS.

On the read limit: no new memtier API call is needed. In 2.2.2, [bufferevent_ratelim_init_](https://github.com/libevent/libevent/blob/release-2.2.2-alpha/bufferevent_ratelim.c#L1101) initializes max_single_read from MAX_SINGLE_READ_DEFAULT (16384) and immediately passes it to evbuffer_set_max_read on the input buffer. The 2.1.x inner evbuffer clamp is what prevents its nominal 16 KiB bufferevent limit from taking effect. The source review and Linux syscall preflight confirm 16 KiB plaintext reads with the alpha library. Platforms without FIONREAD have a separate fallback, so this is not an all-platform performance claim.

One additional measurement detail: alpha uses read() for a single iovec, while stable uses readv(). The diagnostic now accounts for both and verifies every response byte. The timed comparison will report whole-release changes; it will not attribute the entire difference to the read cap alone. Local cluster runs and the timed comparison are still in progress, and the expanded CI is running on the new head.

fcostaoliveira:
The two affected macOS jobs passed on b3985ac, including executable startup and the existing linkage checks:

- [OpenSSL 3.0](https://github.com/redis/memtier_benchmark/actions/runs/35616203697/job/106387441469)
- [OpenSSL 1.1](https://github.com/redis/memtier_benchmark/actions/runs/35616203697/job/106387441200)

Libevent's [APPLE target configuration](https://github.com/libevent/libevent/blob/release-2.1.13-stable/cmake/AddEventLibrary.cmake#L147) explicitly sets INSTALL_NAME_DIR to the installation prefix's lib directory, so the generic CMake @rpath default does not apply here. No additional loader workaround is needed.

Added tests/run_tests.sh, tests/test_requirements.txt, and tests/gen-test-certs.sh to both path filters in 0e4c622. The PR description now explicitly distinguishes the full integration suites in CI from the focused local ASan/UBSan runs. All four local default-build integration configurations (standalone/cluster × plaintext/TLS) have passed; minimal-build runs and the timed comparison are still progressing. CI is running again on the latest head; I canceled superseded in-progress runs to avoid duplicating their work.

fcostaoliveira:
Addressed the stable-selection safeguard in 709dab0: the existing macOS wrapper explicitly sets `LIBEVENT_VERSION=2.1.13-stable`, while the generic helper retains the alpha opt-in. Verified both unset and exported-alpha caller environments.

The affected macOS jobs also passed on 0e4c622: [OpenSSL 1.1](https://github.com/redis/memtier_benchmark/actions/runs/35617417285/job/106391580315) and [OpenSSL 3.0](https://github.com/redis/memtier_benchmark/actions/runs/35617417285/job/106391580349). Libevent explicitly sets an absolute [Apple INSTALL_NAME_DIR](https://github.com/libevent/libevent/blob/release-2.1.13-stable/cmake/AddEventLibrary.cmake#L147). This evidence is now in the PR description as well.

All six local integration configurations and both focused sanitizer runs passed: 594 passed, 234 skipped, zero failures in total. Upstream alpha CTest passed 84/84. CI is rerunning for the final one-line safeguard.

fcostaoliveira:
Thanks; these are intentional scope choices:

- The macOS-wrapper path trigger is conservative: that wrapper is a consumer of the same helper and pinned default, so changes to how it selects/configures the dependency run the shared-helper compatibility matrix too. The actual wrapper is exercised by the existing macOS CI jobs. All four macOS jobs passed on the final 709dab0 head, including [OpenSSL 1.1](https://github.com/redis/memtier_benchmark/actions/runs/35620699852/job/106402729098) and [OpenSSL 3.0](https://github.com/redis/memtier_benchmark/actions/runs/35620699852/job/106402729099). Keeping this extra build-related trigger is a deliberate coverage/cost tradeoff; ordinary C++/documentation changes do not trigger this matrix.
- Stable 2.1.13 is the control for the isolated-prefix/shared-helper path. Distribution CI uses the distribution's libevent, not necessarily this release or the same build/linkage path. It also protects the pinned dependency now used by the existing custom macOS builds. The scheduled/manual coverage remains useful independently of whether alpha eventually becomes suitable as a default; this PR makes no such default-upgrade decision.
- Requiring an empty install prefix intentionally prevents mixing stale installed headers/libraries after a failed or changed build. A failure before install leaves an empty reusable prefix; a partial install requires a fresh prefix. The message and documentation recommend an empty/fresh prefix. I prefer not to suggest an unconditional recursive deletion command against a caller-supplied directory. Fresh GitHub runner prefixes are unaffected.
- This workflow guards compatibility and actual loader selection. It does not claim to guard throughput or every libevent backend's read limit. A public-API max-read value alone would be a misleading regression assertion: 2.1 already advertises a 16 KiB bufferevent limit while its inner evbuffer clamps actual reads. A Linux syscall-level check can observe the distinction, as the separate diagnostic does, but scheduling/availability and non-FIONREAD backends need different expectations. Keeping that experiment separate avoids encoding a performance claim into a functional compatibility job.
- When adding another supported version, maintainers must update the helper checksum case and CI matrix/runtime-version expectations together. The existing inline comment explains the alpha runtime suffix; the workflow's exact-version assertion deliberately fails if upstream changes that convention.

The completed diagnostic accounts for exactly 10,330,005 response bytes in each of four traces. At P16, stable made 2,526 positive socket reads (max 4,096 bytes) and alpha made 913 (max 16,384 bytes); P1 used 10,001 calls for both. Alpha uses read() for a single iovec, so both read and readv were counted. An independent reviewer reparsed the raw traces and all 24 untraced performance samples. Rising shared-host load makes the numerical throughput deltas inconclusive; the read-size/path observation itself is verified. The result supports an opt-in evaluation path, not a promised speedup or default dependency upgrade.

fcostaoliveira:
Additional dependency-instrumented validation found a real compatibility defect and changed the scope. Fixed in 05a8415:

- A TLS-enabled memtier called `bufferevent_get_openssl_error` for plaintext BEV_EVENT_ERROR too. Stable safely returned zero, but alpha's SSL-only accessor accesses outside the plain bufferevent allocation. Instrumenting only memtier did not expose this; instrumenting libevent did.
- The call is now guarded by the same `openssl_ctx` condition used when constructing the bufferevent. Added bounded plaintext connection-refusal and TLS handshake-failure tests, with reconnect and sanitizer assertions.
- The existing alpha plaintext CI cell now instruments both memtier and libevent with ASan/UBSan. It remains a TLS-capable binary, so both new regression paths run there. No additional matrix jobs were added. Connection implementation and regression-test paths now trigger the matrix.
- Documented the separate alpha limitation: its accessor currently returns no queued OpenSSL reasons for genuine TLS bufferevents. The guard prevents the plaintext invalid access; it does not pretend to repair that upstream diagnostic behavior.

After the fix, instrumented plaintext and TLS suites each passed 40 tests, with one expected skip and no sanitizer findings. All six local build configurations rebuilt and passed unit tests. Both independent reviewers approved the fix/tests/workflow. Final-head CI is running; merge remains gated on its result.

fcostaoliveira:
Addressing the latest review:

1. Yes, the regression fails on the preserved master-source binary **with instrumented alpha**, and passes after the guard. The before-fix ASan stack is `heap-buffer-overflow` → `bufferevent_get_ssl_error` (`bufferevent_ssl.c:1108`) → `bufferevent_get_openssl_error` (`bufferevent_openssl.c:543`) → `shard_connection::handle_event` (`shard_connection.cpp:1475`). The test fails first on the signal exit, and the diagnostic is also present. The direct driver explicitly requires this expected baseline failure. Added the evidence to the description. Stable's defensive upcast makes this specific to alpha; forcing an artificial stable failure would misrepresent the regression. The single instrumented cell is intentional and still runs the full suites plus both new transport regressions.
2. Agreed this is an alpha diagnostics regression, including the risk that fallback errno text is not an SSL protocol reason. The guard does not introduce that TLS behavior: on a genuine TLS connection it leaves the accessor loop/fallback unchanged. This opt-in build documents the upstream limitation rather than adding a version-specific workaround or claiming alpha is production-ready. `ERR_get_error()` is not an appropriate substitute because libevent already drains that queue.
3. Intentional: the BEV_EVENT_ERROR callback operates on the current live bufferevent, installed by setup_event; it is not delivered for a null m_bev. The constructor's TLS-context predicate is the relevant type guard. A second defensive null check is unnecessary here.
4. Keeping abort_on_error is intentional. Any sanitizer finding, including a leak on a failure path, should fail this validation. Turning abort off changes the failure exit form, not whether a leak should be accepted. The tests reject both signal death and sanitizer diagnostics, and both instrumented local suites passed with these exact options.
5. The trigger is intentionally scoped to this shared dependency build and the connection implementation/regressions under test. The weekly/manual matrix covers other libevent consumers; existing distribution CI still runs for their ordinary source changes. This is the coverage/cost compromise described in earlier review, not a claim that every libevent API caller triggers this workflow.
6. Widening the test window from 2 to 5 seconds gives margin over the fixture's one-second read timeout and initial reconnect delay while retaining the stronger two-actual-connections assertion. The subprocess remains bounded at 15 seconds. Switching the capability probe to the existing `--version` openssl field also removes the help-format dependency.

Restoring the unrelated trailing blank line. The OpenSSL@3 documentation selection stays because it aligns the optional macOS build recipe with the selected CI OpenSSL. The small test/documentation follow-up passed all six local regression configurations (10 passes, two TLS-disabled skips), and the original binary again produced the expected ASan failure. It is pushed; production code is unchanged from 05a8415. Independent review approved the timing and capability-probe changes.

fcostaoliveira:
Settling the remaining scope choices for this PR:

- Keeping all six cells for the listed PR paths. The actual compatibility failure was alpha-specific and invisible to the stable control and memtier-only sanitizers, so moving alpha to a weekly-only check would remove the pre-merge regression guarantee we just added. The matrix remains scoped rather than running for every source change. This is a deliberate acceptance of the added CI cost to validate the optional dependency path. The comment directly above SANITIZE already records why dependency instrumentation is necessary; the description and regression test explain the plaintext/TLS-enabled shape. There are three stable cells (plaintext, TLS, TLS-disabled), with the same three alpha configurations.
- Negative Python subprocess return codes indicate signal termination. The `>= 0` assertion specifically rejects those; the following assertions independently require connection-error/reconnect evidence and reject sanitizer diagnostics. Expected ordinary benchmark failure exit codes are allowed because these fixtures deliberately cannot serve a workload.
- The PKG_CONFIG_PATH selection remains useful: CMake's FindOpenSSL invokes pkg_check_modules(_OPENSSL ... openssl) and uses the resulting hints. OPENSSL_ROOT_DIR is also explicit; both point at the same selected OpenSSL to avoid ambient Homebrew/pkg-config discovery selecting a different one.
- The nonempty-prefix rejection applies to an existing nonempty directory. A newly created directory is empty and passes that guard; it cannot be the directory just created by mkdir that triggers the rejection absent concurrent external modification. Empty prefixes left by pre-install failures remain reusable; partially installed prefixes deliberately require a fresh destination.

No further code change is needed for these decisions. Both independent implementation reviewers approve the current outcome. Waiting for all final-head checks before merge.

